PR #6, opened 2015-06-04
Use nodemon to restart app server on file change/save

jeremiak:
This PR isn't related to a documented issue, but Sails does not automatically reload the app server once you make a file change, which can get kind of annoying when you start debugging an issue only to remember that you didn't restart the server. :disappointed: 

I added [`nodemon`](http://npm.im/nodemon) as a dependency and have `npm start` use nodemon to watch for file changes and restart the Sails server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ To build sites using Hugo, install [Hugo](http://gohugo.io/overview/installing/)
     secret: 'testSecret'
   }
   ```
-* Run the server with `npm start` or `node app.js`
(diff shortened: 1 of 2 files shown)


dhcole:
Yea, good idea. How about doing it as a separate script from `npm start` so that in production we can run it with plain `node`. Something like this? https://github.com/18F/midas/blob/devel/package.json#L83

Then we can include `nodemon` as a dev dependency instead (`npm install nodemon --save-dev`)

jeremiak:
Good points, totally agree and changed.

dhcole:
Nice. :ship:

PR #22, opened 2015-06-08
Add policies to controllers

### LINKED ISSUE #21: API should enable front end to check for authentication
This is somewhat related to #19 but we'll want to show different views based on if the user is logged in or not. For instance, having a log in link if the user isn't authenticated already.

@dhcole can you think of a way that the front end can test for auth with the current API? If not, should we just place another cookie to reference the authentication status?

dhcole:
- Requires an authenticated session for  `/build`, `/site`, and `/user`
- Filters `/build` and `/site` to only return models associated with the authenticated user
- Filters `/user` to only return the authenticated user

Resolves #21 

@jeremiak mind validating?

### DIFF
--- a/config/policies.js
+++ b/config/policies.js
@@ -49,6 +49,10 @@ module.exports.policies = {
 		// feed : ['isNiceToAnimals', 'hasRabbitFood']
 	// }
 
+  BuildController: ['passport', 'sessionAuth', 'filterCurrentUser'],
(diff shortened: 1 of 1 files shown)


jeremiak:
@dhcole nope, I'll go ahead and check it out and merge if it's all working. Thanks for these, btw :)

jeremiak:
:+1:

PR #24, opened 2015-06-08
Remove passports from user model API response

### LINKED ISSUE #19: /v0/user endpoint shouldn't return access tokens
Currently, the `v0/user` endpoint returns all of the users in the database including valid access tokens for the Github accounts.

I'd imagine that the `v0/user` endpoint should simply return the representation of the authenticated user? @dhcole any thoughts on what this endpoint should provide?

dhcole:
Blocks passport models from being returned as part of user responses.

Resolves #19

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -20,6 +20,13 @@ var User = {
     builds: {
       collection: 'build',
       via: 'user'
+    },
+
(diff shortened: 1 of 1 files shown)


jeremiak:
Looks good to me.

PR #25, opened 2015-06-09
Front end basic app

### LINKED ISSUE #14: Front end website for users to interact with Federalist
Federalist is an API-driven application, so it needs a front-end website for users to interact with it:

- [x] Log in with GitHub
- [x] Add sites
- [x] Browse sites associated with user
- [x] Launch a site in Prose to edit it
- [x] Track build status messages

jeremiak:
This is the basic backbone app (and horrendous styles) to close #14 and provides a front end to interact with Federalist.

All items in #14 were addressed and I want to make sure we don't have too many long-living branches, especially since I merged in `master` after #22, #23 and #24.

The readme and `package.json` have been updated to reflect new workflows for working on the front end. I'm going to open a new issue for ensuring responsiveness of the UI (in terms of different device sizes) and making it look better than a pile of browser default styles.


assets/app/models/Site.js
```diff
@@ -0,0 +1,23 @@
+var Backbone = require('backbone');
```
dhcole: Does Backbone require jQuery for for ajax requests?

```js
Backbone.$ = require('jquery');
```

assets/app/models/Site.js
```diff
@@ -0,0 +1,23 @@
...
+
+var SiteModel = Backbone.Model.extend({
+  defaults: {
+    'builds': [],
+    'repository': 'Repo goes here',
+    'engine': 'jekyll',
+    'branch': 'master'
+  },
+  urlRoot: '/v0/site'
```
dhcole: Is this redundant with the `collection.url`? Should we pull out the `/v0` into a global `api.version` var, or somehow automatically prefix it to the request? Or we could follow GitHub's lead and version the API with a header instead of in the URL.

assets/app/templates/AddSiteTemplate.html
```diff
@@ -0,0 +1,19 @@
+<form>
+  <label for="owner">owner</label>
+  <input type="text" name="owner" id="owner" value="jeremiak"></input>
```
dhcole: Do we need to commit these default values?

assets/app/views/AddSiteView.js
```diff
@@ -0,0 +1,41 @@
+var fs = require('fs');
+
+var Backbone = require('backbone');
+var _ = require('underscore');
+
+var SiteModel = require('../models/Site').model;
+var templateHtml = fs.readFileSync(__dirname + '/../templates/AddSiteTemplate.html').toString();
```
dhcole: Have you considered using https://github.com/JohnPostlethwait/stringify for string requires? I've found that it's cleaner to just require the file and not need `fs` or `__dirname`. Your call though.

assets/app/views/AddSiteView.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  initialize: function initializeSiteView() {
+    this.render();
+  },
+  render: function renderSiteView() {
+    this.$el.html(templateHtml);
+  },
+  onSubmit: function onSubmit() {
+    var data = {};
+    this.$('form').serializeArray().map(function(d) {
```
dhcole: Do we need to set jQuery for this to work?

```js
Backbone.$ = require('jquery');
```

assets/app/templates/AddSiteTemplate.html
jeremiak: Nope, we should rip them out - #32

assets/app/models/Site.js
jeremiak: I'm open to any of these approaches. I like the idea of using the api header because it removes it as a concern across all of the models but I also like it in the URL for debugging purposes.

It is slightly redundant but we can refactor it pretty easily. Will do in #33

assets/app/views/AddSiteView.js
jeremiak: Cool, I'll take a look at stringify but the current way of doing it is fine for me as well.

assets/app/models/Site.js
jeremiak: No, I'm pretty sure it has a built in version of jQuery. I generally use this particular style ^^ when I want to set a particular version of jQuery to be used by Backbone or I already have it on the page.

assets/app/views/AddSiteView.js
jeremiak: See other comment but tldr: no, just if we want to set a particular version

dhcole:
Cool, looks great! :ship:

PR #29, opened 2015-06-09
Readme improvements

### LINKED ISSUE #28: Make README easier for all types of people to get started
We'll open more issues like this in the future, but since we have @bradnunnally going through the process for the first time I'm going to use this issue to log all the bits that are currently confusing so we can check through them and improve!

jeremiak:
Used the onboarding of @bradnunnally as a test of the documentation. Here are the improvements to close #28


dhcole:
This is redundant with #45 now, right?

jeremiak:
Not redundant, @dhcole because #45 updated the `readme-improvements` branch instead of `master`. This is to merge back down to `master`.

dhcole:
Got it. :+1:

PR #34, opened 2015-06-11
Run browserify before starting node app

jeremiak:
In the case where somebody is running `npm start` such as on a deploy, they're going to need a bundle that is freshly compiled. So I've altered `npm start` to run `browserify` so that there is a `bundle.js` and puts it where expected.


dhcole:
:+1:

PR #35, opened 2015-06-11
Dry up url for Site model and collection

### LINKED ISSUE #33: URL is not DRY in Site model
Right now we have a redudancy with the URL in the [Sites model](https://github.com/18F/federalist/blob/front-end/assets/app/models/Site.js). Let's get a quick refactor so that those make sure to always stay insync with each other.

jeremiak:
The Site model and collection share the same URL but it was listed in both models. We still have an open question of how we will version the API, but until we decide one way or the other here is a small dry commit.

This will close #33

### DIFF
--- a/assets/app/models/Site.js
+++ b/assets/app/models/Site.js
@@ -1,5 +1,7 @@
 var Backbone = require('backbone');
 
+var siteUrl = '/v0/site';
+
 var SiteModel = Backbone.Model.extend({
   defaults: {
     'builds': [],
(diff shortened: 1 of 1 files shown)


dhcole:
I'm pretty you can just set the `Collection.url = '/v0/site'` and the model will delegate to its collection to form a URL + id, like `/v0/site/1`. http://backbonejs.org/#Model-url

Does that work here?

jeremiak:
You can do that but because I use the model outside of a collection, namely to [create a new one with the new site view](https://github.com/18F/federalist/blob/pull/dry-site-model/assets/app/views/AddSiteView.js#L26) I need to have the URL explicitly in the model as well. 

Or the new site form could just delegate the submission to the collection from within the main view. Is there a pattern that you prefer? I'm open to either at this stage

dhcole:
Cool, that makes sense. Let's go with it.

PR #36, opened 2015-06-11
Only show one add new site form

### LINKED ISSUE #31: Clicking anywhere in the top section results in yet another add site view
Bug found by @bradnunnally when getting the project set up. It is also pretty easy to reproduce, here are the steps:

0. Launch [the app](http://localhost:1337)
0. Authenticate with Github if you haven't done so already
0. Click 'Add Site' link near the top
0. Try and click into one of the fields on the 'Add Site' form
0. Observe how the form gets repeatedly added to the page

jeremiak:
This PR should close #31 by only having one add site form on the site. I've gone ahead and made sure that there is only one add new site view instantiated at a time in the application and I store a reference to it. Then instead of the action triggering a view creation and insertion it just toggles a display class.

### DIFF
--- a/assets/app/templates/SiteListTemplate.html
+++ b/assets/app/templates/SiteListTemplate.html
@@ -1,6 +1,6 @@
 <% if(authenticated) { %>
   <h1>Your current sites</h1>
-  <a href="#" class="add-site">Add a site</a>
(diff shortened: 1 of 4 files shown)


dhcole:
:+1:

PR #37, opened 2015-06-11
ok, lets add materialcss!

### LINKED ISSUE #26: Front end should look good on any device
The result of #25 is pretty ugly and doesn't work that well for mobile viewports. The app should look professional and render well on phones and desktop machines

jeremiak:
This is our experiment with the [materialcss](http://materializecss.com) design framework. The sites listing page looks something like this 
screen shot 2015-06-10 at 11 19 58 pm

As a fortunate side effect, it will also close #26.

### DIFF
--- a/assets/app/app.js
+++ b/assets/app/app.js
@@ -1,3 +1,6 @@
+var Backbone = require('backbone');
+var $ = window.jQuery = window.$ = Backbone.$;
+
 var UserModel = require('./models/User');
 var UserView = require('./views/UserView');
 var SiteListView = require('./views/SiteListView');
--- a/assets/app/templates/SiteListItemTemplate.html
+++ b/assets/app/templates/SiteListItemTemplate.html
@@ -1,9 +1,18 @@
-<p>repository: <%= repository %></p>
-<p>branch: <%= branch %></p>
-<p>owner: <%= owner %></p>
-<p>engine: <%= engine %></p>
-<p>builds: <%= builds.length %></p>
+<div class="collapsible-header active"><%= repository %>
+  <span class="right">
+    <a href="http://prose.io/#<%= owner %>/<%= repository %>" target="_blank" class="edit" alt="edit the site <%= repository %> using prose.io"><i class="mdi-editor-mode-edit"></i></a>
+    <a href="#" class="delete" alt="delete the site <%= repository %>"><i class="mdi-action-delete"></i></a>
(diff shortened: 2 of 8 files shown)


jeremiak:
hey @dhcole looks like there's a gonna be some merge conflicts. Would you prefer that I merge from master into this branch and add that to the PR so it merges cleanly?

dhcole:
@jeremiak nice. Looks great. Would love to hear your thoughts on this as you get more familiar with the it.

Yea, please resolve the conflict and merge in master.

jeremiak:
Cool, done.

So far the framework is OK. It looks very pretty but the types of elements aren't as varied (read: comprehensive) as Bootstraps and the documentation leaves a bit to be desired. Also, IDK if you noticed but I'm actually getting the [css](https://github.com/18F/federalist/blob/materialcss/assets/index.html#L4) and [js](https://github.com/18F/federalist/blob/materialcss/assets/index.html#L21) for the framework from CDNs because the `npm install` is.... weird and frankly kind of worthless.

dhcole:
OK enough to keep running with it or would you prefer to try something else? Your call. 

Also, we should figure out how to avoid the CDN links -- we generally don't include resources from any third-party domains in our projects (but fine for now).

jeremiak:
Let's just keep going with it for now. It looks pretty and works well enough (we're not doing things that are too complex).

I agree re: pulling out the CDN links. Just for background:
the materialize npm doesn't have a main or any JS for that matter so it just pulls down compiled JS and CSS to `./node-modules/materialize-css/bin/` - we could easily have a small NPM task to move those over to something like `assets/vendor`. How about we merge and use #43?

dhcole:
:+1: 

I'll look into #43. Have an idea there.

dhcole:
@jeremiak Just sharing some thoughts here... Materialize is an implementation of Google's Material, which is just a style guide (as opposed to an official thing from Google). To better understand the how and why for Materialize, I've been reading through the actual Material style-guide: http://www.google.com/design/spec/material-design/introduction.html

I think there's a lot of really interesting research there that we use to inform how we implement Materialize while making the branding our own. 

(I realize there is nothing ground-breaking in this comment :) Just posting for discussion / posterity.)

dhcole:
@bradnunnally also would love your thoughts on the above. Not sure if you're already familiar with Material.

PR #44, opened 2015-06-11
Copy materialize assets on npm install

### LINKED ISSUE #43: Pull out all CDN links
We don't want to request anything from not our domain. To achieve this we'll need to pull out the materialize css links for both the CSS and JS

dhcole:
Resolves #43

@jeremiak went with your suggested approach. We could call in the repo from github, but then we're just pulling from the repo `head` and we'll be getting constantly whatever the latest code is... or we pin to a commit but then we're off release cycle. I was also getting issues using browserify with materialize, as if there are missing upstream dependencies... so this works for now! /shrug

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,11 @@ assets/site
 # Browserify bundle
 assets/js/bundle.js
 
+# Materialize assets
+assets/font/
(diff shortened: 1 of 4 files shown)


jeremiak:
Cool :+1:

Issue #16, opened 2015-06-05
Add tests for API controllers, services, and models


dhcole:
All custom methods for the API should have sets. These are bootstrapped in `/test/` but needs to be built out.

dhcole:
Testing framework and model tests are in the above PR. Closing and will add more tests as we build.

PR #46, opened 2015-06-11
Unit tests for API models

dhcole:
Includes unit tests for all custom model methods and attribute methods for the API models.

Part of #16 

Next up: controller tests.


jeremiak:
hey @dhcole getting an error when I run `npm run test`:
```
> mocha test/bootstrap.test.js test/unit/**/*.test.js



  1) "before all" hook
  2) "after all" hook

  0 passing (2s)
  2 failing

  1)  "before all" hook:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
  

  2)  "after all" hook:
     TypeError: Cannot read property 'lower' of undefined
      at Context.<anonymous> (test/bootstrap.test.js:20:8)

```

Any ideas? I have the dependencies installed with `npm install`

dhcole:
Hmm... that means sails isn't booting up before the test times out. 2 seconds may be too fast of a timeout. Pull down and try it again. I've increased the timeout time to 1 minute.

jeremiak:
:+1: works great! I'll merge it now

Issue #32, opened 2015-06-10
Add Site has bad defaults


jeremiak:
There are two problems with the current add sites template:

0. It has [committed defaults](https://github.com/18F/federalist/blob/front-end/assets/app/templates/AddSiteTemplate.html#L3), which are almost definitely a bad thing
0. The defaults should be dynamic so that they actually help the user complete the task of adding a new site. For instance, make sure the user reflects the authenticated user, and not just user 1

PR #52, opened 2015-06-13
Pull/sane defaults for add site and refactor

jeremiak:
This should close out #32 and make it so that the logged in user can actually add in sites and there are no hardcoded defaults besides using jekyll.

This PR also includes a somewhat large PR of how the orchestration happens on the front end. The views now communicate with each other through an event bus that lives in `app.js`. It makes the front end much easier to work with :smile:


dhcole:
:+1:

Issue #54, opened 2015-06-13
Starting from template should fork repo and set up webhook


jeremiak:
If you select a new site using one of our templates, then we'll need to fork that repo over to the users and then set up the normal webhooks we use on that new repo. This should also be exposed as an API endpoint so that the front end app can leverage it.

Issue #12, opened 2015-06-05
Set up starter templates for new sites


dhcole:
Instead of building their own Jekyll sites, developers may prefer to start with a template. We could offer a few templates that are pre-configured to work well with Federalist / Prose. Developers could select one, which would fork it to their own GitHub account, and allow them to make edit content through Prose or make changes to the layout through GitHub / local editing.

This raises some general questions: 

- What should a generic starter template include? For instance, what templates, page types, basic features does it need?
- What open source templates are available that we can use / modify for our purposes?

Steps to ship:

- [ ] Identify template requirements
- [ ] Fork and modify open source sites to meet requirements
- [ ] Allow Federalist users to fork a starter template to their own account

dhcole:
Initial thoughts for template requirements:

- 508 compliant
- DAP code
- Prose configuration

dhcole:
Initial template ideas

- Basic "brochure" site for new department, agency
- API doc site
- Interactive report site (like http://www.shopify.com/2013)

jeremiak:
From what you know of the potential first users (or other static sites that 18F has done so far) is there a need for a blog-like site with feeds such as RSS/Atom?

jeremiak:
Also, just because I'm still relatively uninitiated, what is a "DAP code"? I assume its analytics but does it function in a manner similar to Google Analytics?

dhcole:
Yea, sorry, we should [ban acronyms](https://twitter.com/collision/status/602950284864692224). DAP is the [Digital Analytics Program](https://www.digitalgov.gov/services/dap/). It runs a government-wide Google Analytics account that feeds https://analytics.usa.gov/. 

Federalist should be a steward of best practices for government websites, so we should just bake things like this (and correctly configured HTTPS) into the templates and hosting.

dhcole:
> From what you know of the potential first users (or other static sites that 18F has done so far) is there a need for a blog-like site with feeds such as RSS/Atom?

Good question. @gboone @meiqimichelle @shawnbot  @melodykramer @ultrasaurus @blacktm @gbinal: care to weigh in on what you'd want to see out of Federalist templates?

dhcole:
There are some interesting ideas here: https://github.com/mdo/jekyll-snippets

meiqimichelle:
None of the sites I've worked on so far have been blog-like.....except our own 18F blog. In my world, I don't see a use for the in the near future. Let's see what the others have :)

blacktm:
[...]

PR #55, opened 2015-06-13
Create sites from site templates

### LINKED ISSUE #53: Have views for template creation
Users need to be able to select templates as well as add github repos. Since we'll have a small number of templates to start we need to have a view for the user to select from.

jeremiak:
This should be merged after #52 as it was cut from that branch.

This PR is pretty great, it lets the user create repos from our "templates" from the front-end. The user flow is to select one of our templates (easily configured in `sails.config.templates`) and then the API will fork that repo and add it as a `Site` model so that it gets a webhook registered. The error handling is pretty poor, basically the callbacks just pass the err and success objects all the way back up the chain to the client. I'm sure we can make this much better in the future.

This PR should close #53 #54 and IMHO #12 as it is the final piece in the prototype being mostly functional.


dhcole:
This is awesome. Love how fast it is. From nothing to new site in under 90 seconds: https://www.dropbox.com/sh/28fjlprnt9ju6oe/AADQ8jv9QdOzVonaqa6MsCdFa?dl=0

Just added two commits:
- No need to define the route since Sails will automatically create a route in the form of `/[prefix]/[controller]/[method]` (The custom route is there for the webhook so it's not versioned with the prefix...seemed more stable but that could be changed too)
- Refactored the `forkRepository` method for the GitHub service to make it independent. The convention with Sails services is that they should be able to be called at any point in the application, not just as part of the request flow. So this removes `req` and `res` as arguments.

@jeremiak please take a look when you have a chance. If you're :+1: feel free to :ship:

jeremiak:
:tada:

PR #56, opened 2015-06-14
Escape HTML in templates

### LINKED ISSUE #42: Make sure templates escape user-generated input
Anything that comes from a user should be escaped to avoid XSS vulnerability. In general, I would escape every template variable unless it actually needs to include rendered HTML. If it does need to include HTML, then we need to make sure that HTML is escaped when it's created, either by passing through something like a markdown converter that sanitizes input or by escaping the variables that form that HTML.

Example where we can escape: https://github.com/18F/federalist/blob/master/assets/app/templates/SiteListItemTemplate.html#L1-L5

> If you wish to interpolate a value, and have it be HTML-escaped, use <%- … %>

[underscore.js](http://underscorejs.org/#template) and lodash work the same way. We're not using ejs, but in case we do, ejs reverses the `-` and `=`

jeremiak:
This uses Underscores `<%-` operator to escape the HTML used in templates. This touched the add new site template, the login button/view template, and the individual list item view template.

To close #42

### DIFF
--- a/assets/app/templates/AddSiteTemplate.html
+++ b/assets/app/templates/AddSiteTemplate.html
@@ -33,10 +33,10 @@
   <section>
     <h5>Or add your own Github repository</h5>
     <form>
-      <input type="hidden" name="owner" value="<%= user.username %>"></input>
+      <input type="hidden" name="owner" value="<%- user.username %>"></input>
       </br >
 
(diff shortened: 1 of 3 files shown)


dhcole:
:+1:

Issue #49, opened 2015-06-12
Make sure API is secure


dhcole:
To prepare for hosting outside of local development, we need to ensure the following:

- [x] API access permissions set
- [ ] Users can only access data related to their accounts
- [x] lock down who can sign up
- [x] handle Jekyll and Hugo plugins

PR #62, opened 2015-06-16
Restrict users to a GitHub organization

dhcole:
This PR validates each new user and returning user before authentication to ensure they are a member of an approved GitHub organization. 

Approved GitHub organizations are specified in `config/passport.js`, `github.organizations` as an array of organizations IDs. These IDs are numerical — not the organization name — in case the organization changes its name. You can find these IDs through the API, or just look at the path for the organization's avatar image:

screen shot 2015-06-16 at 1 40 37 pm

This PR also removes unused routes (for displaying authentication forms as views) and makes sure authentication errors are passed through to the client as an `error` query string parameter.

Refs #49

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -62,14 +62,19 @@ var AuthController = {
       // Only certain error messages are returned via req.flash('error', someError)
       // because we shouldn't expose internal authorization errors to the user.
       // We do return a generic error and the original request body.
-      var flashError = req.flash('error')[0];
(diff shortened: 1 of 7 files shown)


dhcole:
Also, for reference, the 18F org is `6233994`

jeremiak:
Runs like a charm. I don't have another user not in the 18F org to test that it doesn't get in but the positive case works and the code makes sense so :ship:

dhcole:
I tested with a new account. Thanks for verifying

PR #64, opened 2015-06-17
Update list on site delete

### LINKED ISSUE #63: Site list view doesn't refresh after site is deleted
To reproduce, click delete on a site. The page doesn't change, but the delete request is sent successfully. Hit refresh, and the site is gone.

jeremiak:
To close #63 this has the list view listen for the `update` event instead of the `sync` on the collection of sites.


dhcole:
:+1:

PR #65, opened 2015-06-17
Improve access policies

### LINKED ISSUE #49: Make sure API is secure

dhcole:
This PR refactors the access policies and makes related changes, including the following:

- Ensure that requests have an ID or parentID (for looking up associated models) and that it matches the requesting user
- If the request is a "find" action, for listings, redirect it to the associated listing off of the user API so that it's filtered according to the current user (`/v0/site => /v0/user/1/sites`)
- Redirect user listing requests to the logged in user's endpoint (`/v0/user => /v0/user/1`)
- Enforce model schemas (avoids default ORM behavior of saving any arbitrary value set on a model)
- Update the front-end to set up a `users` association with new sites
- Update the fork API hook to do the same
- Turn off Sails blueprint shortcut routes
- Set up webhooks with an explicit user instead of depending on one from the site model (needed since sites can have multiple users; by default it pulls the first user, who is the one who created the site)
- Minor dev improvements

Resolves #49


jeremiak:
This works great. Merging and opened #67 for a resulting front end task to relay the error to the user

PR #68, opened 2015-06-17
Fix fork route policy

dhcole:
The previous policy work was blocking `/v0/site/fork`. This fixes that and handles some possible errors with a missing or invalid `templateID`


jeremiak:
Ha, I was just making some changes and ran into this problem, thanks for the quick fix!!

PR #69, opened 2015-06-17
Set up Travis for CI

dhcole:
Adds a Travis config file and removes the step to install Jekyll automatically.

Also tried to remove Grunt but backed out because we are using Sail's default static file serving from `./tmp` which depends on Grunt. We'll revisit when we break out the client app.


jeremiak:
You can check out our shiny new Travis instance at [travis-ci.org/18F/federalist](travis-ci.org/18F/federalist)

Issue #66, opened 2015-06-17
POC 1 Overall Design Critique


bradnunnally:
Here is a list of items that could be improved for POC 1

- [x] Smaller color palette - limit to 2 to 3 primary colors. pink background should be replaced with something light. E.g. gray
- [x] Balanced font usage - avoid using all caps, strengthen user account area
- [x] Consolidate new site button into single call to action
- [x] Repo name on "Add your own" give it a border. Doesn't feel like a text box

jeremiak:
Cool, agreed! A few followups:

* Balanced font usage - avoid using all caps, strengthen user account area - I don't understand what "strengthen" means in this context. Could you give me a few examples of what you mean?
* Consolidate new site button into single call to action - what do you think of the green button when there are no sites in the list but the red plus button when there are? This way there will only be one present at a time on the screen but they will look like two different features.

bradnunnally:
Hey @jeremiak, The balance between the title "Federalist" and the user account area is off. It's hard to notice/determine that user account area is there and what you can do with it. It's mostly a balance thing. 

New site button - Yes! I like the picture you describe with the different states.

jeremiak:
@bradnunnally cool what should we do about the balance? we could have like a user avatar/profile pic up in that corner so you know its the "user actions" corner?

jeremiak:
Closing this one out

PR #70, opened 2015-06-17
Design feedback implementations

jeremiak:
This PR addresses the design feedback that @bradnunnally gave in #66 and implements the first iteration on it.

Specifically, it:

0. Changes the font balance, including no capitals
0. Simplifies the color palate because pink was a bad choice
0. Adds some border to text inputs so you know they are there
0. Only shows one add new site button at a time

@bradnunnally since it was your feedback, do you want to go ahead and review this PR and if it looks good then merge?


bradnunnally:
Only piece of feedback is the buttons for Use This Template. The foreground and background doesn't have enough contrast. Perhaps just go with white text?

jeremiak:
Cool, merging for @bradnunnally since he's on his way to the airport

PR #76, opened 2015-06-19
Cloud Foundry Settings

### LINKED ISSUE #51: Host on CloudFoundry
Move the application to CloudFoundry for shared access / demos

Depends on:

- #49 
- #50

dhcole:
Some settings to host on Cloud Foundry. 

Closes #51


jeremiak:
Just to confirm, this is our departure from using `local.js`, correct?

I just pulled down the branch and it looks like Sails no longer uses `local.js` to get the Github OAuth vars set up correctly.

dhcole:
@jeremiak local.js will still override these settings, so you can use it for your local environment. But CF handles all secure settings through ENV vars, so this just defaults anything we need to set to reading it from a `process.env`.

jeremiak:
Gotcha, @dhcole.

So I think there's a bug then (probably just an overlooked local var) which happens when I am routed back to the app from Github. I am ultimately routed to `http://localhost:1337/auth/not_set` with a `code` query string parameter. My application on Github is set to have the callback URL be `http://localhost:1337/auth`. Is this from the `callbackUrl` not being included in the `local.js` file?

dhcole:
Yea, the callback url should be specified as `callbackURL: 'http://localhost:1337/auth/github/callback'`, and that can be done either in `local.js` under `passport.github.options` or as an env var. 

Previously, we were letting sails set it, but that only works with localhost. So we need to define it explicitly to get it to work with a public domain.

jeremiak:
Great, just added that bit of documentation to the README since it could pose a problem for new people getting started on the project. Otherwise this LGTM

jeremiak:
The build failure looks like the result of the tests timing out, which makes me think this branch was cut from master before we fixed that bit. Should I just merge master into this branch and continue from there?

dhcole:
I just reran the build. I think you're right. Shouldn't be an issue after we merge (unless Travis needs the timeout value set somewhere else).

Issue #71, opened 2015-06-17
New filtered sites listing doesn't populate builds


dhcole:
Sails isn't populating two levels of associations (user.sites[].builds), so we don't have access to builds in the front end. We could request the builds for each site with`/v0/site/1/builds`, all builds for the user with `v0/build`, or just use the data in the user response, which includes the builds and sites (this would eliminate additional ajax calls but breaks the url pattern a bit...

dhcole:
Looks like we're not the only ones with this issue... https://github.com/balderdashy/waterline/issues/308

dhcole:
@jeremiak any experience with http://backbonerelational.org/#RelationalModel? This might be a solution...

PR #79, opened 2015-06-22
Request build from its own API endpoint

dhcole:
Quick fix for #71


dhcole:
@jeremiak would love a quick review of this when you have a chance. Seems like splitting the difference between trying to read from the user model vs setting up a completely new build collection.

jeremiak:
Taking a look right now

jeremiak:
Cool, I like this!

PR #80, opened 2015-06-23
Create a build upon adding a site

### LINKED ISSUE #77: New sites should build on afterCreate
So there's a hosted site without needing to edit it.

jeremiak:
Uses the Sails lifecycle callbacks to create a related `Build` model immediately after saving a new `Site` from the user. This has the result of the user being able to see their site without having to edit it.

To close #77 - @dhcole can you take a look when you get a chance?


jeremiak:
Per our conversation on Friday, I think this is ready for a looksie @dhcole

dhcole:
Looks great!

PR #81, opened 2015-06-23
More CloudFoundry Configuration

### LINKED ISSUE #50: Set up to use external database
Optionally connect to external database. Set up RDS instance for staging environment.

### LINKED ISSUE #78: Need to set up Jekyll on the CF server

dhcole:
Some more configuration changes to get the app working well in CloudFoundry:

- use `cfEnv` if available in production to set up the database with the CF service db
- serve static files directly from `/assets` instead of from a copy in `.tmp/public`
- implements [sails-db-migrate]() for database migrations, including initial schema and migration
- uses both the node.js and ruby buildbacks and builds Federalist with `npm` and Jekyll with `bundler` via the Gemfile

Closes #78 
Closes #50


dhcole:
Some notes on using postgres with Federalist:

1. make sure you have postgres installed and that it's running (http://www.postgresql.org/download/)
2. if you don't already have a database, create one: `$ createdb federalist`
3. in config/local.js, add the following:
    ```js
    connections: {
      postgres: {
        adapter: 'sails-postgresql',
        database: 'federalist'
      }
    },
    models: {
      connection: 'postgres'
    }
    ```

jeremiak:
This looks great to me!

Issue #11, opened 2015-06-05
Figure out routing for main branch vs previews


jeremiak:
Pulling from [code comment](https://github.com/18F/federalist/blob/master/api/services/buildEngines.js#L112) but also related to #10 - right now the sites are built under a directory named for the "owner" and then in another directory named for the branch.

How should the routing work for this when people are trying to see their sites?

dhcole:
Made good progress on this in #91. Just two open tasks to complete. 

@ultrasaurus: we should be able to ship this feature early next week.

jeremiak:
#91 got merged and this is done!

PR #86, opened 2015-07-01
S3 File Sync

dhcole:
If `build.s3Bucket` is preset in the `sails.config`, sync built sites to the given S3 bucket. Uses the standard AWS SDK, so authentication may be handled a few different ways: http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html#Setting_AWS_Credentials

Some follow-up tasks that can be separate:

- Use S3 urls in the UI instead of `/assets/site`
- gzip files before syncing to S3 (this is partially implemented but needs better testing)


dhcole:
Updated to gzip compress files as needed. Also, this uses selective sync where only files that have changed are synced to S3 (this makes it really fast).

We could go even further by using `git pull` instead of `git clone` to only pull down changed source files, but that would require a shared temporary storage across all web servers. So not worth doing now.

dhcole:
> Use S3 urls in the UI instead of /assets/site 

Can be addressed during #11

jeremiak:
Looks great to me, files are pushed successfully to S3 bucket

PR #87, opened 2015-07-02
Set collection on new site model to fix error for missing builds.

### LINKED ISSUE #85: Add existing repository as site fails
Steps to reproduce:

0. Click add site
0. Attempt to enter an existing repository in the "Or add your own Github repository" section
0. App gives no feedback but console errors:
```
Uncaught TypeError: Cannot read property 'builds' of undefined
```

dhcole:
Also triggers new site process on submit instead of click so pressing enter works

Fixes #85

### DIFF
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -42,11 +42,11 @@ module.exports = {
   },
 
   afterCreate: function(model, done) {
-    User.findOne({username: model.owner})
-        .exec(function(err, user){
+    Site.findOne({id: model.id }).populate('users')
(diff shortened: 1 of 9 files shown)


dhcole:
Updated to return builds sorted by most recent and to run the initial build based on the creating user, not the owner, so that users can add sites they have access to but don't own (like those owned by an organization. Also updated the new site form to allow specifying site owner and default branch.

jeremiak:
This works great, I like the new form where you can add a different user/org's repo.

Issue #83, opened 2015-06-29
Single page editor based on Sir Trevor


jeremiak:
For this next milestone we'll want a single page editor that we can start to use in place of Prose. This ticket is completed when:

* We have a single page editor that is passed a Github URL to pull the original content from
* Use Sir Trevor to enable authors to edit the page content
* Spit out the new markdown to somewhere (could be a console log or w/e for now) so that we can verify the new markdown respects front matter and content updates.

dhcole:
Done in #90

PR #90, opened 2015-07-07
[Ready for review] Pull/new editor

jeremiak:
This is the majority of the first implementation of the editor for #83 and this PR has the following user facing features:

0. The ability to edit pages from within Federalist through a GitHub URL. The editor is based on a library called [Sir Trevor](https://github.com/madebymany/sir-trevor-js)

Technically, this PR has the following features:

0. A main view to manage switching out the content views
0. A few new separate views to make things a bit more modular. For example, the unauthenticated view is now pulled into its own view
0. A whole new model to represent how we move content in and out of the Sir Trevor blocks

A few last things before this is ready for testing, but I'd love to get feedback on the editor itself from @GailSwanson to see what we need to get in there.

Last things:
- [x] A bug where the authenticated view always loads unless authd variable is forcibly set to true
- [x] Heading block types (right now all h* are one block type)
- [x] Finish up the way a list block renders itself
- [ ] ~~Implement image blocks~~ (Not worth holding up this PR since it requires server components)
- [x] Implement code blocks


dhcole:
Love this. Can we make the edit link configurable between this or prose? For users who want to on-board now-ish, they may prefer prose while we build this out. But it looks great to me otherwise. Really solid foundation.

jeremiak:
Cool, really like that idea. Made a quick modification to include a prose.io link for now

jeremiak:
@dhcole I think this is finally ready for a look-over.

Todos but shouldn't prevent this from being merged:

0. Do something with the markdown, don't just log to console
0. Tests? We don't have any of those for the Backbone application at this point and it makes me kind of worried <- I realize I'm the one writing the Backbone code mostly so I should just start including tests with these things

dhcole:
Looks great. Just two things:

1. I'm seeing bullets for `li` elements, and I think it's because of https://github.com/18F/federalist/pull/90/files#diff-2a18822445dbc9ad010e79be05b23a07R28

    screen shot 2015-07-17 at 3 19 47 pm

2. I think we should keep the main edit link as a link to Prose and have this editor as a hidden feature until we wire up posting content back to the server and a file listing view.

If you're on board with these, we can get this merged in and figure out the next step.

dhcole:
Nice work on this! :rocket:

PR #91, opened 2015-07-10
Build default and preview sites separately based on branch

dhcole:
This sets up separating publishing processes for the main site (when building from the default branch) and draft preview sites (any other branch).

By default, these will publish to `/site/[owner]/[repo]/` for main sites and `/preview/[owner]/[repo]/[branch]/` for preview sites.

The same pattern applies when publishing to S3. To restrict public access to main sites only and not preview sites, we're using the following bucket policy:

```json
{
	"Version": "2012-10-17",
	"Statement": [
		{
			"Sid": "PublicReadGetObject",
			"Effect": "Allow",
			"Principal": "*",
			"Action": "s3:GetObject",
			"Resource": "arn:aws:s3:::federalist.18f.gov/site/*"
		}
	]
}
```

Next steps:

- [x] proxy preview sites from S3 with an authentication policy that checks to make sure the user has access to the site
- [x] add links to the published site and drafts to the user interface

### DIFF
--- a/api/controllers/PreviewController.js
+++ b/api/controllers/PreviewController.js
@@ -18,6 +18,9 @@ module.exports = {
           res.set(headers);
         }),
         stream = object.createReadStream().on('error', function(error) {
+          var file = key.split('/').pop().indexOf('.') !== -1,
+              notFound = error.statusCode === 404;
+          if (!file && notFound) return res.redirect(req.path + '/');
           res.send(error.statusCode, error.message);
         }).pipe(res);
   }
--- a/api/controllers/PreviewController.js
+++ b/api/controllers/PreviewController.js
@@ -1,0 +1,25 @@
+var AWS = require('aws-sdk'),
+    s3 = new AWS.S3();
+
+module.exports = {
+
+  /*
+   * Proxies requires from S3 for so those requests can be authenticated
+   *
+   */
+  proxy: function(req, res) {
+    var key = req.path.slice(1);
+    if ((key).slice(-1) === '/') key = key + 'index.html';
+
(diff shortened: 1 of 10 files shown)


dhcole:
@jeremiak: should be ready now. Would love your :eyes: on this.

I added a simple list of recent builds:

screen shot 2015-07-14 at 1 52 07 am

This is also on `prod` with S3 configured.

jeremiak:
This is a great feature! A few things I've noticed that we might want to improve before this PR is merged.

0. The "Recent branches" is awesome except its actually just recent builds that aren't on the `defaultBranch`. But we don't save access to each individual build; when a new commit comes in on a branch we wipe away anything we have currently for that branch and put in the new build. Therefore, I think the recent branches should just display one link per branch name, as opposed to each build. Here is what my series of changes to the `yooo` branch look like:


0. This works great for us as a team but it requires anybody new to the project to understand how S3 works, set up proper keys **and** get a tunnel open on their local machine, all before they can even get the project to run correctly. Can/should we alter the `previewController` to serve up preview files from the local build destination in the case that the user hasn't configured the app to use S3?

dhcole:
@jeremiak good call. Fixed both. Want to check again / merge? Particularly love how easy this was: https://github.com/18F/federalist/commit/91d39f31704760c6e636926927fac17e07ecd382#diff-7da801ca771ec6dbe6443c553554cf38R12

jeremiak:
This is super cool! Works great so I'm going to merge it

dhcole:
:+1: thanks for the review. I'll deploy it

PR #95, opened 2015-07-14
Match npm commands to other concurrent projects

jeremiak:
Right now both @dhcole and I are working on this project as well as [midas](https://github.com/18f/midas) which are both JavaScript projects using Sails and Backbone+browserify. Yet, they have slightly different `npm` commands for starting the dev environment which was driving me crazy.

I've modified the Federalist `package.json` to use the same command so that `npm run watch` executes the same build processes on both. I've also updated the documentation.

@dhcole what do you think? By no means necessary, but is a bit nicer to have the same interface on both projects.


dhcole:
Good call.

PR #96, opened 2015-07-14
Adds security headers to all requests

dhcole:
Per @NoahKunin's security review:

screen shot 2015-07-14 at 5 22 33 pm

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -37,6 +37,7 @@ module.exports.http = {
 
   order: [
     'forceSSL',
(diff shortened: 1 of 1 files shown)


NoahKunin:
Thanks! Looks great. We should probably get a secure node.js baseline into FISMA Ready sooner than alter...

dhcole:
Happy to work on that. Probably best approach would be an express
middleware that we can add to any project. Are there other controls besides
these headers that it should include?

Issue #92, opened 2015-07-10
Document env vars


dhcole:
We're using some `env` vars for configuration (mostly in `config/build.js`. These should be documented.

PR #100, opened 2015-07-15
Quick stab at updating docs with env vars

jeremiak:
I took a quick stab at documenting our environment variables per #92 - I think I capture all of them, but please let me know if I missed a few or if this section of the docs are confusing.


dhcole:
Looks great. I added two more. Thanks for taking the time to capture these.

Issue #97, opened 2015-07-15
Windows support proposal


anweiss:
Proposing platform support for Windows. Thoughts as follows:

- Updated `README` to include instructions for developing on Windows
  - Required dependencies
  - `npm dev` script that also supports Windows
  - Workaround for Sails issue [here](https://github.com/balderdashy/sails/issues/2884)
- `templateBuilder` service which builds *run* commands based on detected OS

Would be willing to submit PRs for proof-of-concept phase, but could also hold off depending on the current state of the development roadmap

jeremiak:
Hey @anweiss, thanks for the issues. A couple of questions (pardon my ignorance but I have never developed JS on a Windows machine):

0. What would change about the dependencies?
0. Consequently, which scripts would need to change for Windows compatibility?
0. Is the Sails workaround required even for IOJS? I have been using iojs-v1.6.3 to develop the application on my machine, not quite sure @dhcole is using.
0. Can you go into more detail about what you mean by the `templateBuilder` service and how you envision it working?

Of course, please feel free to illustrate these things with PRs, even for proof-of-concept ideas, if you think it would be more illustrative. What questions do you have about the current roadmap? Have you seen [the wiki page](https://github.com/18F/federalist/wiki/Development-Roadmap)?

jeremiak:
I would also be really interested in hearing about how Microsoft is thinking about using this tool or at least what y'all are evaluating it for. That might help us understand the nature of these (#97, #98) issues

anweiss:
Certainly! Some thoughts below:

1. Contributing to Federalist as a developer on Windows requires Ruby, Ruby devkit, Python 2.7, and Visual Studio 2013+ in order to install Sails and its dependencies
  - Would want to include links to instructions for installing these components on Windows
2. The `dev` npm script does not seem to properly concatenate the `watch` and `watchify` scripts on Windows...not sure if it was an issue on my end but would need to dig a bit further
3. I have yet to try io.js on Windows with Sails so would need to look in to that as well
4. I've started to stub out some of the *run* command templates for Windows on this branch [here](https://github.com/Microsoft/federalist/blob/pull/windows/api/services/buildEngines.js) and will submit a PR once I get it cleaned up. Once we start having multiple *run* command templates to support Windows and Linux/Mac, it starts to duplicate code in `buildEngines.js`; hence a proposed `templateBuilder.js` service. But will go whichever route you guys decide on.

[...]

PR #103, opened 2015-07-20
Initial buildEngine support for Windows

anweiss:
In reference to #97. Includes necessary updates for running build commands on Windows. A section for developing on Windows has also been added to the `README`. There are a few items that may still need to be tweaked, but this should be a decent start.

Adding @chadwach for comments

### DIFF
--- a/api/services/buildEngines.js
+++ b/api/services/buildEngines.js
@@ -1,65 +1,112 @@
+/// <reference path="../../typings/node/node.d.ts"/>
+
 var exec = require('child_process').exec;
+var path = require('path');
+var isWindows = /^win/.test(process.platform);
 
 /**
  * A service to managing build processes. Each engine gets its own method,
  * which takes a model, runs a shell process, and then returns the model
  * and any error message.
+ * 
+ * NOTE: On Windows, run commands should include a check for directory
+ * existence prior to RMDIR execution but IF EXIST statements don't concatenate
+ * in one line with '&&' which results in 'The system cannot find the path specified' error.
  */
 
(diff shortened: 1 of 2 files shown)


api/services/buildEngines.js
```diff
@@ -84,7 +131,8 @@ module.exports = {
           branchURL: defaultBranch ? '' : '/' + model.branch,
           root: defaultBranch ? 'site' : 'preview'
         },
-        template = _.template(cmd.join(' && '));
+        // Temporary until workaround for single line IF EXIST logic is implemented
```
dhcole: @anweiss can you say more about what you mean by this comment?

anweiss: Sure thing. So currently the code concatenates each of the run commands in the array into a single line with each command separated by an `&&`. On Windows, the `RMDIR` command cannot check the existence of a directory before it exists so if it tries to delete a non-existing directory, it throws an error message. If you concatenate the commands with `&` on Windows, it will run each command regardless of whether or not the previous command completes successfully which is what `&&` imposes.

To prevent `RMDIR` from failing on non-existent directories, one could execute an `IF EXIST <directory> RMDIR <directory>' command. However, Windows does *not* support concatenating this command with other commands on the same line as follows:

```
IF EXIST <dir> RMDIR <dir> && MKDIR <newdir>
```

dhcole:
@anweiss Looks like we've got a new approach in #115 

Are you ok with closing this and #114 in favor of a separate module?

anweiss:
Sounds good.

PR #105, opened 2015-07-22
Allow non-authorized users to trigger builds on public sites

dhcole:
Previously, Federalist tried to clone a site on build with the oauth token from the user who pushed the change. If that user hadn't authorized through federalist, and therefore had no oauth token, builds would fail. Builds would also fail because `build.user` is required, and those users wouldn't have a user record in the database.

This fixes the problem by doing two things:

- Creates a new user if one is missing when creating a build.
- Builds a site without a token if none is available

For public repos, this will work. For private repos, any user wanting to trigger a build of the site will need to authorize their account with Federalist for builds to work. Errors from unauthorized access attempts on private repos will be logged to the `build.error` field of the build record. Eventually, we could send emails to users letting them know their build failed because they need to authorize with a link to do so.

### DIFF
--- a/api/controllers/WebhookController.js
+++ b/api/controllers/WebhookController.js
@@ -36,9 +36,8 @@ module.exports = {
 
       // Find a matching user
       user: function(next) {
-        User.findOne({
-          username: payload.sender.login
-        }, next);
+        var record = { username: payload.sender.login };
+        User.findOrCreate(record, record, next);
       },
 
       // Find a matching site
--- a/api/services/buildEngines.js
+++ b/api/services/buildEngines.js
@@ -104,7 +104,8 @@ module.exports = {
       // Set populated token values
       tokens.repository = model.site.repository;
       tokens.owner = model.site.owner;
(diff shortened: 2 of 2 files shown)


dhcole:
@jeremiak read for your review

jeremiak:
This works pretty well! It is great that a public site will now be cloned, built and deployed regardless of the committer, but I think we've officially hit the end of the road for our `/v0/user/:id/builds` hack.

If I own a site called `microsite-template` and a contributor on Github makes a commit the site will build (which is great). However, when I look at my site listings, those builds won't show up because they're not made by my user. In this way, the latest build text on the site listings is pretty out of date and confusing.

I think a `Site` will have to return its builds regardless of the user who they are associated with.

dhcole:
> I think a Site will have to return its builds regardless of the user who they are associated with.

Agreed.

dhcole:
@jeremiak given this PR makes the build notification no worse and stops the server error, is it worth merging this as is and following up with a change to the build listing endpoint?

jeremiak:
I'm OK with that @dhcole!

PR #106, opened 2015-07-23
Fix missing collection error

### LINKED ISSUE #101: Uncaught TypeError: Cannot read property 'builds' of undefined
Not sure if #85 has regressed, but experiencing this issue again related to [this](https://github.com/18F/federalist/blob/master/assets/app/models/Site.js#L16) line in `assets/app/models/Site.js`. Occurs when clicking "Submit Repository-Based Site".

I'm also getting the error below which could be related since `this.collection` [here](https://github.com/18F/federalist/blob/master/assets/app/models/Site.js#L27) is undefined.

```
http://localhost:1337/v0/user/1/builds Failed to load resource: the server responded with a status of 404 (Not Found)
```

dhcole:
Each new site model needs to know it's collection. This sets that collection to the view, so it can set it on the new site. Fixes #101

### DIFF
--- a/assets/app/views/AddSiteView.js
+++ b/assets/app/views/AddSiteView.js
@@ -51,6 +51,7 @@ var AddSiteView = Backbone.View.extend({
     });
   },
   onSuccess: function onSuccess(e) {
+    this.collection.add(e);
     console.log('winning', e);
     this.trigger('site:save:success');
   },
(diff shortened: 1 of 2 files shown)


dhcole:
Also added in a commit to smooth out the redirect and rerender process after adding a new site.

jeremiak:
This looks great

PR #109, opened 2015-07-24
Refactor filter policy / remove proxy filter

### LINKED ISSUE #107: Site listings only show builds for the authenticated user, not the site
Right now the site listings only show the builds for the currently authenticated user. This looks like a bug from the user's perspective because there could be a more recent build than the one listed in the list view.

To illustrate let's say that I have a public repository that has two contributors, one of whom has authorized Federalist and one who has not. #105 enabled the site to still be pulled, built and deployed even if the second (non-Federalist) user made a commit to the repository on Github. However, the resulting build will be attributed to this second user. When I load up my list of sites, that build won't show as the most recent build on a given site because the builds are retrieved based on a user through `/v0/user/:id/build`.

Instead, the builds should be retrieved based on the site. That is to say that my site should inform me when the latest build was, not the latest build by my user.

dhcole:
- Refactors the CurrentUser policy to filter lists by models associated with the requesting user, rather than redirecting to an API call off of the current user. Fixes #107 
- Removes the PreviewProxy policy (for now) since it was not allowing multiple users to preview the same site
- Small formatting changes for the list item view, such as using Moment for better date formatting and opening the Prose edit link in a new tab


jeremiak:
:shipit:

PR #112, opened 2015-07-27
Add support for custom domains and site configuration

### LINKED ISSUE #93: Support custom template variables
Per a conversation with @ultrasaurus, it would be useful to have a way to store key-value pairs in Federalist that are available to a site when rendering its templates. For instance, an API key could be stored and made available when the site is rendered. 

It would be helpful if these variables were also available based on the branch, so draft branches could have access to different variables than published sites.

Currently, we're not supporting arbitrary plugins because of the security implications of running them on the build server. We hope to resolve this at a later time by running the build process in a safe container.

In the mean time, custom variables could be set in the `_config.yml`, which would make them available under the `site` object in Liquid templates. The problem with that approach is that some data may be sensitive and shouldn't be committed to a public repository.

Thinking through this, it seems like one solution is to leverage Jekyll's ability to combine multiple `_config.yml` files. We could expose a form field to the user to set arbitrary `yaml` data per branch that would be printed to a separate `_config` file during the build process, which would make that data available under the `site` object and avoid the need for a plugin.

@ultrasaurus does this sound like it would address your need?

- Enter `yaml` data in the Federalist interface with the ability to specify to which branch the data should apply
- That data is rendered as an additional `_config` file making it available through `site`
- Access the data in Liquid templates through `site.key_name`

Open question as to how securely we would want to treat this data. Perhaps it should be encrypted in the database.

cc @jeremiak

### LINKED ISSUE #94: Add a site settings view
A site settings view would allow users to change per-site settings like the default branch, the rendering engine, potentially `_config` data (#93), and likely more as we continue to build.

dhcole:
Support for custom domains and site configuration data by adding additional data to the Site model and presenting it to users with a site edit view.

Setting custom configuration works by adding the `site.config` data to a `_config.yml` file used while rendering the Jekyll site.

Open ends:

- [x] Migration script for new site model attribute
- [x] display the origin url on the UI for custom domains
- [x] write detailed instructions for configuring a custom domain
- [x] update the “view site” link on the site list item view
- [x] adjust `baseurl` to use `site.domain`

screen shot 2015-07-27 at 3 55 27 pm

Closes #94 
Closes #93

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
-
+ruby '2.2.2'
 gem 'github-pages'
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -10,7 +10,7 @@ module.exports = {
   schema: true,
 
   attributes: {
-    // The name of the GitHub user or organization that owns the site's repository 
+    // The name of the GitHub user or organization that owns the site's repository
     owner: {
       type:'string',
       required: true
@@ -38,6 +38,12 @@ module.exports = {
       collection: 'build',
       via: 'site'
     },
+    domain: {
+      type: 'string'
+    },
+    config: {
+      type: 'string'
+    },
     toJSON: function() {
       var obj = this.toObject(),
           config = sails.config.build || {};
@@ -63,6 +69,10 @@ module.exports = {
           };
           Build.create(build, done);
         });
+  },
+
+  afterUpdate: function() {
+    Site.afterCreate.apply(this, _.toArray(arguments));
   },
 
   registerSite: function(values, done) {
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -62,6 +62,7 @@ module.exports = {
     Site.findOne({id: model.id }).populate('users')
         .exec(function(err, site) {
           if (err) return done(err);
+          if (!site.users[0]) return done();
           var build = {
               user: site.users[0].id,
               site: model.id,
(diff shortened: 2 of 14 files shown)


dhcole:
@jeremiak over to you for review. I just need to write a wiki page for how to set up the custom domain in AWS.

You can test by adding in some yaml to the site config field and changing the custom domain to some other URL. After doing so, you should see the site build with no baseurl and with the custom yaml you entered. Also the link to view the site on the listing page should change to your custom domain. Draft branches should be unaffected by the custom domain but should get the config settings.

dhcole:
Added a link to AWS's docs for setting up the domain.

jeremiak:
The custom YAML works well, that's pretty cool!

I'm confused about what the desired outcome of the custom URL change is. I understand that the custom URL should replace any baseurl in the templates, but I'm not sure what the correct behavior is for the view link on the dashboard.

I changed one of my sites to have a custom domain of `jeremiak.com` and when I clicked on the view link I was taken to `localhost:1337/jeremiak.com` which was a blank page. My expected behavior of that link was to take you to a preview of the site (perhaps with the custom URL in the path).

Here is a screenshot of what happens _after_ I clicked on the view link: http://up.jeremiak.com/c8NR

dhcole:
Good usability testing :)

The custom domain is supposed to have a full url including the protocol, as formatted in the placeholder example. But there's no validation, so this can happen. I suppose there should be some validation / auto-formatting to handle this. 

If you have a full domain + protocol, you'll just link to the main site for the "view site" link.

The reason for all of this is we need to clear out the baseurl so that links on the site will work at a root domain. That makes the site display correctly on the custom domain, but when you look at the origin url (`s3-bucket/owner/repo/`) the site will be broken since it will have been rendered without a baseurl setting to account for the subdirectories. 

To avoid linking to a broken view of the site, I just have the view site button take you to the custom domain. Since that's where the site is officially hosted, this seems like the correct behavior.

jeremiak:
:ship: and we can add the validation step later

Issue #104, opened 2015-07-21
Separate BuildEngine service as a separate module


dhcole:
To support #103 and make `BuildEngine` something that could be run on its own worker process, we should package this service into its own npm module. Then #103 can be repackaged as a separate module too and included in configuration with something like `buildEngine: require('federalist-build-engine-windows')`.

dhcole:
Looks like making this an installable hook is a good way to go: http://sailsjs.org/documentation/concepts/extending-sails/hooks/installable-hooks

dhcole:
Closed by #115

Issue #98, opened 2015-07-15
Azure support proposal


anweiss:
Proposing support for Azure. Immediate roadmap indicates AWS S3 for static content hosting but there could be an opportunity to get Azure included pending the current state of the roadmap. Some thoughts:

- Develop an `Azure` service with stubs similar to those of the `S3` service for static content hosting
  - Provisions Azure Web App
  - Pushes content to Azure Web App with Git
- Include Azure configuration options in `sails.config.build.azure`
- UI support for establishing connectivity to AWS, Azure or both
- Include one-click provisioning support to Azure for hosting the Federalist platform itself for development and production purposes

anweiss:
Tracking in [pull/azure](https://github.com/Microsoft/federalist/tree/pull/azure) branch. Will re-work pending the creation of a new buildEngine installable hook per #104

dhcole:
Closing in favor of #121

PR #114, opened 2015-07-30
Initial support for Azure site publishing endpoint

anweiss:
In reference to #97 and #98, this provides initial support for Azure site publishing. This PR is for tracking purposes and simply provides a starting point for enabling Azure support. Completely open to feedback and recommendations for ensuring that this fits in to the Federalist platform per the roadmap and Sailsjs best practice. This will be re-worked to support #104 once an installable hook has been developed. As such, it does not yet integrate with `buildEngine`.

The process by which the Azure service publishes a new static site is as follows:

1. Obtain Azure security token from user credential
2. Check for existing [Azure Web App](http://azure.microsoft.com/en-us/services/app-service/web/)
3. If Web App does not exist from a previous deployment, it creates a new [Azure Resource Group](https://azure.microsoft.com/en-us/documentation/articles/resource-group-overview/)
4. Deploy template to new Resource Group which provisions a new Azure Web App within that Resource Group
5. Upload static site content to Web App via FTPS (Web Apps support Git too so could also consider this for publishing)

The code is well-documented and should be relatively straightforward to follow. It takes advantage of the `async` library, but feel free to make comments regarding whether or not the code is organized per Sails and Node best practice.


dhcole:
Nice. Thanks @anweiss! I have reviewing these + the hook-based build engine early on my list for next week. I appreciate all your work here

anweiss:
Thanks! Looking forward to seeing this all come together over the next couple of weeks.

anweiss:
Closing per #115

PR #115, opened 2015-08-07
Use hook for buildEngine

dhcole:
The buildEngine service is where we process build and site models into instructions for building the sites, and it also shells those instructions to child processes. In order to make Federalist run in a non-unix environment, this service needs to be overridden with platform-specific commands.

As a service, there is no easy way to override the buildEngine, but Sails provides a [hooks API](http://sailsjs.org/documentation/concepts/extending-sails/hooks/hook-specification), which is much more flexible.

By moving the buildEngine to a hook, it can be overridden by a separate npm module. This PR converts buildEngine to a hook and sets up a config option for specifying what build engine to use.

## Extending Federalist to another platform

@anweiss has done really great work thinking through how Federalist could run on Windows / Azure. With this PR, that work can be bundled into its own npm module following Sails's [installable hook pattern](http://sailsjs.org/documentation/concepts/extending-sails/hooks/installable-hooks):

- create a new repository called something like `sails-hook-federalist-ms`
- add a `package.json` file to that repository with something like this:

    ```json
    {
     "name": "sails-hook-federalist-ms",
     "version": "0.0.0",
     "description": "a hook to add MS support to Federalist",
     "main": "index.js",
     "sails": {
       "isHook": true
     }
    }
    ```
- add an `index.js` file that should replace `api/hooks/buildEngine/index.js`
- add any additional module dependencies to `package.json`
- add [tests for the hook](http://sailsjs.org/documentation/concepts/extending-sails/hooks/installable-hooks#?testing-your-new-hook)
- add a `readme` to let users know how to install the hook, which is simply: 1) `npm install sails-hook-federalist-ms`, and 2) set the `build.engine` config option to `'federalist-ms'` in `config/local.js` or with the `FEDERALIST_BUILD_ENGINE` environment variable
- publish the package to npm so users can easily install it

## Next steps

This PR should enable the `buildEngine` changes needed for MS support. To enable Azure publishing instead of S3, we need to move the S3 service into the buildEngine hook, so that can be overridden too. This should be pretty easy, but we also need to move the previewController into the hook, since that proxies S3. We can work on that as a separate PR.

@anweiss How does this sound? It gives you the ability to retain full control over the integration while allowing you broad access to override where necessary.

@jeremiak what do you think? Mind giving this a review?


anweiss:
This looks fantastic! Great work @dhcole. Should be able to get started on this early next week.

dhcole:
:+1: happy to help review anything and also add in a note and link to your module in the Federalist readme so users can find it easily.

jeremiak:
LGTM, great change to make Federalist a more extensible platform!

PR #118, opened 2015-08-13
Remove security headers since CF now sets them for us

dhcole:
CloudFoundry now sets security headers on all HTTP requests, so this removes that code to prevent them from conflicting.


jeremiak:
Woah, that's really awesome platform level thinking on the part of our devops team :)

PR #125, opened 2015-08-17
Editor should display files in a repo for easier editing

### LINKED ISSUE #110: Editor should list files in a repo and enable editing of those files
Right now the editor takes a full file path, but that's pretty useless unless the user can select a file from the repository to open in the editor.

The editor should make a request to the Github API to get the files for a repository and then display those files to the user. The user can then select a file to edit and it should open in the file editing view.

This should all be handled client site. We want to keep the editor module as loosely coupled as possible from Federalist the application. To this end, the server will share the Github access token with the client and the client will be able to use but the client should take that and run with it to display the file listing and push the changes back to Github.

jeremiak:
This is a PR to achieve two things:
1. To make sure the editor code doesn't go for too long in its own branch
2. To close #110

This should be easily testable as the links for "Edit" will now open that repo/branch in the Federalist editor. There should be a list of files and folders and you can work your way through the directory structure.


api/models/User.js
```diff
@@ -26,7 +26,7 @@ var User = {
     // Method to return JSON to the API
     toJSON: function() {
       var obj = this.toObject();
-      delete obj.passports;
+      // delete obj.passports;
```
dhcole: Do you know off hand if this will expose other users' passport access tokens?

jeremiak: I don't know but my guess would be that it wouldn't since all the passports will be associated with the User model. Since I can't think of a case where a passport would be shared amongst two users I think we're safe.

That being said, the goal for this was to get the token to the front end for use by the editor. Do you think setting this as a cookie would be better instead of including it in the JSON response?

jeremiak:
@dhcole The user should now be getting feedback on the status of their attempted save, so I think this is ready for review

dhcole:
Error while saving a change to `index.md` in the microsite template:

- added a bullet list block
- clicked save

screen shot 2015-08-20 at 1 12 16 pm

screen shot 2015-08-20 at 1 13 31 pm

jeremiak:
@dhcole I'm having a hard time replicating the bug, can you confirm this is what I should do:

0. Add a new site to federalist using the microsite-template
0. Edit the `index.md` file
0. Try to add a bullet list block (`<ul>`) to the page
0. See error

Instead, I'm successfully able to add a bullet list block to the page and have it render locally.

dhcole:
Yea, that's exactly what I'm doing:

aug 20 2015 15 39

jeremiak:
@dhcole we good to merge this one?

dhcole:
:+1: nice work on this

PR #127, opened 2015-08-19
Re-organize front end file structure

jeremiak:
**This is a branch off of `pull/editor-lists-files` so #125 should be considered before this. 9fdfa01730d4925e19c21292c13d06ef4fd37224 is the difference between the two branches**

@dhcole I'm curious to know what you would think of this. I'm not trying to do a bunch of file trashing, but the front end file structure was getting cumbersome to type and was more verbose than we needed it to be. In addition, there was only one level of organization for both views and templates which was a bit annoying.

I tried to rename and re-organize the views and templates to be easier to work with.


assets/app/templates/editor/edit-main.html
```diff
@@ -0,0 +1,24 @@
+<!-- <div class="row editor-preload" style="display: none;">
```
msecret: Why is this all commented out?

assets/app/templates/nav.html
```diff
@@ -1,7 +1,7 @@
 <ul id="nav-mobile">
 <% if(username) {%>
   <li>hey <%- username %>!</li>
-  <li><a href="/logout">logout</a></li>
+  <li><a id="logout">logout</a></li>
```
msecret: How are you using ids? I hope not for styling. I'd suggest `js` prefixing selectors for use in js.

msecret:
In terms of the file structure, I like having a separate folder for all compiled assets, or assets that are going to be served from the server. This would be separate from a src folder where the source files are. The reasons I like this:

- You can git ignore the compiled folder if this is OK with the deployment flow (more on that after).
- When you're running the server, you can just serve the compiled or public folder. This could mean not having to copy files around in the deployment that aren't ultimately needed.
- I find it to be cleaner.

This is just my preference though, I don't care strongly if its not like that.

assets/styles/styles.css
```diff
@@ -79,9 +79,6 @@ select {
 .preloader-wrapper {
```
msecret: I think you'll want a place or structure for "src" CSS. This could be if you want to use postcss for things, or use sass. I think most people that are using just basic CSS regret it as some point because code organization and keeping things dry. This, again, depends on your app, but I'm just taking a guess.

assets/app/views/site/add.js
```diff
@@ -5,10 +5,10 @@ var _ = require('underscore');
```
msecret: Why would these files need "View" in their name when they're in a "view" folder? It might be unnecessary.

msecret:
In the view structure, I'm confused what "blocks" is. Also, so I'm guessing "authenticate.js" is used by all site, editor so it's on hte outside? I think thats fine.

assets/app/views/editor/edit-list.js
```diff
@@ -0,0 +1,39 @@
...
+    this.files = opts.files || [];
+  },
+  render: function renderUserView() {
+    var self        = this,
+        owner       = this.path.owner,
+        repo        = this.path.repo,
+        branch      = this.path.branch;
+
+    if (this.files.length === 0) {
+      self.$el.append('<strong>No files in this folder</strong>');
```
msecret: Most people don't do this and I have mixed feelings, but keeping content in JS like this makes things like localization harder in the future. Keeping string like this in a constant would help a bit.

assets/app/views/editor/edit-file.js
```diff
@@ -0,0 +1,159 @@
...
+      blocks.push(jsonMLToBlock(mdTree[i]));
+    }
+
+    this.$('.js-st-instance').text(JSON.stringify({ data: blocks }));
+    this.editor.reinitialize();
+
+    return this;
+  },
+  saveDocument: function (e) {
+    var msg = this.$('#save-content-message').val(),
```
msecret: I'd probably be consistent and use one way to select elements form JS. I'd say prefixed classes ("js-") or data attributes.

This is also because the way I build components, I rarely envision them being used just once on a page, I don't like that constraint.

assets/app/views/site/edit.js
```diff
@@ -3,7 +3,7 @@ var fs = require('fs');
 var Backbone = require('backbone');
 var _ = require('underscore');
 
-var templateHtml = fs.readFileSync(__dirname + '/../templates/SiteEditTemplate.html').toString();
+var templateHtml = fs.readFileSync(__dirname + '/../../templates/site/edit.html').toString();
```
msecret: Why aren't you just using "require" for these templates? I was pretty sure that was possible in some sort of way.

I would say its cleaner and better structure then using readFileSync.

assets/app/views/nav.js
```diff
@@ -20,6 +23,11 @@ var NavbarView = Backbone.View.extend({
     this.$el.html(this.template(user));
 
     return this;
+  },
+  logout: function (e) {
+    e.preventDefault();
+    window.localStorage.removeItem('token');
```
msecret: I'd use a simple wrapper class for localStorage. It might help a little with testing (although stubbing and mocking is extremely easy in JS). More so, it might help if you want an abstraction on your data. So your data could technically come from local storage or your server, but the view doesn't need to know that. I like the way flux works in this respect, where you have a store for all your data.

assets/app/views/editor/edit-main.js
```diff
@@ -0,0 +1,115 @@
...
+      data: JSON.stringify({
+        path: self.path.file,
+        message: save.msg || 'Some changes to ' + self.path.file,
+        content: encodeB64(save.md),
+        sha: self.path.sha,
+        branch: self.path.branch
+      }),
+      complete: function (res) {
+        var responseText = {
+          0:   'The internet is not connected. Please check your connection.',
```
msecret: This kind of code looks like it should be global to all ajax requests?

msecret:
I'd suggest adding JS unit testing to this code base. I think you could start with at least the models as those are easy to test. Testing views can be more difficult because I'd likely require phantomjs or jsdom running.

dhcole:
Removed ready for review tag until the conflict are resolved. @jeremiak just let me know when you'd like another look.

jeremiak:
Thanks @dhcole, this should be ready to go now

assets/app/views/site/edit.js
jeremiak: It is my understanding that `require()` doesn't work for files that aren't `.js` or `.json`. When I attempt to use `require` I get errors.

Right now we have `browserify` use the `brfs` to inline the contents of those files into the JS files during the build step

dhcole: Right, this `readFileSync` approach is probably the most common way to browserify templates. You could use [stringify](https://github.com/JohnPostlethwait/stringify) to `require` html / text files (we use it on Midas). But it's just an esthetic difference.

dhcole:
:+1: Will be merged in #133

PR #130, opened 2015-08-22
Users can add, edit and remove YAML front matter from documents

### LINKED ISSUE #126: As a user, I want to be able to edit the meta data for any document in my site
Right now our editor supports Markdown but a YAML page for Jekyll (or most other static site generators) also leverages the YAML front matter for maintaining meta data (things like page title, date, where it lives in the navigation, etc).

In preparation for our demo to the college choice team at the end of the week, we want our editor to handle the basic administration of meta data: adding, editing and removing. Since YAML front matter is essentially just key value pairs, the first interface iteration should allow a user to add or edit meta data through text fields.

Ultimately (out of the scope of this issue), we'll want to make our meta data interface configurable in a similar manner to [prose](https://github.com/prose/prose/wiki/Prose-Configuration) where we can ensure that the end user will enter valid values.

jeremiak:
This PR adds meta data editing functionality so that users can add, edit and remove meta data from their documents. There is a tabbed interface on the editor for the user to toggle between content and "Settings", which is a more user friendly name for the YAML over "front matter" or "meta data".

The save button saves both the front matter and the content to Github with every successful click.

To close #126

### DIFF
--- a/assets/app/views/EditorView.js
+++ b/assets/app/views/EditorView.js
@@ -21,10 +21,13 @@ var templateHtml = fs.readFileSync(__dirname + '/../templates/EditorTemplate.htm
 var EditorView = Backbone.View.extend({
   tagName: 'div',
   events: {
-    'click #save-content-action': 'saveDocument'
+    'click [data-show-area]': 'toggleAreas',
+    'click #save-content-action': 'saveDocument',
+    'click .front-matter-delete': 'deleteMetaDataRow',
+    'click #add-front-matter-row': 'addMetaDataRow'
   },
   initialize: function (opts) {
-    this.doc = new Document({ markdown: opts.content });
+    this.doc = window.d = new Document({ markdown: opts.content });
     this.fileName = opts.file;
     return this;
   },
@@ -33,7 +36,7 @@ var EditorView = Backbone.View.extend({
         blocks  = [],
         editor, mdTree;
 
-    this.$el.html(_.template(templateHtml)({ fileName: this.fileName }));
+    this.$el.html(_.template(templateHtml)({ fileName: this.fileName, frontMatter: this.doc.frontMatter }));
     this.editor = new SirTrevor.Editor({
       el: this.$('.js-st-instance'),
       blockTypes: ["H1", "H2", "H3", "Text", "Unordered", "Ordered"]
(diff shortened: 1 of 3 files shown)


assets/app/views/EditorView.js
```diff
@@ -21,10 +21,13 @@ var templateHtml = fs.readFileSync(__dirname + '/../templates/EditorTemplate.htm
...
   events: {
-    'click #save-content-action': 'saveDocument'
+    'click [data-show-area]': 'toggleAreas',
+    'click #save-content-action': 'saveDocument',
+    'click .front-matter-delete': 'deleteMetaDataRow',
+    'click #add-front-matter-row': 'addMetaDataRow'
   },
   initialize: function (opts) {
-    this.doc = new Document({ markdown: opts.content });
+    this.doc = window.d = new Document({ markdown: opts.content });
```
dhcole: Is the global window attribute necessary or just left over from development?

assets/app/views/EditorView.js
```diff
@@ -41,17 +44,46 @@ var EditorView = Backbone.View.extend({
...
+  },
+  deleteMetaDataRow: function (e) {
+    e.preventDefault();
+    $(e.target).parents('.meta-data-rows .row').remove();
+
+    return this;
+  },
+  addMetaDataRow: function (e) {
+    e.preventDefault();
+    var rowTemplate = _.template('<div class="row"><div class="col s5"><input type="text" class="front-matter-key" placeholder="key"></div><div class="col s5"><input type="text" class="front-matter-value" placeholder="value"></div><div class="col s2"><button class="front-matter-delete">Delete</button></div></div>')();
```
dhcole: Is it worth moving this template string into its own file?

dhcole:
A few minor questions — otherwise looks great.

assets/app/views/EditorView.js
jeremiak: Yes, I think I'll do that in the merge conflicts with #127, but it was hard to imagine how to name it in a sane way

assets/app/views/EditorView.js
jeremiak: Left over from development. I"ll go ahead and remove that.

jeremiak:
Great, thanks @dhcole. Removed the global and see above re: template string

PR #135, opened 2015-08-26
Add a log of recent builds

dhcole:
Lets site owners see builds from their site. Adds a button for "builds" to the main site listing and a new view listing the recent builds and any error messages.

screen shot 2015-08-26 at 12 44 31 pm

screen shot 2015-08-26 at 12 44 14 pm

Also moves the delete button to the bottom of the settings page to keep it out of the way.

screen shot 2015-08-26 at 12 44 42 pm

Could definitely use some more formatting on the list and error messages, but my goal here is to ship all the basic functionality we're missing and improve the look as we have time.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -6,5 +6,16 @@
  */
 
 module.exports = {
-
+  /*
+   * Return an object of usernames
+   */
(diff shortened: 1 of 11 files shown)


jeremiak:
Looks great!

Issue #121, opened 2015-08-13
Windows and Azure Installable Hook


anweiss:
Creating this issue for tracking purposes. To support #97 and #98 and in reference to the new [buildEngine](https://github.com/18F/federalist/blob/master/api/hooks/buildEngine/index.js) hook, an additional installable hook is being developed and maintained here: https://github.com/Microsoft/sails-hook-federalist-ms. A separate PR will be submitted upon completion of the hook's development so as to provide instructions in the README for enabling build support on Windows and publishing to Azure.

dhcole:
:+1: this is looking great

dhcole:
Followed up in #137. Next action is to refactor as a documentation change.

PR #137, opened 2015-08-31
Build support and instructions for Windows and Azure

anweiss:
In reference to #121 and to support #98 and #97, this PR adds instructions to the README for running Federalist on Windows and publishing sites to Azure. All integration is supported by way of the [sails-hook-federalist-ms](https://github.com/Microsoft/sails-hook-federalist-ms) installable hook.

NOTE: Running on Windows still requires manual installation of all dependencies. Support for running in a Docker container on Windows using the provided Dockerfile is on hold until Windows Server is released and fully supports Windows containers

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ GEM
   remote: https://rubygems.org/
   specs:
     RedCloth (4.2.9)
+    RedCloth (4.2.9-x86-mingw32)
     activesupport (4.2.3)
       i18n (~> 0.7)
       json (~> 1.7, >= 1.7.7)
@@ -24,6 +25,7 @@ GEM
     execjs (2.6.0)
     fast-stemmer (1.0.2)
     ffi (1.9.10)
+    ffi (1.9.10-x86-mingw32)
     gemoji (2.1.0)
     github-pages (39)
       RedCloth (= 4.2.9)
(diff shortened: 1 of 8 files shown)


jeremiak:
Very cool @anweiss - Will take a look soon.

JW, is there a public instance of Federalist running on Azure we could see?

anweiss:
Awesome! Thanks @jeremiak. I'm actually in the process of setting up a running instance of Federalist in an Azure Web App and will follow-up this PR with a link. For reference, Azure Web Apps are pure PaaS with no VM dependencies.

anweiss:
@jeremiak as an update, you can now find a public instance of Federalist on Azure here: https://federalist.azurewebsites.net. I'm still getting some dependencies loaded on to the Web App so Jekyll and Hugo builds may fail. Will update once complete.

I'm also updating the MS hook to support execution of build tasks as part of an [Azure Webjob](https://azure.microsoft.com/en-us/documentation/articles/web-sites-create-web-jobs/) instead of directly from within the main app engine.

When all is said and done, I'll include an [Azure Resource Manager template](https://azure.microsoft.com/en-us/documentation/articles/resource-group-overview/#template-deployment) in a separate PR for simple deployment of Federalist to Azure. You can think of this template kind of like a Dockerfile but for the resources in Azure required to run Federalist itself.

dhcole:
Hey @anweiss, sorry to let this go stale. Thanks for all the work you did to get Federalist working well on Windows and in Azure.

I'd like to the platform-specific code changes out of the main repository since we're primarily focused on Federalist as a hosted application in our own environment. If you're up for refactoring this into something more discreet and hosted on your own repository, we can do an update to this README to point people that way.

PR #140, opened 2015-09-03
Add SBST and generic federalist-users organization

dhcole:
Allows users in the generic [federalist-users](https://github.com/orgs/federalist-users/teams/users) and [us-federalist-sbst](https://github.com/us-federal-sbst) organizations to access Federalist.


jeremiak:
Great idea, thanks @dhcole

PR #143, opened 2015-09-09
Add homepage with sign-up for private beta

dhcole:
In advance of next week's blog post, this adds a simple home page with some general audience content about Federalist as well as a link to a Google form to sign up as a beta user.

screen shot 2015-09-09 at 1 35 46 pm

screen shot 2015-09-09 at 1 33 46 pm

### DIFF
--- a/assets/app/templates/authenticate.html
+++ b/assets/app/templates/authenticate.html
@@ -1,7 +1,0 @@
-<div class="row">
-  <div class="jumbotron">
-    <h1>Welcome to Federalist</h1>
-    <p>Federalist is super cool and useful for publishing websites. But first, we need access to your Github account.</p>
-    <p><a class="btn btn-primary btn-lg" href="/auth/github" role="button">Authenticate with GitHub</a></p>
-  </div>
-</div>
--- a/assets/app/templates/home.html
+++ b/assets/app/templates/home.html
@@ -1,0 +1,32 @@
+<div class="row">
+  <div class="jumbotron text-center">
+    <h1>Welcome to Federalist.</h1>
+    <p>Federalist is a unified interface for publishing static government websites.</p>
+    <p>Have an account? <a class="btn btn-primary btn-lg" href="/auth/github" role="button">Log in now</a></p>
+  </div>
+</div>
+
+<div class="row">
+  <div class="col-md-4">
+    <h2>Secure and scalable</h2>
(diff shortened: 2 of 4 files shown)


dhcole:
By the time we publish the launch post, I think we should change all alpha => beta, so the Google form mentions beta (just avoiding something I'd forget to change).

jeremiak:
Looks great to me!

dhcole:
@GailSwanson let me know if you have thoughts or suggestions on this. We can update it as often as we like. Just throwing this together with default Bootstrap styles as a placeholder so it's not blocking launch.

dhcole:
Thanks @jeremiak

GailSwanson:
The form looks good. Just thinking about being more precise on the language so people know what to expect. Are only government agencies able to access this form? If it is open to the public, we'll need to be clear about only accepting requests from gov employees. 

Is this a request for info about being a beta user, or are they requesting a beta invite? There's a somewhat standard process for beta invites, and therefore expectations. I would recommend we make it clear that it is a request for information about being a beta user and that we will get future email with more info. I'd skip the join language. I've seen many folks get confused by the idea of joining something when it wasn't really a membership and then it bites everyone later. 

So, to bring clarity... I recommend copy on the form such as "Federalist is currently in private beta and accepting limited users. If you'd like to participate, fill out the form below." And then have a checkbox at the bottom that says "send me updates on the Federalist Platform" and on submit, display a message such as "Thank you for interest in Federalist. We will notify you if you are selected to participate in the private beta, and when it is open to the broader audience."

dhcole:
@GailSwanson that's great! Thanks. Here's the updated form: https://docs.google.com/forms/d/1iB8aW7c9r1QH3s8XElQCrnXRGjAiPUYpWG1CMeEqGIo/viewform

Do you have any thoughts on the copy on the homepage?

GailSwanson:
I wonder if the main tagline/subtitle on the page could be less jargonny and speak to the idea of getting content out to people with easy and speed. What about "Federalist is a new way to publish fast and easy-to-read government websites."

Under scalable I'd add something about not worrying about crashing under heavy traffic.

The home page mentions alpha, but the form talks about beta. We should sync this.

dhcole:
Cool, I'll look into those updates. 

> The home page mentions alpha, but the form talks about beta. We should sync this.

Yes, this is deliberate. By the time we publish the launch post, I think we should change all alpha => beta, so the Google form mentions beta (just avoiding something I'd forget to change). There's a ticket to track that here: #144

PR #145, opened 2015-09-09
Pin node to known working version

dhcole:
This pins node to the version I'm using locally. We could update it to 0.12, but let's get this working first and revisit when we have a need for it.


jeremiak:
:ship:

Issue #131, opened 2015-08-24
Monitor webhook responses


jeremiak:
We've seen a few instances of the webhook `POST` request from Github generating errors in the application and therefore preventing builds from working. It is hard to track down the exact cause, so we want to have a bit more monitoring (and ideally some notifications) around this part of the application.

Since api.data.gov is a available to us, let's start with implementing that.

jeremiak:
I renamed this issue to capture that we'll want more than just this endpoint monitored as our application is about to be part of two client launches in the month of September.  So this issue is now larger to include monitoring for the build errors as well as general application errors.

dhcole:
We have New Relic for server errors and uptime. Last step is to monitor responses from the application to GitHub webhook requests, just to make sure we don't miss anything if the server is down.

dhcole:
This has only very rarely been an issue, but if a build doesn't start after a change, check the webhook section on the repository's settings on github. There will be a list of the status returned from past webhooks. On some particularly large webhook requests, it seems like the code that validates the signature of the webhook body may be breaking, such as https://github.com/18F/college-choice/settings/hooks/5292591#delivery-request, `61340c00-c48a-11e5-8725-acd1494c6ea9`

wslack:
I think existing controls/monitoring are good enough

PR #146, opened 2015-09-09
Add New Relic for monitoring

dhcole:
Sets up New Relic. Requires the following `env` variables:

* `NEW_RELIC_APP_NAME` - application name to report to New Relic
* `NEW_RELIC_LICENSE_KEY` - license key for New Relic

Also ignores `403` errors because they are expected for unauthenticated users.

Requires setting the environment variables in CloudFoundry and running `cf restage`. I'm looking into whether we can do that without downtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Federalist 
+# Federalist
 [![Build Status](https://travis-ci.org/18F/federalist.svg?branch=master)](https://travis-ci.org/18F/federalist)
 
 ***Under active development. Everything is subject to change.***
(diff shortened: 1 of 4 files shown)


dhcole:
`env` vars are set. For future reference, running `/scripts/deploy.sh federalist` on an updated local copy of `master` will deploy the same commit as running on production and update environment variables without any downtime.

dhcole:
Refs #131

jeremiak:
LGTM

Issue #139, opened 2015-09-02
Add a "need help?" link


dhcole:
A simple link to https://github.com/18F/federalist/issues/new would give users a way to notify us of issues they find.

PR #147, opened 2015-09-09
Add support for error messages

### LINKED ISSUE #67: Only way to know you aren't authorized because of orgs is the URL
If you try and authenticate with a user that is not in one of the whitelisted organizations, you can't get in. This is awesome!

However, the only way the user would know this is if they looked at the resulting URL and saw `?error=Error.Passport.Unauthorized`. The frontend should handle this error and relay this message to the user so that they know how to get access and that the app is not just broken.

dhcole:
Sets a simple alert system and shows message for unauthorized GitHub users. 

Error keys are passed as an `error` query parameter. The main view checks for an error key, matches it to the `messages` object to get a human-readable error message, and shows the error message. If an unknown error key is passed, it will display a default error message.

Could be updated after #139 to include includes for feedback.

Closes #67

### DIFF
--- a/assets/app/views/main.js
+++ b/assets/app/views/main.js
@@ -42,10 +42,6 @@ var AppView = Backbone.View.extend({
       $('.alert-container').html(
         '<div class="alert alert-danger" role="alert">' + message + '</div>'
       );
-      console.log(1, window.location.pathname + window.location.hash);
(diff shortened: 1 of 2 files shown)


jeremiak:
:rocket:

PR #150, opened 2015-09-10
Automatically update the UI when new builds take place

dhcole:
This smoothes out the user experience by updating the site listing when a new build takes place.

add-site

It uses websockets to listen for recent build events from the API and re-renders the API when they fire.

Also, this includes status labels on the site to let users know of the last build's status, and it replaces public-facing mentions of "build" with "publish" as it is a less wonky term.

screen shot 2015-09-10 at 9 57 01 am

Staged here to validate it on CloudFoundry, though this shares the same database as production, so don't edit any sites you're concerned about: https://federalist-b.18f.gov/

### DIFF
--- a/api/models/Build.js
+++ b/api/models/Build.js
@@ -34,8 +34,13 @@ module.exports = {
   },
 
   afterCreate: function(model, done) {
+    Build.publishCreate(model);
     this.addJob(model);
     if (done) return done();
+  },
+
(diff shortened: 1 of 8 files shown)


jeremiak:
Very cool, that's great that so much comes for free with Sails. As you said in Slack, lots of possibilities.

Issue #151, opened 2015-09-10
S3 sync sometimes fails


dhcole:
Here's the log message:

```
2015-09-10T16:12:56.61-0400 [App/0]      ERR Build error:  RequestTimeout: Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed.
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.extractError (/home/vcap/app/node_modules/aws-sdk/lib/services/s3.js:322:35)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.callListeners (/home/vcap/app/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.emit (/home/vcap/app/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.emit (/home/vcap/app/node_modules/aws-sdk/lib/request.js:595:14)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.transition (/home/vcap/app/node_modules/aws-sdk/lib/request.js:21:10)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at AcceptorStateMachine.runTo (/home/vcap/app/node_modules/aws-sdk/lib/state_machine.js:14:12)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at /home/vcap/app/node_modules/aws-sdk/lib/state_machine.js:26:10
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.<anonymous> (/home/vcap/app/node_modules/aws-sdk/lib/request.js:37:9)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.<anonymous> (/home/vcap/app/node_modules/aws-sdk/lib/request.js:597:12)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.callListeners (/home/vcap/app/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.emit (/home/vcap/app/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
2015-09-10T16:12:56.61-0400 [App/0]      ERR     at Request.emit (/home/vcap/app/node_modules/aws-sdk/lib/request.js:595:14)
[...]

PR #152, opened 2015-09-11
Fix S3 sync bug

dhcole:
Bug reported in #151 seems to be related to files over 1mb failing to upload with the aws-sdk. This implements a workaround. 

It also raises the maxSockets as recommended by `node-s3`, and it removes the callback in the `error` event handler, in case an error is not actually the end and the library wants to retry the upload.

Finally, this also adds SVGs to the list of files to gzip.


dhcole:
heads up @jeremiak: I'm going to merge this now to fix the immediate issue.

PR #154, opened 2015-09-11
New front matter editor experience

jeremiak:
After we were able to observe the SBST team trying to use our editor, we came up with a list of improvements. The first and immediate things were to:

0. Change the settings/metadata/front matter editor so that it was more intuitive and easy to access all the data.
0. Move the Save options to the button of the edit page
0. Make sure that the editor would gracefully handle pages with markdown and frontmatter, just yaml, or just markdown

This PR does the above. The major difference is that we now have a new dependency on CodeMirror which is used to display any and all yaml in the editor experience.


dhcole:
Works well and looks much better. Nice job!

PR #164, opened 2015-09-14
Automatically set redirect objects based on custom domain

### LINKED ISSUE #160: CFront => S3 trailing slash confusion
When using CFront with an origin path to a S3 bucket prefix, if the request is for a directory without a trailing slash, S3 returns a redirect to the directory's full key, which includes the origin path. So you end up getting something like:

CFront: `/data` => S3: `/site/dhcole/repo/data` => 301 redirect: `/site/dhcole/repo/data/`

When the browser follows that redirect, it ends up requesting  `/site/dhcole/repo/data/` from CFront, which goes to the origin as  `/site/dhcole/repo/site/dhcole/repo/data/` :fire: 

Temporary workaround is to add a redirect object to as `/site/dhcole/repo/data` that redirects to `/data/`

We need to figure out a sustainable solution. Something like S3 redirect rules might work...

dhcole:
Discussed in and fixes #160 

Automatically sets empty objects with redirect headers for directories. The redirects default to `/site/:user/:repo/[:branch]/path => /site/:user/:repo/[:branch]/path/` and switch to `/site/:user/:repo/[:branch]/path => /path/` for sites with custom domains. This second part is what S3 was not doing on its own, so setting these redirects ourselves gets around S3's default behavior and avoids the need for global redirect rules.

### DIFF
--- a/api/services/S3.js
+++ b/api/services/S3.js
@@ -17,10 +17,7 @@ http.globalAgent.maxSockets = https.globalAgent.maxSockets = 20;
 
 // Work around upload bug for files > 1mb
 AWS.util.update(AWS.S3.prototype, {
-  addExpect100Continue: function addExpect100Continue(req) {
-    sails.log.verbose('Depreciating this workaround, because introduced a bug');
-    sails.log.verbose('Check: https://github.com/andrewrk/node-s3-client/issues/74');
-  }
+  addExpect100Continue: function addExpect100Continue(req) {}
 });
 
 module.exports = function(config, done) {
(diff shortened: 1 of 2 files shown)


api/services/S3.js
```diff
@@ -17,10 +17,7 @@ http.globalAgent.maxSockets = https.globalAgent.maxSockets = 20;
 
 // Work around upload bug for files > 1mb
 AWS.util.update(AWS.S3.prototype, {
-  addExpect100Continue: function addExpect100Continue(req) {
-    sails.log.verbose('Depreciating this workaround, because introduced a bug');
-    sails.log.verbose('Check: https://github.com/andrewrk/node-s3-client/issues/74');
-  }
```
dhcole: This logging was unnecessary

dhcole:
Going to :ship: in advance of a site launch tonight

PR #165, opened 2015-09-15
Set redirects for nested directories too

dhcole:
Follows on #164, which only set redirects for top level directories. This applies directory redirects to all nested directories as well.


jeremiak:
LGTM. Using the headers for redirection is pretty elegant, nice job!

Issue #167, opened 2015-09-15
Disable sir trevor


dhcole:
@jeremiak I know we've got a really promising alternative. I think we should disable sir trevor in the short term though, as it's munging content that doesn't have a block associated with it.

https://github.com/us-federal-sbst/federalist-modern-team-template/commit/9f049fdee06cdc9188f9139444ab2dfc54815614

dhcole:
Done in #168

PR #168, opened 2015-09-15
Implement ProseMirror instead of Sir Trevor

jeremiak:
We learned from our experience with the SBST folks that Sir Trevor was a little less usable than we had wanted and it is currently munging some of the content in markdown files (#167). This PR swaps out Sir Trevor for [ProseMirror](http://prosemirror.net/).

The changes include:

0. Removing Sir Trevor as a dependency
0. Listing ProseMirror as a dependency and setting up the proper install scripts to ensure it builds


jeremiak:
This is ready for your review, @dhcole

dhcole:
:ship: :dash:

PR #172, opened 2015-09-16
Use CodeMirror if ProseMirror errors out

### LINKED ISSUE #170: Error with ProseMirror
@jeremiak check out this page: https://federalist.18f.gov/#edit/us-federal-sbst/federalist-modern-team-template/master/pages/2015-annual-report.md

Wondering if it's because of unstandard markdown, such as `View 2015 Annual Report - PDF{: .usa-button }`?

What do you think? Should we make buttons includes or just normal html, which should be supported in a markdown document?

jeremiak:
ProseMirror errors out more readily than CodeMirror does (#170) and we want to ensure that every document can be displayed in Federalist.

This PR changes the behavior so that if loading content into ProseMirror errors out, we will use CodeMirror as the content editor instead.

@dhcole ready for your review to close #170

### DIFF
--- a/assets/app/views/editor/edit-file.js
+++ b/assets/app/views/editor/edit-file.js
@@ -21,7 +21,8 @@ var EditorView = Backbone.View.extend({
   template: _.template(templateHtml),
   initialize: function (opts) {
     var self      = this,
-        activeTab = 'content';
+        activeTab = 'content',
+        settingsEditorEl, contentEditorEl;
 
     this.editors = {};
     this.path = opts.path;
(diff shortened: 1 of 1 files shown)


dhcole:
:+1: works great locally. I think this is the way to go. Now you just get a degraded editing experience instead of a hard fail or (worse) missing content.

jeremiak:
I agree, the editor should always display some content. I imagine that its going to be jarring to see this degraded experience, so we might want to give the user feedback about the reason we did so.

PR #178, opened 2015-09-18
Check for content editor before saving

### LINKED ISSUE #177: Documents with no markdown don't save
We treat some documents differently than others. For instance, if you open a `.md` file that has front matter and markdown, it will open in the two tabbed interface where you can edit both content and front matter.

If you try and open a `.yml` document, it will not display the content editing section because there is no content. `yml` files are just for structured data.

However, the save mechanism has a bug where it won't save `.yml` files because it expects them to have content. The user gets no feedback, but the console has the following error:
> `Uncaught TypeError: Cannot read property 'content' of undefined`

The editor should be able to handle both types of document and save them successfully.

jeremiak:
To fix #177 we now check for the existence of the content editor (implying that
there was markdown in the document and not just a `.yml` file) before
we attempt to get the text from that editor.

### DIFF
--- a/assets/app/views/editor/edit-file.js
+++ b/assets/app/views/editor/edit-file.js
@@ -50,7 +50,7 @@ var EditorView = Backbone.View.extend({
       tabSize: 2
     });
 
-    contentEditorEl = window.e = this.$('[data-target=content]')[0];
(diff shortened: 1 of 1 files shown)


dhcole:
thanks for the quick fix. looks good!

PR #182, opened 2015-09-18
Use redis service in production for sessions and sockets

### LINKED ISSUE #123: Shared session store
The session store is still server memory. Need to move to a shared store so users aren't logged out during deployment / server failure.

dhcole:
This uses a redis service as a shared store for sessions and socket.io identifiers. It allows for sessions to stay active across multiple instances, giving us the ability to scale the API by adding more instances, and avoiding logging out users when a server instance is restarted.

Closes #123 

## Steps to test locally:

1. install redis `$ brew install redis`
2. start redis: `$ redis-server`
3. add the following to `/config/local.js`

    ```js
    module.exports.session = {
      adapter: 'redis',
      db: 0
    };
    module.exports.sockets = {
      adapter: 'socket.io-redis',
      db: 1
    };
    ```
4. start up the server as usual (`npm start`)

You should be able to log in, restart the server, and still be logged in.

@jeremiak this is ready for your review.

After we merge, and the blue-green deploy finishes successfully, we should check to make sure we don't need to manually deploy to update the manifest.

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -2,7 +2,8 @@ var AWS = require('aws-sdk'),
     cfenv = require('cfenv'),
     appEnv = cfenv.getAppEnv(),
     dbURL = appEnv.getServiceURL('federalist-database'),
-    s3Creds = appEnv.getServiceCreds('s3-sb-federalist.18f.gov');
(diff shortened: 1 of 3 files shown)


jeremiak:
Super cool, looks good @dhcole!

PR #195, opened 2015-10-08
should the sails port be 1337 in Dockerfile

openbrian:


dhcole:
This doesn't really matter, since Sails will listen on any port specified in the `PORT` environment variable. In the docker file, the first line you changed sets the `PORT` env var to `3000` and the second line exposes it from the Docker VM to the host machine.

openbrian:
But at least it should agree with the GITHUB_CLIENT_CALLBACK_URL, right?

dhcole:
Yes, that's correct. That's left undefined so you can specify it based on your environment.

openbrian:
cool

dhcole:
Also, (I remember now) this is specifically set to 3000 instead of the default so that it works well with [Kitematic](https://kitematic.com/), which auto-detects applications running on a common ["web" port](https://docs.docker.com/kitematic/userguide/#working-with-a-container).

Make sense? I'm going to close this PR to we keep compatibility with Kitematic.

openbrian:
Thanks for explaining.

Issue #183, opened 2015-09-18
Add SBST template to the template gallery


wslack:
I don't see the SBST site on the list of templates. Could the list have screenshots of "live" sites on those templates? I might also be misunderstanding.

dhcole:
It's not added to Federalist yet — we have to do a bit cleanup of work to pull out SBST-specific content (just ran out of time after the launch). This is high on our list for when we pick up the next round of work. You can see it here though: https://github.com/us-federal-sbst/federalist-modern-team-template

mogul:
Refer also to [this pull-request](https://github.com/18F/federalist-modern-team-template/pull/8)

dhcole:
Closed by #197

Issue #191, opened 2015-10-01
Build a library of includes


dhcole:
To make templates more modular, we can design a library of includes that can be dropped into any template. Could be something site-specific like a Digital Gov Search box, or something page-specific like a youtube embed that takes a youtube url and formats as a responsive embed using the no-cookie domain.

el-mapache:
@wslack This is kind of similar to the template work scott is doing, correct? I think we can close this.

wslack:
Similar yes, but don't know if we should close. Scott's work is for a config setting to create whole pages - vs HTML components that can dropped into templates.

adborden:
IMHO, this is the kind of thing that having a community site would help with. Federalist users can post snippets of html/css/liquid that can be copy/pasted and reused. We could seed the site with some initial examples, but this would be maintained by the community.

The community site could also be implemented as a Federalist site (and ideally with Disqus, but I understand there's some hoops to jump through, first) :)

PR #197, opened 2015-10-08
Add team template

jeremiak:
This adds the team template to the new site gallery. The template is a derivative of the work we did for [SBST.gov](https://sbst.gov) and this addresses #183.

@dhcole This is ready for your review. Should we open an issue to describe the template builder we discussed on Monday? That will replace the whole template gallery.


dhcole:
@jeremiak notes from that are here: https://github.com/18F/federalist/wiki/Customizable-Templates

I think we can break it apart into separate issues when we're ready to tackle it. Likely that will also address #191

dhcole:
And I think the page-level components may work as `{% include ... %}` on the back end and Prosemirror buttons on the front end with fields that set the [include parameters](http://jekyllrb.com/docs/templates/#includes) (such as "Youtube video URL"). Lots of scoping to figure out there, so definitely a separate issue(s).

dhcole:
:rocket:

PR #204, opened 2015-10-15
Add support for multi-user sites

### LINKED ISSUE #166: Sites should support multiple users
The basics are in place for this, but the interface needs to be worked out. Currently if two users add the same repository to Federalist, the will have separate copies of the site in their view of the interface, but the site will publish to the same location. This becomes a problem when they have conflicting configuration, such as one user setting a custom domain, which would not be present when the other user builds the site. Also users only see build logs associated with their copy of the site.

A better flow could work like this:

## First time a site is added to Federalist

- user must have admin access to the repo on github (to add the webhook)
- creates a new site model
- builds the default branch

## Subsequent users add an existing site

- user must have commit access to the repo on github
- adds the user to the existing site model

## Github admin users

- Only users who are admins on the github repo would see the site configuration settings
- On each github authentication, each user's permissions should be validated
- Could optionally have an "invite users" feature, but we'd need to figure out on-boarding both Federalist and GitHub

### LINKED ISSUE #176: Multiple Sites are actually the same site
I created a new site using the 'microsite template'.  I then created another two sites, again using the 'microsite template`.   

screenshot 2015-09-17 at 12 43 32 pm

They all three actually appear to be the same site.  When I click 'View', I see the same URL.  When I edit a page in one, it's edited in all three.  

Sorry if I'm missing something, but it doesn't seem that I can create and manage 2 sites using that template.  This isn't surprising given #138, but it's problematic that it looks like it worked (see screenshot), when it doesn't.

### LINKED ISSUE #194: Users should see useful feedback when trying adding a repo that doesn't exist
@meiqimichelle ran into an issue while attempting to add a repository on the new site page that doesn't actually exist. There is no message displayed to the user. It appears as if the form never submits.

dhcole:
This allows multiple users to share a site on Federalist. The first user to add a site must be an admin. Subsequent users must have write access (based on GitHub permissions).

Also includes better error messages for failures to add a site and prevents creating new sites if a template is already forked.

As a follow-up, we need a migration script to clean up existing cases of duplicate sites with multiple users and duplicate fork sites.

Fixes #194 
Fixes #176 
Resolves #166


dhcole:
Now includes a migration script to merge users on duplicate sites and delete duplicates. Looking at the production DB, this really only affects two sites (sbst.gov and 18f.gsa.gov). Lots of users have the same site added multiple times, which this will also clear up.

assets/app/views/site/add.js
```diff
@@ -32,8 +32,9 @@ var AddSiteView = Backbone.View.extend({
       if (d.name === 'users') d.value = [+d.value];
       data[d.name] = d.value;
     });
-
-    new SiteModel(data, { collection: this.collection }).save(null, {
+    $.ajax('/v0/user/add-site', {
```
jeremiak: Is this an example of what you meant by #206 @dhcole?

api/controllers/UserController.js
```diff
@@ -17,5 +17,52 @@ module.exports = {
...
+        if (_(site.users).pluck('id').contains(user.id)) {
+          return done('You have already added this site to Federalist');
+        }
+        if (site.users.length) {
+          site.users.add(user);
+          site.save(function(err) {
+            if (err) return done(err);
+            return res.json(site);
+          });
+        } else {
```
jeremiak: Would this block ever get triggered? It seems to me that theoretically all sites should have at least one user in its `users` collection

jeremiak:
:+1: this looks great to me.

My local database isn't in a state where I can test the migration script, but it looks fine to me. Given that we're on CD now, and this merge will go out to production automatically how do we trigger the migration script? I'll hold off on merging this PR for now

dhcole:
Migration runs on every `npm install`

jeremiak:
Thanks @dhcole looks great! Merging now.

PR #212, opened 2015-10-25
Fix page listing for S3 configured Federalist

dhcole:
When Federalist uses S3 to host a site, the site is not available at `/site/...`, so we need to add the `siteRoot` to the URL for `_navigation.json`. This should address that.


jeremiak:
This looks good to me

PR #213, opened 2015-10-25
Always use HTTPS for S3 urls

jeremiak:
We were instructing the application to try and access the `_navigation.json` over `http` instead of `https`. This resulted in a mixed content warning and the JavaScript would throw an error. The user facing result was a blank view instead of a page listing.

Now all S3 URLs in the `/site` API response will be using `https`.

@dhcole - to you for review

### DIFF
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -48,7 +48,7 @@ module.exports = {
       var obj = this.toObject(),
           config = sails.config.build || {};
       // Add siteRoot to the API response for previews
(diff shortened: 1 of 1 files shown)


jeremiak:
Closing this because this method actually has unintended consequences. From @dhcole in chat:

> so we want to use the website bucket for the model
> because we want redirects and /index.html to work
> but for the navigation file specifically, we dont need that
> so we can use the s3 api endpoint instead of website endpoint

PR #214, opened 2015-10-25
Conditionally set root for _navigation.json

dhcole:
If the `siteRoot` is a S3 bucket, use the S3 API endpoint over https.


jeremiak:
Great, thanks @dhcole

PR #216, opened 2015-10-25
Check for site model before getting siteRoot

jeremiak:
We are creating a new Github model every time the URL changes, and the view was breaking because we tried to call `get()` on `undefined`. Now we check for the presence of the model, and then call `get()`.

In the case that an error happens (does not include ajax related errors such as 404s), we should log it to the console.


jeremiak:
I'm closing this issue, because I don't think it properly fixes the issue. In doing some more debugging it looks like the siteRoot RegEx statement doesn't match anything because it has a trailing "/"

PR #217, opened 2015-10-26
Remove trailing slash

dhcole:
Fix the regex pattern for matching S3 buckets. I mistakenly added a trailing slash to the pattern.


dhcole:
cc @jeremiak 

Merging now so we can get this fixed.

PR #220, opened 2015-10-28
Add support for external builds

dhcole:
This is the final step in containerizing builds. It queues build jobs as SQS messages (which `federalist-builder` checks and runs build tasks on ECS using the `federalist-docker-build` image).

It's a relatively lightweight override that based on the presence of an environment variable, bypasses the whole build engine.

This PR also adds a new callback endpoint that the build container and `POST` the status of a build back to on completion.

I think this is pretty close to ready to go. Feel free to start reviewing the code. I'm going to set up a plan for manual testing to we can validate this approach against all existing sites in production, our templates, and a few other 18F sites like 18f.gsa.gov and the Hub.


dhcole:
This should be ready for testing. It's staged here:

https://federalist-test.18f.gov/

Notes on testing:

To test a site you have previously added to the main Federalist application, go to the repo's settings on github and change the webhook domain to `federalist-test.18f.gov` instead of `federalist.18f.gov`.

To try rebuilding a site, just open the site settings on federalist-test.18f.gov and clicking "save".

**This test server shares a DB with production, so don't make any changes to production sites. It's best to fork sites first and test your fork instead of the organization-owned site.**

dhcole:
All tests successful except `college-choice`, because that repo specifies an old version of Ruby. I'm working with the college-choice team to get that fixed (https://github.com/18F/college-choice/pull/1394)

Otherwise, I think the last step here is to fix the breaking tests.

dhcole:
:rocket:

PR #221, opened 2015-10-28
Remove github oauth tokens from log messages

dhcole:
Replaces github oauth tokens with a redaction message.


jeremiak:
Code looks good and runs locally. Will merge when tests pass

PR #223, opened 2015-10-30
Remove temporary build files after publishing

dhcole:
This removes temporary files (the build Jekyll sites) to avoid the instance disk space from filling up.


dhcole:
@jeremiak read for review

jeremiak:
Looks great to me, here's to hitting our CF disk quota less often! :tada:

Issue #224, opened 2015-11-09
Improve new file addition


jeremiak:
This is a follow up to #174 because it is a bit of a weird experience for end users as it is currently implemented. Right now the experience is three steps:

1. Click "Add a new page" on the listing of pages in a site
![screen shot 2015-11-09 at 4 16 45 pm](https://cloud.githubusercontent.com/assets/780941/11046907/cca4b130-86fd-11e5-968d-2f18ae7b5011.png)

2. Enter the "path" of the new file. This is pretty confusing for anybody who uses Federalist because our pages/`_navigation.json` feature enables us to hide as much of the project structure as we want
![screen shot 2015-11-09 at 4 16 51 pm](https://cloud.githubusercontent.com/assets/780941/11046926/ecdfe71c-86fd-11e5-8781-8cf09370a461.png)

3. The editor loads with your new post. It looks for a file called `_defaults.yml` in the root of the project directory
![screen shot 2015-11-09 at 4 16 59 pm](https://cloud.githubusercontent.com/assets/780941/11046963/1c63693c-86fe-11e5-877b-38bfe20aef1f.png)

I'm sure this process can be easier to understand for end users. In particular, it should be easier for them to understand where the page will sit in relation to other pages in their site.

@GailSwanson and @dhcole, thoughts?

jeremiak:
In addition, the defaults should be pulled from `_config.yml` and be the regular Jekyll post defaults. Right now they are expected to be in their own file `_defaults.yml`.

dhcole:
@jeremiak all done, right?

jeremiak:
:+1 all done

Issue #174, opened 2015-09-16
New post with default "settings"


gboone:
On 18f.gsa.gov we require blog posts to have a title, authors, tags, and a few other things. It'd be handy if there were a button somewhere in the edit interface that adds a new document in a folder with any required settings added at post creation.

The UX for a site owner might be having an interface for adding required settings on a per-collection basis (or something).

The UX for the editor might be:
1. Click to add a new post
2. Fill in a webform with any required settings
3. Go right into editing the content

I'd be interested in helping out with this, I've done some work with validating frontmatter in Jekyll. This seems somewhat dependent on #166 (or at least lower priority).

dhcole:
@jeremiak good to close or should more be done before this is settled?

dhcole:
Not sure if #228 would be considered a blocker here.

PR #225, opened 2015-11-10
First pass at adding a new page

jeremiak:
This is a first pass at implementing functionality to allow users to create new pages on their site. It leaves a bit to be desired in terms of interactions, but we can clean that up after we figure out another approach in #224.

The front end code now also has a test runner set up and the `Github` model (and the new methods related to adding pages) should be covered by tests.

Refs #174


assets/app/views/editor/edit-main.js
```diff
@@ -60,6 +63,27 @@ var EditView = Backbone.View.extend({
     this.pageSwitcher.set(childView);
 
     return this;
+  },
+  newPage: function(e) {
+    var self = this,
+        path = prompt('Please the new file path', 'pages/thing.md'),
+        opts = {
+          path: path,
+          content: '---\nyo: dude\nlayout: page\n---\n'
```
dhcole: Have you implemented the defaults.yml file you mentioned in standup, or is that coming later?

test/mocha.opts
```diff
@@ -1 +0,0 @@
---timeout 60000
```
dhcole: What's the reason for deleting this? I put it in because in Travis, the tests sometimes wouldn't finish in the default time allotted.

jeremiak: My understanding of mocha is that it looks for this file in the root of `/test` so it seemed to not have any effect after I moved the test files to a new directory.

I added the [config option](https://github.com/18F/federalist/blob/pull/new-post/package.json#L73) to the `package.json` so that the test doesn't timeout on Travis. Would it be better to keep this file and pass in the path to `mocha`?

assets/app/views/editor/edit-main.js
jeremiak: Whoops, this is some debugging code, I'll go ahead and remove that.

The way I implemented the `defaults.yml` (which I think we should maybe to change to `_federalist_defaults.yml` and change `_navigation.json` to `_federalist_navigation.yml` so that they're consistent) is that it is a markdown document and the content of that document is just put into the new file being created. If there is no document, then nothing should be put into the page by default.

Any changes you think I should make to that process?

test/mocha.opts
dhcole: :+1: good call. Sorry, missed that in the package.json.

assets/app/views/editor/edit-main.js
dhcole: I think eventually we'd want to support more granular default content / settings, like by path or collection type. Also, there's a format for defaults in config.yml, which we could parse and read (that already supports type and path filtering).

I'd be fine with merging this as is though and revisiting.

jeremiak:
@dhcole The offending debug code has been removed, so this should be OK for a review

dhcole:
When creating a new page on a repo that does not have a _default.yml, I see the following:

dhcole:
Looks like the way you've increased the content area is not increasing the editable area in Prose Mirror:

screen shot 2015-11-12 at 11 04 50 am

As a result, if you click anywhere within the content area but outside of the outlined box above, the editor doesn't get focus and is not usable.

dhcole:
Otherwise, it's looking good to me!

dhcole:
:+1: works well now

jeremiak:
Merging per our conversation in Slack.

PR #226, opened 2015-11-10
Pin buildpack versions

jmcarp:
I noticed that the CF team pushed a new release of the node buildpack a few days ago, and it doesn't seem to compile on our setup. This patch pins the node buildpack to the last working version, and the ruby buildpack to the current version, since I'm not aware of any problems with it. In general, it's probably a good idea to pin versions of all custom buildpacks (and any dependencies installed straight from github).


dhcole:
Good call.... thanks for this @jmcarp, probably saved us a lot of heart ache :)

jeremiak:
Thank you @jmcarp!!

PR #229, opened 2015-11-18
Adjust `rm` commands 

dhcole:
...to make sure temp files are deleted before and after running a build process.


dhcole:
Also hard-codes the CF scale settings to `2G` disk, `1G` memory, and `2` instances.

jeremiak:
:+1: power delete!

PR #230, opened 2015-11-19
Allow site-level config for preview visibility

### LINKED ISSUE #210: Add an option to make previews public
For public repos I think there's a good case to be made for allowing public access to branch preview URLs. This is something that would've made College Scorecard much easier to test, and will definitely ease our work on EITI, which is now running on Federalist.

Maybe this could be an opt-in option?

dhcole:
This adds a user option to the site settings for specifying the visibility of draft site previews.

By default, previews continue to require authentication. If a user chooses, the preview may be visible to the public instead.

The setting is on the site settings page:

screen shot 2015-11-19 at 9 43 51 am

Resolves #210

### DIFF
--- a/assets/app/views/main.js
+++ b/assets/app/views/main.js
@@ -86,7 +86,7 @@ var AppView = Backbone.View.extend({
     var siteEditView = new SiteEditView({ model: this.sites.get(id) });
     this.pageSwitcher.set(siteEditView);
     this.listenToOnce(siteEditView, 'site:save:success', function () {
-      this.home();
+      federalist.navigate('', { trigger: true });
     }.bind(this));
     return this;
   },
--- a/assets/styles/styles.css
+++ b/assets/styles/styles.css
@@ -71,3 +71,10 @@ body {
 .list-group-item ul {
   padding-left:0;
 }
+
+.checkbox input[type=checkbox],
+.checkbox-inline input[type=checkbox],
+.radio input[type=radio],
+.radio-inline input[type=radio] {
+  margin-left: 0;
+}
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -44,6 +44,10 @@ module.exports = {
     config: {
       type: 'string'
     },
+    publicPreview: {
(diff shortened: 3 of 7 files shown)


assets/app/views/main.js
```diff
@@ -86,7 +86,7 @@ var AppView = Backbone.View.extend({
     var siteEditView = new SiteEditView({ model: this.sites.get(id) });
     this.pageSwitcher.set(siteEditView);
     this.listenToOnce(siteEditView, 'site:save:success', function () {
-      this.home();
+      federalist.navigate('', { trigger: true });
```
dhcole: This was causing a bug where it was not possible to go back to the site settings view after saving the form.

assets/styles/styles.css
```diff
@@ -71,3 +71,10 @@ body {
 .list-group-item ul {
   padding-left:0;
 }
+
+.checkbox input[type=checkbox],
+.checkbox-inline input[type=checkbox],
+.radio input[type=radio],
+.radio-inline input[type=radio] {
+  margin-left: 0;
+}
```
dhcole: Bootstrap shifts this alignment by default expecting nested labels. For 508 compliance, we prefer labels with `for` elements instead, so this adjusts the default spacing.

jeremiak:
It doesn't look like we're actually using the `previewProxy` policy on any of the routes, so sites are still blocked if you're unauthenticated.

I'm not sure how to configure the policies to use `sessionAuth` or `previewProxy`. If we only use `sessionAuth`, then the `previewProxy` policy will have no effect. However, if we only use `previewProxy`, the app crashes if `publicPreview` is set to be `false` but there is no authenticated user.

dhcole:
Good call. That's a policy for restricting previews to only users of a particular site. We loosened it to all logged in users based on feedback. I just updated this PR to work with `sessionAuth` (and tested it better this time). Thanks for the catch @jeremiak; ready for another review.

jeremiak:
All looks good to me!

PR #232, opened 2015-11-23
Fixes contact and license section of about.yml to address Dashboard errors.

gboone:
Addresses errors in the [Team API](https://team-api.18f.gov/public/api/projects/federalist/) that could be exposed in the Dashboard.


jeremiak:
:+1: thanks @gboone!

PR #233, opened 2015-11-25
Switch to the USWDS instead of Bootstrap

### LINKED ISSUE #231: Transition project to US pattern library instead of bootstrap
We are currently using Bootstrap, but an upcoming design change is based on [18F's web design standards](https://github.com/18F/web-design-standards) so we want to switch out our CSS framework in preparation.

We cataloged the UI elements that we [currently have](https://github.com/18F/federalist/wiki/UI-views-and-components) and some adjustments that we'll want to make while doing this work:

- [x] Ensure that the application interface `<a>` all look like buttons
- [x] Standardize the `<h1>` across the app. Currently it looks different on site settings, site builds and the content editor
- [x] Change the breadcrumb to be an `<h2>` on the content editor that provides useful links for the user to get back to the main listings of that site

jeremiak:
To close #231. The changes should be mostly visual and with some slight functionality change. At a high level the changes are:

0. Remove Bootstrap and implement USWDS
0. Make the `<h1>` usage consistent across the site
0. Remove tabbed views on page edit
0. Alter the behavior of the breadcrumb element so that it makes more sense depending on its context. For instance, it should be only one link on pages that have a `_navigation.json` because all links go to the page listing. For sites without the `.json` file, the breadcrumb should work how it has previously.

@dhcole Ready for your review.


dhcole:
Is there a package-managed way to use the Web Standards? Copying all the files locally seems like it makes maintaining things more complicated.

dhcole:
I remember we discussed in stand-up that it's not currently available through npm... so I guess it's either manually copy (like this) or try a git submodule.

jeremiak:
I agree that copying over is super messy, but I'm not sure a submodule would be much easier to manage. Probably the best is to publish it on NPM, I know they have an open issue for this (18f/web-design-standards#877 & 18F/web-design-standards#767).

My preference would be to merge this (when it looks good, the origin of the USWDS not with standing) and then use the NPM published version once it is up. That's not a strongly held preference, however.

jeremiak:
I opened [PR 902](https://github.com/18F/web-design-standards/pull/902) against the web design standards to add a `package.json` so that we list the library as a proper dependency.

dhcole:
:+1: if we don't get a merge on that tomorrow, we can pin to the PR's branch, right? That will unblock this so we can get it live tomorrow.

jeremiak:
Yup, we can definitely pin to a specific branch, since that's basically a tag under the hood.

The branch is `npm`, in case we need it.

jeremiak:
@dhcole No more conflicts so ready for your review. The USWDS team is moving to a branch release model, so the next release will be `0.9.x` and will include the `package.json`. They were kind enough to already merge that into the `v0.9x` branch which I pinned against in the dependencies. I also updated the `npm scripts` so that the asset moving happens for the web standards like it did for bootstrap.

dhcole:
@GailSwanson just a reminder to check this out if you have a chance. Would be great to merge this in today so @jeremiak can merge it with his image upload branch.

PR #240, opened 2015-12-07
Add image picker to editor

### LINKED ISSUE #188: Support for uploading and managing assets
Things like images or other assets to be linked in content need an upload and management interface

### LINKED ISSUE #193: Figure out how to handle baseurls
Right now, we rely on baseurls for the preview links on the application. However, that can get in the way of links in the markdown editor since we have to use syntax like:
` {{ '/assets/files/my-doc.pdf' | prepend: site.baseurl }}`

One option is for us to use subdomains instead of baseurls.

jeremiak:
Adds the ability for users to upload and use images within their documents, to close #188.

This extends the ProseMirror editor to alter how the image selection works. For that, we need to make sure to have merged 18f/prosemirror#1. A user can select to add an image to their page, and will then be prompted to add a new image or use an already uploaded one. By default, all uploads are configured to be put in a directory `/uploads`.

This should also close #193 because we are able to handle baseurls relatively gracefully now. It is all transparent to the user, but the extension of the ProseMirror editor provides a pattern for showing something different to the user than we commit to the markdown document. This allows us to save the content with `{{ site.baseurl }}` but still have the assets show up in the editor.


jeremiak:
A note for testing: right now the `prosemirror` dependency is pointed to `18f/prosemirror` but it actually needs to be from `prosemirror/prosemirror#master`. I didn't want to change the `package.json` but I do have a PR open on the prosemirror repo to fix this.

We can't merge that PR until just before we want to merge this to production because any new CF instances that get spun up (either by us or by CF) will need the current `master` of our prosemirror

dhcole:
Nice work. this looks great!

PR #241, opened 2015-12-07
Use fonts from USWDS

jeremiak:
Adds new NPM tasks to make sure that google-fonts.css will be copied over from USWDS, so that we can use them instead of declaring the fonts in the application styles.

Since it comes from a library, we want to also make sure it is included in the `.gitignore`

Fixes lack of fonts on production


dhcole:
:+1:

PR #244, opened 2015-12-08
Update manifest and fix template forking

dhcole:
Also pin node version to prevent auto-updating new versions, which recently broke the deployment process

This fixes two small issues that came up while deploying #220


dhcole:
Also added a fix for the 404 error on the "add new site" page.

jeremiak:
:+1:

PR #245, opened 2015-12-08
Hide prosemirror if doc is YML only

### LINKED ISSUE #242: Bug with loading yml files
@jeremiak just noticed this error, which I think is related to the image PR. If you try to open a file like `_config.yml` in the editor, you get an error like the following:

Not sure if this is related to our code or the Prosemirror update. What do you think?

screen shot 2015-12-07 at 6 35 08 pm
screen shot 2015-12-07 at 6 35 18 pm

jeremiak:
Fixes #242 by inserting a blank string in prosemirror if the file extension is "yml" or "yaml" and then hides the ProseMirror editor


dhcole:
:+1:

PR #246, opened 2015-12-08
Create new post from defaults

### LINKED ISSUE #174: New post with default "settings"

### LINKED ISSUE #228: Add new page cannot be canceled
Right now, you can't exit out of the create a new page once you have clicked the button. Clicking cancel will still try and issue the `PUT` request. You can see this if you open your dev console and:

0. Navigate to one of your sites page listing
0. Click new page
0. Click cancel
0. See that the `PUT` request is still made, but the URL returns a 404 since it doesn't have a filename in the path.

If you are already editing a page, the `PUT` request will go throw and remove all of the content in file you are currently editing.

jeremiak:
Changes the add new page behavior to trigger an empty editor (with any post defaults as defined in `_config.yml`), and only saves that page back to Github when the user clicks save. The filename is derived from any `title` attribute in the front matter or a unique string in the case that there is no title.

If the user navigates away from the page without saving then the page won't be committed to Github (to close #228)

This should close #174, #224 and #228


assets/app/views/editor/edit-file.js
```diff
@@ -143,13 +153,43 @@ var EditorView = Backbone.View.extend({
     if (settings) this.doc.frontMatter = settings;
     if (content) this.doc.content = content;
 
-    this.model.commit({
-      content: this.doc.toMarkdown(),
-      message: this.$('#save-content-message').val()
-    });
+    if (this.isNewPage) {
+      pageTitle = fileNameFromTitle(yaml.parse(settings)['title']);
+      pageTitle = pageTitle.toLowerCase().replace(/\s/g, '-');
```
dhcole: Should we tighten this to something like `/^\w/`, which should filter for only alphanumeric and _ characters

jeremiak: Ya, I think that's a good idea. Going to use `/\W/`

package.json
```diff
@@ -59,7 +58,8 @@
     "to-markdown": "^1.2.1",
     "underscore": "^1.8.3",
     "uswds": "git+https://github.com/18F/web-design-standards.git#v0.9.x",
-    "validator": "^3.39.0"
+    "validator": "^3.39.0",
+    "yamljs": "^0.2.4"
```
dhcole: @jeremiak: Curious what made you switch libraries?

assets/app/models/Github.js
```diff
@@ -173,16 +186,15 @@ var GithubModel = Backbone.Model.extend({
...
-      commitOpts = {
-        path: opts.path,
-        message: opts.message || 'The file ' + opts.path + ' was created',
-        content: opts.content || content
-      };
-    this.commit(commitOpts);
+  getDefaults: function () {
+    var config = this.configFiles['_config.yml'],
+        defaultConfigs = (config.present && config.json) ? config.json.defaults : [],
+        defaults = defaultConfigs.filter(function (c) {
```
dhcole: @jeremiak getting an error here if `defaultConfigs` is undefined, `.filter` is not available on `undefined`. What should happen if there is not a default section of the config file?

jeremiak: I think that if condition isn't complete. It should be something like `config.present && config.json && config.json.defaults` but that's pretty verbose. Essentially defaultConfigs should always at least be an empty array.

package.json
jeremiak: There was a problem with browserifying `js-yaml` for some reason. We experienced the same thing in 18F/about-yml-npm#3 so I just switched to the other library here as well.

assets/app/models/Github.js
```diff
@@ -116,9 +129,9 @@ var GithubModel = Backbone.Model.extend({
           url: url,
           complete: function(res) {
             var r = {
-              file: file,
+              file: file.name,
               present: (res.status === 200) ? true : false,
-              json: res.responseJSON || []
+              json: res.responseJSON || yaml.parse(res.responseText)
```
dhcole: This throws an error when one of the config / navigation files is not available:

```
GET http://localhost:1337/site/dhcole/microsite-template/_navigation.json 404 (Not Found)send @ bundle.js:24812jQuery.extend.ajax @ bundle.js:24348(anonymous function) @ bundle.js:246(anonymous function) @ bundle.js:2335(anonymous function) @ bundle.js:1885_arrayEach @ bundle.js:1737_each @ bundle.js:1728async.forEachOf.async.eachOf @ bundle.js:1884_parallel @ bundle.js:2334async.parallel @ bundle.js:2348Backbone.Model.extend.fetchConfig @ bundle.js:260Backbone.Model.extend.initialize @ bundle.js:145Backbone.Model @ bundle.js:3259child @ bundle.js:4703Backbone.View.extend.initialize @ bundle.js:823Backbone.View @ bundle.js:4043child @ bundle.js:4703Backbone.View.extend.edit @ bundle.js:1194Backbone.Router.extend.edit @ bundle.js:47_.extend.execute @ bundle.js:4336(anonymous function) @ bundle.js:4324(anonymous function) @ bundle.js:4607_.some._.any @ bundle.js:45924_.extend.loadUrl @ bundle.js:4605(anonymous function) @ bundle.js:28(anonymous function) @ bundle.js:3183(anonymous function) @ bundle.js:46556triggerEvents @ bundle.js:3225triggerApi @ bundle.js:3211eventsApi @ bundle.js:3009Events.trigger @ bundle.js:3201_.extend.set @ bundle.js:3381options.success @ bundle.js:3448fire @ bundle.js:19281self.fireWith @ bundle.js:19393done @ bundle.js:24446(anonymous function) @ bundle.js:24787
bundle.js:47951 Uncaught <ParseException> You cannot define a sequence item when in a mapping
``

**Comment by dhcole** [comment]
Getting an error when saving a new post without a `title`

```
Uncaught TypeError: Cannot read property 'title' of null
Backbone.View.extend.saveDocument @ bundle.js:752
jQuery.event.dispatch @ bundle.js:20617
elemData.handle @ bundle.js:20303
```

dhcole:
Works well now. :ship:

PR #249, opened 2015-12-15
Upgrade node to LTS branch

jeremiak:
Update node in `package.json` as well as `.travis.yml` to the LTS (long term support) branch of Node. I also updated the `Dockerfile` but I did not publish it to dockerhub. @dhcole, would you mind publishing to dockerhub since I don't think I have access?


package.json
```diff
@@ -99,7 +99,7 @@
     "watchify": "^3.2.2"
   },
   "engines": {
-    "node": "0.10.41",
-    "npm": "2.1.x"
+    "node": "4.2.x",
```
dhcole: This should probably be pinned to a full version number. We had an issue when we were pinned to `0.10.x`, where when `0.10.41` was released, our builds started failing. That was because we pin the buildpack to a specific version and the buildpack did not have support for the newest version of node. If we pin both the buildpack and node to an exact version, we won't have that issue.

dhcole:
Yea, I'll test and push an update to the Docker hub.

dhcole:
Tested locally and as a docker image — works well. Merging, then I'll push the latest master to the Docker hub.

Issue #235, opened 2015-11-30
Implement new designs


dhcole:
@goodnightjj is working on new designs. This is a placeholder issue to implement those.

dhcole:
@jeremiak has begun this work in concert with #253. Moving to `Beta 2` to keep these together.

wslack:
I think this can be closed.

PR #251, opened 2015-12-16
New homepage design

jeremiak:
After our transition to the pattern library, we are ready to implement a new home page. This PR implements the new design and does a few other things along the way:

0. Implements `node-sass` as a dependency of Federalist
0. Breaks out existing CSS into very rough SASS modules, but now the project is set up to have those files be refactored in the future
0. Submits 18F/web-design-standards#930 so that the pattern library doesn't try and do things on its own after installation (namely CSS compilation)
0. Uses `autoprefixer` to apply browser prefixes to compiled CSS.

Refs 18F/web-design-standards#930 and #235


dhcole:
Tested locally and on Docker — this looks amazing. Nice work @jeremiak @goodnightjj and @GailSwanson 

I'll leave this open so others have a chance to review, then merge this afternoon.

dhcole:
:+1: from @GailSwanson via Slack, so :ship:

package.json
```diff
@@ -94,6 +102,7 @@
   "devDependencies": {
     "jsdom": "^3.1.2",
     "mocha": "^2.2.5",
+    "node-sass": "^3.4.2",
```
dhcole: @jeremiak Sorry, I missed this. Our production servers don't install `devDependencies`, so we're getting an error that `node-sass` is not available. OK for me to move this up to regular dependencies?

PR #262, opened 2015-12-17
Move node-sass to dependencies

dhcole:
Our production servers don't install devDependencies, so we're getting an error that node-sass is not available.


dhcole:
Tested locally.

PR #263, opened 2015-12-17
Lock files while editing

### LINKED ISSUE #190: Lock files that are being edited
Use the webhook integration we have to show a message when users try to edit a post another user has open. Have a mechanism for pinging the user who is editing about requesting access. Time out long-running edit sessions, but stash unsaved work somewhere.

dhcole:
To close #190 

This locks the editing form on a page that another use is currently editing to prevent edit conflicts. It subscribes all sockets connected to a page to a room for that page and sends event when uses open or leave the page. If a user is the first in line, they are unlocked. All events take place in realtime without storing data in the main database. After a page is edited and released, the unlocked user's view refreshes to make sure they have the latest content.

Next steps:
- release locks after a period of time
- include user name in lock message so users can collaborate


jeremiak:
This looks great with the masks. Also, great work on the refresh mechanism!

PR #269, opened 2015-12-29
Build log table

dhcole:
Updates the build log view and template to show more data as a table.

screen shot 2015-12-29 at 11 43 42 am


jeremiak:
Looks great to me!

PR #271, opened 2015-12-30
Fix vertical centering in nav bar

### LINKED ISSUE #265: Homepage navbar not vertically centered
Homepage is looking slick!

But the logo and link in the header navbar isn't centered vertically. After making a few tweaks, I was able to center it with the following code:

```
.logo a {
  height: 29px;
}

.nav-mobile li:after {
  margin-bottom: 0;
}

nav {
  padding-top: .45em;
  padding-bottom: .45em;
}
```

I'm not familiar with how your site is organized otherwise would have made a PR but feel free to use the above.

This is what it looks like now:

screen shot 2015-12-22 at 3 08 25 pm

This is what it looked like before:

screen shot 2015-12-22 at 2 57 21 pm

dhcole:
Closes #265. Thanks @maya for the fix.

### DIFF
--- a/assets/styles/sass/_header.scss
+++ b/assets/styles/sass/_header.scss
@@ -2,7 +2,7 @@ header {
   background-color: $header-bg-color;
 }
 
-$nav-padding: .2em;
+$nav-padding: .45em;
 
 nav {
   padding-bottom: $nav-padding;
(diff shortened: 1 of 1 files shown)


jeremiak:
LGTM!

PR #274, opened 2015-12-30
Fix table cell wrapping

dhcole:
Long build messages break the table layout. This fixes it.

screen shot 2015-12-30 at 3 10 04 pm


jeremiak:
:+1:

Issue #266, opened 2015-12-28
Extra horizontal rules are problematic for the editor


jeremiak:
Having horizontal rules along with YAML front matter poses some issues for the current editor. For example, take a look at

* The file: https://github.com/jeremiak/microsite-template/commit/cc8863275a59236ac15c0638f2b2f5948e1de808
* The file in the editor: https://federalist.18f.gov/#edit/jeremiak/microsite-template/gh-pages/README.md ([Screenshot](http://up.jeremiak.com/eB5T))
Steps to reproduce:

0. Ensure the markdown document has front matter
0. Add a horizontal rule in the body of the text
0. Save and reload the editor

PR #275, opened 2015-12-30
New fields for some settings

### LINKED ISSUE #254: Form fields for metadata
Per recent designs, common metadata (page settings) should be presented as metadata fields.

jeremiak:
This modifies the editor experience so that some fields of the YAML front matter in each post is displayed using form fields. These fields are "whitelisted" within the main editor view and configured to be represented as a certain type. The supported types are:

0. `text` - will be displayed with a standard text box
0. `boolean` - will be displayed with True/False radio buttons
0. `select` - will be displayed as a select, expects an array as the value for `options`
0. `date` - will be a calendar picker and stored as an ISO string (i.e. `"2015-12-30"`)

The specifically whitelisted fields are:

0. `title` as `text`
0. `layout` as `select`
0. `author` as `text`
0. `published` as `date`

If there are fields in addition to these, they will be displayed in the CodeMirror editor (as all settings were previously). If a whitelisted field is absent it will be ignored. 

Also adds a series of tests to unit test some of the backbone views.

To close #254 #266


assets/app/views/editor/edit-file.js
```diff
@@ -22,55 +23,139 @@ var EditorView = Backbone.View.extend({
...
-        file      = [
-                      self.model.get('owner'),
-                      self.model.get('repoName'),
-                      self.model.get('branch'),
-                      self.model.get('file')
-                    ].join('/'),
-        raw, content, settingsEditorEl, contentEditorEl;
+    var self      = window.z = this,
+        file      = this.filePathFromModel(this.model),
+        fileExt   = this.model.get('file').split('.')[1],
```
dhcole: Is this stable? Is `file.name.md` a valid file name?

jeremiak: Good point and catch. No, it isn't. For instance, it will break on `.travis.yml` as well. 

I'll fix that up

test/browser/unit/data/horizontal-rule.md
```diff
@@ -0,0 +1,12 @@
+---
+title: 'Fake document'
+layout: 'fake'
+---
+
+# Heading 1
+
+The report, by the World Anti-Doping Agency, implicates athletes, coaches, trainers, doctors and various Russian institutions. The report recommends that Russia be suspended from competition.
```
dhcole: Does another source hold a copyright to this?

assets/app/views/editor/edit-file.js
```diff
@@ -22,55 +23,139 @@ var EditorView = Backbone.View.extend({
...
+        type: 'text'
+      },
+      layout: {
+        type: 'select',
+        options: this.model.getLayouts()
+      },
+      author: {
+        type: 'text'
+      },
+      published: {
```
dhcole: [`published`](http://jekyllrb.com/docs/frontmatter/#predefined-global-variables) is actually a reserved attribute in Jekyll. It's a boolean for whether a post or page should render. [`date`](http://jekyllrb.com/docs/frontmatter/#predefined-variables-for-posts) is the key for the publish date for posts, and while it wouldn't do anything by default for pages, we could still use it.

assets/app/views/editor/edit-file.js
```diff
@@ -246,14 +299,94 @@ var EditorView = Backbone.View.extend({
...
+    else if (this.editors.content) {
+      // CodeMirror is loaded as content editor
+      content = this.editors.content.doc.getValue();
+    }
+
+    return content;
+  },
+  getSettingsDisplayStyle: function (doc) {
+    if (doc.fileExt == 'md' || doc.fileExt == 'markdown') {
+      if (doc.frontMatter == '') return 'regular';
```
dhcole: This is ambigous: do you want to only test for `doc.frontMatter` strictly equals (`===`) `''`, or do you also want the test to pass for other falsy values, such as `false == ''` or `false == ''`?

assets/app/views/editor/edit-file.js
```diff
@@ -246,14 +299,94 @@ var EditorView = Backbone.View.extend({
...
+      else return 'whitelist';
+    }
+    return 'only';
+  },
+  toIsoDateString: function (date) {
+    var d = (date) ? new Date(date) : new Date();
+    return d.toISOString().substring(0, 10)
+  },
+  filePathFromModel: function (model) {
+    return path = [
```
dhcole: is the `path =` assignment necessary here, or do you only want to return the joined string?

dhcole:
@jeremiak posted some comments / questions to you for your review, as you have time. I also added a commit to fix a small scope issue and fix some things that broke the lint process.

Are you linting locally? We may just have different settings. I'm using https://atom.io/packages/linter with https://atom.io/packages/linter-jshint and the default settings.

jeremiak:
Thanks @dhcole I'll address those now. I wasn't linting locally at all, now set up with the linter package you use as well.

test/browser/unit/data/horizontal-rule.md
jeremiak: Not sure, but I'll just replace the test content with ipsum lorem to make sure

dhcole:
Looks great, @jeremiak. Just one thing: how do you define a list of layouts?

jeremiak:
@dhcole Right now you define it in your `_config.yml` ([code](https://github.com/18F/federalist/blob/pull/metadata-fields/assets/app/models/Github.js#L205-L211)). You can define the `layout` key as a list. For example, in your defaults you could have:
```yml
layout:
- post
- home
- form
```

dhcole:
Ok, looks great! Nice work @jeremiak :ship:

PR #290, opened 2016-01-14
Add page layout to options if not already there

jeremiak:
The editor should never prevent you from saving the document in the same same state that it was retrieved in. It wasn’t passing along the current value of the layout value so if it wasn’t configured in the defaults, the editor would change the value to “default”.

@dhcole ready for your review


dhcole:
:+1:

PR #279, opened 2016-01-07
Clone new sites instead of forking them

dhcole:
This adds support for cloning sites through the client-side GitHub model and a `/site/clone` API endpoint.

The process works by checking permissions and creating a new repo on the client, then triggering a special build with source repository information. The source build pulls from the source repo and pushes to the destination repo before building.

- [x] Depends on https://github.com/18F/federalist-docker-build/pull/3
- [x] Needs a front-end form for capturing the new site's name

Resolves #138

PR #291, opened 2016-01-14
Clone take 2

dhcole:
Reimplements #279 with bug fixes found after setting up a test environment in Cloud.gov.


dhcole:
Had to get the git history in order, but it's all set now. Most of this is the same as last time. Notable changes are dea2e024bf56f91d8d07e4791a47206a9991f43e and c0302467f5e63b4c5dd70f2c603bc4cb1d42da00, which I think was the main issue with the last deployment.

jeremiak:
Looks good!

PR #292, opened 2016-01-14
If an AWS redirect header is present on preview, redirect the request

### LINKED ISSUE #236: Previews don't follow S3 redirect files
Redirect files for paths without trailing slashes just get downloaded instead of redirected on previews. We need to adjust the node.js proxy to check the redirect headers and follow them.

dhcole:
For the preview proxy, this looks for the AWS S3 redirect header and issues an actual redirect if present.

Fixes #236


jeremiak:
Meant to make the comment here, looks good to me!

PR #296, opened 2016-01-25
Add commit details to site object

### LINKED ISSUE #136: Include build info in Jekyll site data?
What's really nice about the `site.branch` data that @dhcole added is that I can drop that into my HTML and style elements by branch (e.g., to add a little "staging" or "dev" banner).

I think it'd be nice to also include some other build info, but I don't know what else is available. Maybe time the build started, so we can look at that to determine whether a build is "stale" (e.g. if there was a build error and we can't view the build log on Federalist for some reason?).

dhcole:
Adds commit details to the `site` object.

eg:

```json
commit: {
 "commit": "8ebcd8d344ac417288304ae7cba630fae3dbe1b6",
 "author": "Dave Cole <david.cole@gsa.gov>",
 "date": "Mon Jan 25 12:13:47 2016 -0500",
 "message": "Test message"
}
```

See also https://github.com/18F/federalist-docker-build/pull/4
Closes #136

### DIFF
--- a/api/hooks/buildEngine/index.js
+++ b/api/hooks/buildEngine/index.js
@@ -23,7 +23,9 @@ var hook = {
       'rm -rf ${destination} || true',
       'mkdir -p ${source}',
       clone,
-      'echo "baseurl: ${baseurl}\nbranch: ${branch}\n${config}" > ' +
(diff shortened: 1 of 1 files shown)


jeremiak:
@dhcole Looks good. This is the same functionality that is in the docker image but duplicated in the app as well, correct? The reason for the duplication is for the case where the app is not configured to use SQS and the docker builder.

dhcole:
Exactly, this adds the commit data for local builds run by the `buildEngine`

PR #301, opened 2016-01-28
Use xhr polling instead of websockets

### LINKED ISSUE #300: Websockets don't seem to be working anymore
After saving a draft, I don't get a response over the websocket for the draft build completing, requiring a refresh to see the preview button. Also, when opening the same editor page in two tabs, I don't see the lock alert message.

This could be due to some upstream change in how websockets are routed in Cloud Foundry.

One possible solution is to adjust Sails's socket configuration to use JSON polling instead. http://sailsjs.org/documentation/anatomy/my-app/config/sockets-js

dhcole:
Our production environment seems to have some trouble with websockets, but we probably don't need them anyway. This also fixes an issue where the page lock wasn't being set on first load.

Fixes #300

### DIFF
--- a/assets/app/views/editor/edit-file.js
+++ b/assets/app/views/editor/edit-file.js
@@ -59,7 +59,7 @@ var EditorView = Backbone.View.extend({
       self.socket = data.id;
 
       // Apply the lock
-      self.lockContent.bind(self);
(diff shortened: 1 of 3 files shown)


jeremiak:
:+1:

PR #303, opened 2016-01-29
Load user first, sites second, and then route pages

### LINKED ISSUE #252: View loading sometimes flashes content
The app loads views as AJAX calls return, so users end up seeing flashed of the homepage / add a site / site listings as the page loads. This is pretty minor but could be cleaned up.

### LINKED ISSUE #255: Possible race condition loading page
For instance when viewing a site settings, sometimes an error is returned if you refresh the page. This error blocks the main view from loading.

### LINKED ISSUE #282: Build logs frequently don't display
I'm looking at the [EITI build logs](https://federalist.18f.gov/#site/73/builds) frequently to figure out JS minification in production, but more often than not the page loads and no logs are displayed. There aren't any errors in the console, but I can see in the Network tab that the `https://federalist.18f.gov/v0/user/2` URL has loaded successfully. I don't know if the data is being displayed client-side or not, but if it is then this feels like a race condition to me.

dhcole:
This should fix some odd issues around loading pages including pages not loading due to data not being available as well as views flashing between unauthenticated and authenticated.

Instead of asynchronously loading user and sites data, this loads the user first and then sites, or if the user is not logged it, it doesn't load sites and instead loads the unauthenticated homepage view.

Fixes #282
Fixes #255 
Fixes #252

### DIFF
--- a/assets/app/app.js
+++ b/assets/app/app.js
@@ -12,20 +12,36 @@ var encodeB64 = require('./helpers/encoding').encodeB64;
 
 var Router = Backbone.Router.extend({
   initialize: function () {
-    var self = this;
 
-    this.sites = new SiteCollection();
     this.user = new UserModel();
 
-    this.navbarView = new NavbarView({ model: this.user });
-    this.mainView = new MainContainerView({ user: this.user, collection: this.sites });
-
-    this.listenToOnce(this.user, 'change', function () {
+    this.listenToOnce(this.user, 'sync', function (user) {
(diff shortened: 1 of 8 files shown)


dhcole:
Added a loader image. I'm sure @jeremiak can make something much more exciting — perhaps with the logo.

dhcole:
To test this with production network conditions and data, https://federalist-test2.18f.gov/

PR #306, opened 2016-02-02
Pull/static homepage

jeremiak:
This is an extension of the PR in #303, but alters some of the behavior. Currently, the homepage of Federalist is not returned on the initial request. In fact, every single URL on Federalist is built by the Backbone app. This means that a device needs JavaScript enabled just to learn about Federalist and see the homepage, let alone use the service.

I've altered the initial index so that the HTML returned by the server on the initial request for `index.html` has all of the information (including SVGs for animation) required to display it.

In #303, the initial homepage is still built by Javascript on the client and has a loading animation which feels a bit inappropriate for a static/marketing homepage.


assets/app/app.js
```diff
@@ -12,20 +12,39 @@ var encodeB64 = require('./helpers/encoding').encodeB64;
...
+          this.navbarView = new NavbarView({ model: this.user });
+          this.mainView = new MainContainerView({
+            user: this.user,
+            collection: this.sites
+          });
+          this.navbarView.render();
+          Backbone.history.start();
+        }.bind(this)
+      });
+    }.bind(this));
```
adborden: I think the `var self` trick works better here, otherwise adding an additional nested function requires `bind` which seems error prone to me.

adborden:
OMG loads so fast :+1:

assets/app/app.js
jeremiak: Yup, I also like the `self` method. Will update

PR #309, opened 2016-02-10
Run integration tests on Travis

adborden:
Run the webdriver based integration tests on Travis via Sauce Labs.

Allows the github test user to be configured via environment variables.

Uses [Page Objects](http://martinfowler.com/bliki/PageObject.html) to abstract the webdriver code.

Includes some basic tests on creating a new site based on a template.


jeremiak:
Great, really excited to get this testing framework in use so that we can be more confident that we're not making breaking changes.

PR #310, opened 2016-02-10
Use config template

adborden:
Simplifies the configuration process by including a local config sample for devs to edit.


jeremiak:
:+1: 

Merging even though its red because its just README changes

PR #313, opened 2016-02-10
Change uswds branch to latest v0.9.x

uranther:
Running `npm install` was failing using the uswds `18f-pages-staging` branch as configured in `package.json`. Changing it to `v0.9.x` per @jeremiak suggestion, and it runs without failing now.


jeremiak:
Sweet, thanks for the PR @uranther!

PR #318, opened 2016-02-16
Test generic file listing

### LINKED ISSUE #314: Test that a user can access/walk the directory structure of a generic Jekyll site
If a repository/site is not configured to use the navigation builder, Federalist should enable a user to walk the file structure by editing files that have a `.md` file extension and by listing directory contents otherwise.

Steps:

0. Authenticated user has a site similar to the `microsite-template`, which lacks a `/_navigation.json` and therefore should be displayed as a file listing
0. The user should be presented with a file listing that has an "Open" button for directories and an "edit" button for files. If the file is a markdown file, it should open in the Federalist editor. Otherwise, it should open in Github.
0. Clicking "Open" on a directory should result in a file listing that behaves in the same way.

Screencast:

jeremiak:
Federalist can be enhanced by the Pages schema but it should not break for sites that are not configured to use it. In the case that a site does not have any extra data to tell Federalist how to display the file listing and enable users to walk the directories of the project and edit markdown files.

To close #314


test/integration/fileListing.test.js
```diff
@@ -0,0 +1,89 @@
...
+var fileListingPage;
+
+describe('generic repository file listing tests', function () {
+  this.timeout(15000);
+
+  before(function () {
+    var engine = sails.hooks[sails.config.build.engine];
+    engine.jekyll = function (model, done) {
+      done();
+    };
```
adborden: You can rewrite this with `sinon` as 

```javascript
sinon.stub(engine, 'jekyll').yields();
```

adborden: You probably also want to restore the stub in the cleanup

```javascript
after(function() {
  engine.jekyll.restore(); // with sinon
});
```

test/integration/fileListing.test.js
```diff
@@ -0,0 +1,89 @@
...
+          repository: 'microsite-template',
+          users: [u[0].id]
+        };
+
+        Site.create(data).exec(function (err, s) {
+          if (err) throw err;
+          site = s;
+          done();
+        });
+      });
```
adborden: How do you feel about promises?

The docs are very brief on the topic, but Waterline and the Sails models use promises too. ([Bluebird](http://bluebirdjs.com/docs/getting-started.html) under the hood).

```javascript
before(function () {
  return User.find({...})
    .then(function (user) {
      // ...
      return Site.create();
    })
    .then(function (s) {
      site = s;
    });
});

File: test/integration/fileListing.test.js [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+    after(function (done) {
+      Site.destroy({ id: site.id }).exec(function(err){
+        if (err) throw err;
+        done();
```
adborden: `done` is just a node style callback, so you can pass the error to `done` or even just pass `done` as the callback:

```javascript
after(function (done) {
    Site.destroy({ id: site.id }).exec(done);
});

adborden:
:+1:  Some food for thought, but good to merge as is.

Issue #311, opened 2016-02-10
Unable to add a repo


shawnbot:
Yesterday I tried adding [18F/frontend](/18F/frontend) to Federalist and got a mysterious error that was just a stringified object:

```
[object Object]
```

In the console I see:

![image](https://cloud.githubusercontent.com/assets/113896/12961776/00a76d00-cff7-11e5-879c-7248e49c0c2e.png)

shawnbot:
FWIW, the issue ended up being that I didn't have admin access to the repo on github, so Federalist refused to add the project because it couldn't add a webhook using my credentials.

PR #321, opened 2016-02-18
Show text from error when adding site

jeremiak:
The poor UI that is reported in #311 is fixed and tested. The actual error message was under a `raw` key in `responseJSON`. The error handler method was slightly refactored so that the message generation was easily testable.

Adds tests to the front end suite to test the message returning function.

Reorganizes `/test/browser` and renames `npm` command.


adborden:
:+1: I'll merge this first thing Monday.

adborden:
Oh, the [test failure](https://travis-ci.org/18F/federalist/builds/110412465#L1213) looks related.

jeremiak:
@adborden All fixed and green

PR #322, opened 2016-02-18
Update sauce with 18f blessed account

### LINKED ISSUE #319: Swap Sauce Labs credentials in travis for 18F account
We have an 18F account, but I used my "free trial" account for Sauce Labs because I didn't have the real ones at the time I committed this work. Should be as simple as swapping [SAUCE_USERNAME](https://github.com/18F/federalist/blob/master/.travis.yml#L24) and [SAUCE_ACCESS_KEY](https://github.com/18F/federalist/blob/master/.travis.yml#L23).

adborden:
Fixes #319


jeremiak:
:+1:

Issue #326, opened 2016-03-01
508 Issues from pa11y-rails


nickbristow:
Results can be recreated using [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/)

## Results for https://federalist.18f.gov:
* __Error:__ The html element should have a lang or xml:lang attribute which describes the language of the document.
 * WCAG2AA.Principle3.Guideline3_1.3_1_1.H57.2
 * html
 * `<html><head><style> .ProseMirror-menu...</html>`

## Summary:
* 1 Errors

PR #329, opened 2016-03-10
Include lang attr in html tag

jeremiak:
To address the pa11y issues @nickbristow raised in #326


jeremiak:
Merging this myself since its a really small change and the build is green.

PR #337, opened 2016-03-24
Use npm published version of uswds

jeremiak:
Now that the Draft U.S. Web Design Standards is published to npm, we should use that. The npm build scripts have been updated to use the `/dist` files in the package.

This should fix the build and deploy. @adborden can I get a :+1: or :-1: and I'll merge afterwards?


package.json
```diff
@@ -88,12 +88,9 @@
...
     "uswds": "npm run uswds:css && npm run uswds:images && npm run uswds:fonts",
-    "uswds:css": "npm run uswds:css:sass && npm run uswds:css:uswds && npm run uswds:css:gfonts",
-    "uswds:css:sass": "cd node_modules/uswds && npm run build-sass && cd ../..",
-    "uswds:css:uswds": "cp node_modules/uswds/assets/css/uswds.css assets/styles/",
-    "uswds:css:gfonts": "cp node_modules/uswds/assets/css/google-fonts.css assets/styles/",
-    "uswds:images": "cp -r node_modules/uswds/assets/img/ ./assets/images/uswds",
-    "uswds:fonts": "cp node_modules/uswds/assets/fonts/* ./assets/fonts",
+    "uswds:css": "cp node_modules/uswds/dist/css/uswds.css assets/styles/",
+    "uswds:images": "cp -r node_modules/uswds/dist/img/ ./assets/images/uswds",
+    "uswds:fonts": "cp node_modules/uswds/dist/fonts/* ./assets/fonts",
```
adborden: I've used symlinks in the past, or if your build system supports it, I'll add an include path to the module's directory so sass/whatever can find it.

I don't have a great sense of what approach is best.

adborden:
:+1:

package.json
adborden: Here's an example https://github.com/18F/fec-eregs/blob/53a5983/Gruntfile.js#L11

Maybe copy is the simplest and the most straight forward.

PR #339, opened 2016-04-06
Decode b64 build status message

jeremiak:
To prevent problems with character encoding, the status message from the docker builder will be base 64 encoded as of 18f/federalist-docker-build@b43ee49.

Should be merged with 18f/federalist-docker-build#6

### DIFF
--- a/api/controllers/BuildController.js
+++ b/api/controllers/BuildController.js
@@ -5,14 +5,18 @@
  * @help        :: See http://links.sailsjs.org/docs/controllers
  */
 
+function decodeb64 (str) {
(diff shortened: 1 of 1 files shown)


adborden:
:+1:

wslack:
@jeremiak should this be put back in or closed

jeremiak:
@wslack This is something that should be merged. It fixes an issue that we ran into trying to onboard some sites into Federalist and is probably a good idea to send around shell output encoded in one form or another. Since this is all done, I'd merge it sooner rather than later.

wslack:
This is pending merge until we set up a warning for users.

PR #349, opened 2016-04-15
link directly to the chat room

afeld:
https://chat.18f.gov now supports linking directly to the channel! See https://github.com/18F/chat/pull/5 for more information.


jeremiak:
Sweet, thanks @afeld

PR #358, opened 2016-05-25
update note about Sauce Labs

wslack:
cc @jeremiak to check me


el-mapache:
@wslack @jeremiak  Is this still something we should go ahead and merge?

jeremiak:
I don't think we should. This guidance now seems out of date since we're not requiring integration tests on pull requests and it might be confusing to suggest otherwise

wslack:
@jeremiak what should we say? The current guidance is still confusing for outside contributors.

wslack:
Update: @jeremiak is going to rebase and merge this!

PR #360, opened 2016-05-27
Add script to export all current sites

jeremiak:
There is a business need to know which sites are currently on Federalist and which users we should get in touch with regarding downtime or maintenance.

I have added a script that is executable with `npm run export:sites`. You can pass an optional path after `--` to have the outputted CSV written to a location of your choice. Since we will be running this on cloud.gov instances we probably just want to use the default which is `./current-sites.csv`

I don’t think we can get into production cloud.gov postgres instances from the outside, so we’ll have to run this script on a throwaway instance using something like `cf-ssh` and then run the script with `npm run export:sites`. We can then retrieve the file with [these instructions](https://docs.cloud.gov/apps/databases/#accessing-files-created-on-the-server-using-cf-files) and throwaway the instance.

This one is for you @wslack so let me know if this is useful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "uswds:images": "cp -r node_modules/uswds/dist/img/ ./assets/images/uswds",
     "uswds:fonts": "cp node_modules/uswds/dist/fonts/* ./assets/fonts",
     "codemirror": "cp node_modules/codemirror/lib/codemirror.css assets/styles/",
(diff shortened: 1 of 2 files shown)


wslack:
@jeremiak this LGTM though I didn't run locally. should this get code review from @el-mapache?

jeremiak:
@wslack Sure, it would be good to have it run and make sure it works but I tested and the script won't execute by itself so there's no real risk in merging it and then testing it for real.

PR #361, opened 2016-05-28
test push

testing-tester:
test from @el-mapache


wslack:
per standup, @el-mapache is to update this PR with language clarifying the use of Selenium when running an outside PR (maybe in contributing.md?)

Issue #336, opened 2016-03-22
Landing page doesn't display errors


alexose:
Not a huge deal, but when I'm directed to `https://federalist.18f.gov/?error=Error.Passport.Unauthorized`, there's no indication that anything is wrong other than the URL.

wslack:
@alexose thank you! how did you get sent to that URL? Was it an attempt to auth via GitHub that failed?

alexose:
Yep, exactly!

wslack:
Do you mind sharing a screenshot of what you see? Does it appear to be the homepage?

wslack:
@alexose we should be all good now I think! want to verify and close?

alexose:
Look good-- I went to the error URL directly and I'm seeing an informative message :)

Thanks!

PR #365, opened 2016-06-01
Fixes bug where alerts dont appear

el-mapache:
Relates to https://github.com/18F/federalist/issues/336

* Fixes bug where alerts don't appear properly on main page
* Fixes console error `can't read property 'dashboard' of undefined`
* Adds home view to handle edge case of page being refreshed when error query string is present in URL

### DIFF
--- a/assets/app/views/main.js
+++ b/assets/app/views/main.js
@@ -4,10 +4,13 @@ var querystring = require('querystring');
 
 var SiteListView = require('./list');
 var SiteView = require('./site/site');
+var HomeView = require('./home');
 
-var AppView = Backbone.View.extend({
+var MainView = Backbone.View.extend({
   el: 'main',
   initialize: function (opts) {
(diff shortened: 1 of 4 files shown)


assets/app/views/main.js
```diff
@@ -34,8 +37,13 @@ var AppView = Backbone.View.extend({
     $('.alert-container').html('');
   },
   setAlert: function (text) {
+    var newAlert = $('<div/>', {
+      class: 'usa-alert usa-alert-error',
+      role: 'alert'
```
jeremiak: Nice! I didn't know about this jQuery API but it seems like a much cleaner way to generated HTML elements 👍

el-mapache: Yeah! I had forgotten my jQuery syntax and was googling when I came across it. Not sure when they added it, but it reads way nicer than super long strings.

jeremiak:
Nice work!

PR #366, opened 2016-06-01
ato-owasp-clickjacking

el-mapache:
* Adds x-frame-origin header to prevent clickjacking, per OWASP


el-mapache:
Closing as nginx provides this header already.

PR #367, opened 2016-06-03
Removes socket init code

el-mapache:
* To prevent a flood of errors, removing socket code until we have time to troubleshoot
* Should also makes our owasp scans alert free


el-mapache:
Socket code is tied to initial loading of sites from github, it seems. If this channel `/v0/build?limit=0` isn't hit, some integration tests fail.

el-mapache:
Closing as it requires more work to fix than is worth doing, given the refactor.

PR #369, opened 2016-06-06
Set secure cookie flag in production

el-mapache:
* Sets secure cookie flag in production to comply with ATO requirements

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -5,6 +5,13 @@ var AWS = require('aws-sdk'),
     AWSCreds = appEnv.getServiceCreds('federalist-aws-user'),
     redisCreds = appEnv.getServiceCreds('federalist-redis');
 
+var _ = require('underscore');
+var session = {
+  cookie: {
+    secure: true
+  }
+};
+
 /**
  * Production environment settings
  *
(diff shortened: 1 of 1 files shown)


config/env/production.js
```diff
@@ -47,13 +54,14 @@ if (AWSCreds) {
...
 // If running in Cloud Foundry with a redis service
 if (redisCreds) {
-  module.exports.session = {
-    adapter: 'redis',
+  session = _.extend({}, session, {
+    adapter: 'socket.io-redis',
     host: redisCreds.hostname,
     port: redisCreds.port,
-    db: 0,
+    db: 1,
```
jeremiak: Why does `db` get set to `1`?

config/env/production.js
```diff
@@ -47,13 +54,14 @@ if (AWSCreds) {
 
 // If running in Cloud Foundry with a redis service
 if (redisCreds) {
-  module.exports.session = {
-    adapter: 'redis',
+  session = _.extend({}, session, {
+    adapter: 'socket.io-redis',
```
jeremiak: Why use the socket adapter instead of just regular redis?

config/env/production.js
el-mapache: No clue, it was set to '1' in the original file! I guess that is db it uses in production? Since its a magic number, it should at least be in a variable (and probably be configurable), but I didn't want to mess with it without being able to test it.

config/env/production.js
el-mapache: This was in the original code. I don't know why it's set to use that particular adapter, but it's not something I feel comfortable changing without being able to test the outcome.

el-mapache: you were totally right, I looked again at the diff and saw that I removed the non-socket redis session information. This should be correct now.

PR #371, opened 2016-06-07
Use index to properly output all current sites

jeremiak:
Fixes the `npm run export:sites` script to properly output all the current users and sites as a CSV. I forgot to use a dynamic index so the exported sites were all copies of the first site.


jeremiak:
Merging my own code because we already used this script (via `cf-ssh` to get the data we needed and we have too many open pull requests.

Issue #356, opened 2016-05-02
Production website URLs are broken from "View website" buttons


jeremiak:
Currently, there is a "View website" button on both the settings and the log views but the link points to an empty page. The URL is `https://federalist.18f.gov/site/:owner/:repoName/`.
Screenshot: http://up.jeremiak.com/g2Sq

The "Visit Site" link on the site listing page goes to a different, functioning URL: `http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/:owner/:repoName/`.

The button links should be changed to point to the proper S3 bucket. Thanks to @anthonygarvan for finding this one.

PR #373, opened 2016-06-08
[FED-356]: Normalizes viewLink href across buttons

el-mapache:
Addresses issue https://github.com/18F/federalist/issues/356

* Adds viewLink property to toJSON method of `Site` model
* Uses this property to provide an external site url across buttons

### DIFF
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -5,6 +5,8 @@
 * @docs        :: http://sailsjs.org/#!documentation/models
 */
 
+var DEFAULT_BUCKET = 'federalist.18f.gov.s3-website-us-east-1.amazonaws.com';
+
 module.exports = {
   // Enforce model schema in the case of schemaless databases
   schema: true,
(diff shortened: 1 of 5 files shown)


jeremiak:
Nice, its good to have this variable in one place. Those URLs aren't working for some reason though, and  I think we actually want the production build URLs (in the case where a custom URL isn't set) to be something more like `https://s3.amazonaws.com/${awsBucketName}/site/${owner}/${repository}/index.html`

So far though the sites I have checked out look broken at this URL because assets generally aren't delivered properly. It makes me think that we want to supply a baseurl for the Jekyll site if there isn't a custom url when the build is on the default branch.

el-mapache:
I was under the impression that the siteRoot was always set by the app?

When you say it doesn't work, what do you mean? Locally, for me, I have never been able to click 'view site' as the URL always points to my localhost. What I mean is, this PR doesn't actually address not being able to view the site on the internet from a local machine; I was under the impression that never worked.

el-mapache:
I just hardcoded the default path to be the one mentioned in the issue.

jeremiak:
Cool, we might want to make this a config variable or something later but for now this is great!

Issue #372, opened 2016-06-08
Uncaught TypeError in editor for customer


wslack:
![screen shot 2016-06-08 at 12 09 43 pm](https://cloud.githubusercontent.com/assets/10016577/15901683/45c63e4c-2d72-11e6-955a-fe1a3916bb6e.png)

The issue is not caused by the markdown file.

URL of breaking page:
https://sbst.gov/about/

URL of working page:
http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/wslack/wslack-demo4/about/

- [Working page](https://github.com/wslack/wslack-demo4/blob/master/pages/about.md)
- [Non working page](https://github.com/us-federal-sbst/federalist-modern-team-template/blob/master/pages/about.md)

wslack:
https://github.com/us-federal-sbst/federalist-modern-team-template/issues/53 has fixed the bug I saw, fwiw.

PR #374, opened 2016-06-09
[FED-372]: Attempt to fix function call on undefined permalink

el-mapache:
References issue https://github.com/18F/federalist/issues/372

I couldn't load an actual copy of the customer's site, but I did go in and clean up some of the code around setting a permalink on the template data object. The changes should make the code a bit more failsafe, but I don't know if it will fix the error without more information.

* Checks for existence of data.file and defined `permalink` property on navigation object

### DIFF
--- a/assets/app/views/site/pages/pages.js
+++ b/assets/app/views/site/pages/pages.js
@@ -101,16 +101,23 @@ var EditView = Backbone.View.extend({
       file: model.get('file'),
       branch: model.get('branch')
     };
-    var r = new RegExp(/\//g);
-    var n;
+    var navigationJSON = model.configFiles['_navigation.json'];
+    // matches all '/' in a string.
+    var matchSlashes = new RegExp(/\//g);
+    var currentPageObj;
 
     if (data.draft) {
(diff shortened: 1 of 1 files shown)


assets/app/views/site/pages/pages.js
```diff
@@ -101,16 +101,23 @@ var EditView = Backbone.View.extend({
...
 
-    if (model.configFiles['_navigation.json'].present) {
-      n = _.where(model.configFiles['_navigation.json'].json, { href: data.file });
-      data.permalink = n.length ? n[0].permalink.replace(r, '') : n;
+    // If there is no file, we are on the tree page, and there won't be any
+    // permalink to compare to.
+    if (navigationJSON.present && data.file) {
+      var permalink;
+
+      currentPageObj = _.findWhere(navigationJSON, { href: data.file });
```
jeremiak: Just reading this through, but shouldn't `navigationJSON` actually be `navigationJSON.json`?

el-mapache: yep! good catch

jeremiak:
Nice, this definitely makes the code clearer!

Issue #294, opened 2016-01-21
New page button seems broken


dhcole:
Not sure what happened here, but clicking the "Add a new page" button throws an error:

<img width="614" alt="screen shot 2016-01-21 at 2 39 32 pm" src="https://cloud.githubusercontent.com/assets/170641/12492127/d0ac30cc-c04c-11e5-8342-d7121013521b.png">

@jeremiak any idea what might be causing this?

dhcole:
I believe this is also addressed in @jeremiak's latest branch for #253

PR #375, opened 2016-06-10
Fed 294

el-mapache:
Fixes issue here https://github.com/18F/federalist/issues/294

* Allows users to create news pages
* Refactors document model and editor view to better separate concerns

### DIFF
--- a/assets/app/models/Document.js
+++ b/assets/app/models/Document.js
@@ -20,7 +20,7 @@ var DocumentModel = Backbone.Model.extend({
       this.initializeYamlDocument(opts.content);
     }
     else {
-      this.initializeMarkdownDocument(['---', 'key: value','---', 'whatever content'].join('\n'))
+      this.initializeMarkdownDocument(['---','---'].join('\n'))
     }
 
     return this;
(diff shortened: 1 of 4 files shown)


assets/app/models/Document.js
```diff
@@ -2,27 +2,68 @@ var _ = require('underscore');
 var Backbone = require('backbone');
 var yaml = require('yamljs');
 
+var fileTypes = {
+  'md': ['md', 'markdown'],
+  'yml': ['yml', 'yaml']
+};
+
+var yamlDefaults = [
```
jeremiak: I don't think we want to hardcode these values (`yamlDefaults`) into this view because it can change from site to site. I think any site developer should be able to specify which layouts are available to the federalist editor by specifying in the `_config.yml` file.

jeremiak:
Nice, breaking out that editor view into multiple functions is great and is much more understandable! I left a comment about some hardcoded values but otherwise I think this looks great.

assets/app/models/Document.js
el-mapache: I think these defaults are only here as a fallback. On line 27 we initialize the document with opts.content in the file is a yml file. We only use these if no content is provided. Do you think it should be changed to an empty string?

jeremiak: I think so, because if there is a layout that doesn't have a corresponding file Jekyll will break the build and it will be hard for folks without that context to know how to fix it

PR #378, opened 2016-06-14
Replace deployment user password

jeremiak:
Deploys are currently broken because the deploy user's password expired. This should fix it with a new password for cloud.gov

Once guidance is written per 18F/cg-product#32, we'll want to follow that and use tokens.


jeremiak:
Merging to try deployment password change.

PR #379, opened 2016-06-15
Passes X-Forwarded-For header

el-mapache:


jeremiak:
Merged to see if this fixes a bug where login is unavailable for all users

PR #381, opened 2016-06-15
Adds fetch, wrapper code

el-mapache:
Let me know what you think @jeremiak  about the fetch wrapper and the changes I've been making. It doesn't look like we actually use the githubApi methods anywhere yet, so I'd prefer to not make too many changes to that file.

### DIFF
--- a/assets/app/util/githubApi.js
+++ b/assets/app/util/githubApi.js
@@ -1,4 +1,5 @@
 import http from 'axios';
+import fetch from './fetch';
 
 import store from '../store';
 import { encodeB64, decodeB64 } from './encoding';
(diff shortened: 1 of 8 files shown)


assets/app/util/githubApi.js
```diff
@@ -1,25 +1,31 @@
...
 
 import store from '../store';
 import { encodeB64, decodeB64 } from './encoding';
 import errorActions from '../actions/errorActions';
 
 const API = 'https://api.github.com';
 
+// Thinking this function should not live here
+// Service should be agnostic as to what the store is.
+// Access token should be passed in to the functions
```
jeremiak: I'm fine with passing in the token but since it will likely be the same user and same token it might be nice to only need to use it once. Maybe something like a factory function where it returns an object that has access to an internal token?

Either way works for me in the end though.

el-mapache: Thats what I was thinking too, just store the token on instantiation. This only happens once per client session, provided the object gets initialized after the user logs in, and cleaned up when they log out.

jeremiak:
I think this looks great. I love using a polyfill for the `.fetch()` standard more than a different library.

jeremiak:
Hey @el-mapache, should we adjust the `.travis.yml` for this refactor/rewrite branch so that we don't fail on our old tests for now?

el-mapache:
Yes, that is an excellent idea! Side question related to testing, do you have a preferred library for mocking out http requests? It would be great to get started on testing these api service objects.

jeremiak:
@el-mapache I don't have a preferred mocking library but I have used `sinon` with some success and have been totally loving [`smocks`](https://www.npmjs.com/package/smocks) on [18f/cg-deck](https://github.com/18F/cg-deck/tree/jk-mock-test-api/test)

jeremiak:
Looks good to me!

PR #382, opened 2016-06-20
[WIP] Refactor using React and Redux to simplify code

jeremiak:
**This is a work in progress**

The Backbone app was hard to reason about and was delaying new engineers from contributing code. React is much easier to pick up because of the unidirectional data flow, so we are doing a refactor/re-write of the front end application using `react` and `redux`.

Checklist of views and features to migrate:
- [x] Unauthenticated homepage
- [x] Authenticated homepage with a site listing
- [x] Add a new site to Federalist https://github.com/18F/federalist/issues/444

Each site should have
- [ ] A simple media gallery to view all of their assets https://github.com/18F/federalist/issues/445
- [x] A view to view and change a site's settings https://github.com/18F/federalist/issues/446
- [x] A view with recent log activity
- [x] A tree view to walk the file structure if there is no `_navigation.json` https://github.com/18F/federalist/issues/447
- [x] A simplified page listing view that utilizes a `_navigation.json` file https://github.com/18F/federalist/issues/448
- [ ] An editor view to edit documents (issues listed below)

The editor view has the following components
- [ ] A CodeMirror based editor to edit the front matter of a markdown document. If there are some fields configured to be displayed as regular UI form elements, those should be shown to the left with the markdown text field to the right. https://github.com/18F/federalist/issues/450
- [x] A ProseMirror based text editor to edit the content of a document https://github.com/18F/federalist/issues/451
- [ ] An image button in the prosemirror menu bar that toggles a media uploading/gallery component. The user should be able to put assets in their document using this component https://github.com/18F/federalist/issues/452
- [ ] The ability to cancel the changes, save as a draft, or publish the content https://github.com/18F/federalist/issues/453
- [x] The ability to add a new page, with a custom path and file name


assets/app/actions/siteActions.js
```diff
@@ -0,0 +1,71 @@
...
+        siteId: site.id,
+        configs
+      });
+
+      return Promise.resolve(site);
+    });
+  },
+
+  fetchSiteConfigsAndAssets(site) {
+    return this.fetchSiteConfigs(site).then((site) => {
```
el-mapache: Do these need to be discreet states? From the code, it looks like we don't really use the intermediate state of `SITE_CONFIGS_RECEIVED` and just care about when the assets are loaded

assets/app/components/app.js
```diff
@@ -0,0 +1,53 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let state = this.props.state;
+    let navigation = this.props.state.navigation;
+    let content;
+
+    switch (navigation.name) {
```
el-mapache: Wondering if you have used react-router before, and if so, what your thoughts are on it? I find its a really nice way to package up the routing abstractions and provides a clear view of the structure of your application.

assets/app/components/nav.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+class Nav extends React.Component {
```
el-mapache: Nitpciky, but I think we want to delineate this nav form the side nav.

assets/app/main.js
```diff
@@ -0,0 +1,38 @@
...
+import store from './store';
+
+const mainEl = document.querySelector('#js-app');
+
+store.subscribe(() => {
+  const state = store.getState();
+  const isLoggedIn = (state.user) ? true : false;
+  if (!isLoggedIn) return;
+
+  ReactDOM.render(
```
el-mapache: Does this mean we are re-rendering the entire app when a state change occurs?

el-mapache:
In general, this is great. I think this is going to be a pretty big improvement over the existing organization of the codebase.

The two pieces that jumped out at me were the router and the re-rendering of the app when the store emits changes. I think that moving to a third party library for routing will remove a lot of the router specific code in out components and let us listen to store events on a more granular level.


I guess the route structure would look something like this (although I'm not 100% clear on the hierarchical relationship of the app). I see that the splash page is separate from the rest of the app, not sure if that is something we should integrate into the rest of the structure?

Example Route struture
```
<Router history={browserHistory}>
  <Route path="/" component={App}>
    <Route path="sites" component={Dashboard}/>
      <Route path="/:siteId" component={Site}>
        <IndexRedirect to="/pages" />
        <Route path="/pages" component={Pages} />
        <Route path="/media" component={Media} />
        <Route path="/settings" component={Settings} />
        <Route path="/logs" component={Logs} />
      </Route>
    </Route>
    <Route path="*" component={Dashboard}/>
  </Route>
</Router>
```

assets/app/main.js
jeremiak: Yes. Well, we're having each component call `.render()` and the cumulative result is diffed against the current DOM from the mount node.

My understanding was that this was an OK method because the DOM won't actually change if there are no updates to make.

Is there a better way to do this?

assets/app/actions/siteActions.js
jeremiak: Well, right now the sites reducer acts ([reducers.js#L55-L59](https://github.com/18F/federalist/blob/react-redux/assets/app/reducers.js#L55-L59)) on `SITE_CONFIGS_RECEIVED` to update the respective site with the configuration files. We could have an event that is for both configs and assets, but I think its fine to keep them discreet states for now.

el-mapache: I guess my point is, we never use this particular action. But, if we _will_ be using it in the future, it should definitely be discreet

wslack:
Added component issues

wslack:
Can the `react-redux` branch be deleted?

el-mapache:
Github won't let you close a branch that has an active PR against it. I left some comments on this PR https://github.com/18F/federalist/pull/472. If we merge the PR without making changes, I don't think the code won't work in the way that you expect.

We can either update that PR, and merge it into staging, or just close both of them.

wslack:
@el-mapache closed that PR; not worth making it work right now

PR #383, opened 2016-06-21
Adds page list component

el-mapache:
* Adds Pages and PagesListItem components
* Fixes bug in fetch wrapper
* Adds repo contents to fetch chain in siteActions

### DIFF
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -36,7 +36,8 @@ export default {
   fetchSiteAssets(site) {
     const config = site['_config.yml'];
     const assetPath = (config && config.assetPath) || 'assets';
-    github.fetchRepositoryContent(site, assetPath).then((assets) => {
+
+    return github.fetchRepositoryContent(site, assetPath).then((assets) => {
       return assets.filter((asset) => {
         return asset.type === 'file';
       });
@@ -65,7 +66,15 @@ export default {
 
   fetchSiteConfigsAndAssets(site) {
     return this.fetchSiteConfigs(site).then((site) => {
-      return this.fetchSiteAssets(site);
+      return this.fetchSiteAssets(site).then((site) => {
+        return github.fetchRepositoryContent(site).then((files) => {
+          store.dispatch({
+            type: siteActionTypes.SITE_CONTENTS_RECEIVED,
+            siteId: site.id,
+            files
+          });
+        });
+      });
     })
   }
 }
--- a/assets/app/components/linkButton.js
+++ b/assets/app/components/linkButton.js
@@ -1,0 +1,16 @@
+import React from 'react';
+
+const propTypes = {
+  className: React.PropTypes.string,
+  text: React.PropTypes.string,
+  href: React.PropTypes.string
+}
+
+const LinkButton = ({className, text, href}) =>
+  <a href={href} className={className}>
+    {text}
+  </a>;
+
+LinkButton.propTypes = propTypes;
+
+export default LinkButton;
--- a/assets/app/components/site/pageListItem.js
+++ b/assets/app/components/site/pageListItem.js
@@ -1,27 +1,21 @@
 import React from 'react';
 
 const propTypes = {
-  page: React.PropTypes.object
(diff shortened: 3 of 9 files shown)


assets/app/actions/siteActions.js
```diff
@@ -65,7 +66,15 @@ export default {
 
   fetchSiteConfigsAndAssets(site) {
     return this.fetchSiteConfigs(site).then((site) => {
-      return this.fetchSiteAssets(site);
+      return this.fetchSiteAssets(site).then((site) => {
```
el-mapache: Currently we make 4 (I think?) requests to GitHub to get various pieces of content for the site. One request should be able to get the `_config.yml`, `_navigation.json`, and general content of the site correct? I'm not sure when the user need access to the `assets` folder, but my feeling is that its necessary when they enter the editor or click the 'Media' link. Is that accurate?

jeremiak: > but my feeling is that its necessary when they enter the editor or click the 'Media' link

That is accurate.

> One request should be able to get the _config.yml, _navigation.json, and general content of the site correct?

I don't think so. From what I can tell ([Github API docs](https://developer.github.com/v3/repos/contents/#response-if-content-is-a-directory)), the response from an endpoint like `https://api.github.com/repos/:user/:repo/contents` returns an array of the root contents, but a request is still needed to get the actual content of that file.

el-mapache: Are both config.yml and navigation necessary to view the tree?

el-mapache: I'm asking because this is a pretty slow part of the app, and we have to render 4 separate times before the user sees anything on the page.

el-mapache: I'm wondering then if we shouldn't try to enforce a navigation.json to be in every repo, since the extra request is basically only required to develop locally.

jeremiak: I think that just the navigation file might be necessary to view the tree. And not even the file, just if it is there or not.

Ya, I'm not totally opposed to the idea of forcing repos to have a `navigation.json`-like thing (and our templates should definitely have them) but I also like the idea that any Jekyll site be added to Federalist and they can configure it themselves to take advantage of features. And I agree that its out of scope of this PR :)

assets/app/components/site/pageListItem.js
```diff
@@ -0,0 +1,34 @@
...
+    const { page } = this.props;
+
+    return (
+      <li className="list-group-item">
+        <div className="usa-grid">
+          <div className="usa-width-two-thirds">
+            {page.name}
+          </div>
+          <div className="usa-width-one-third">
+            <a className="usa-button usa-button-outline file-list-item-button" href={page.url}>
```
jeremiak: I think we'll want to change `href` here. In the current application, we basically have a route like `site/:id/edit/:branch(/)*file` where `file` is the value of `path` in the Github API response.

In order to do that though, we'll probably need to pass more of the state all the way down, as we'll need the site id and branch. Is this something that would be easier to do once we have react-router in place?

el-mapache: We'll have the `:siteId` but not the `:branch`. The latter will have to be manually passed down. We'll only have the branch slug once the user navigates

assets/app/components/linkButton.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const propTypes = {
+  className: React.PropTypes.string,
+  text: React.PropTypes.string,
+  href: React.PropTypes.string
+}
+
+const LinkButton = ({className, text, href}) =>
```
jeremiak: Woah, you can make components with just function like this?

What does `React` do in this case?

el-mapache: React will magically call createClass under the hood! It also infers that
this is a component with just a render method and no additional logic.

jeremiak:
Nice work! I don't think this is totally done since none of the edit links work (if the link is to another directory it should use this same view and if it is to a file it should show the content editor view) but I'm going to merge it since its a great starting point and we don't want long lived branches anymore :)

PR #384, opened 2016-06-22
Adds super basic react router

el-mapache:
* Adds react router
* Organizes components a bit more
* start using pure render components where applicable


jeremiak:
This looks good, going to merge this now so that we can build off the react router.

PR #434, opened 2016-06-26
Add Org whitelist instructions and sample code

wslack:
Helps user deploy Federalist locally

PR #328, opened 2016-03-07
Add organization whitelist instructions to readme

timwis:
Had to dig into the code a bit to figure out why I couldn't login locally. Thought this might help.

PR #437, opened 2016-06-28
Add Org whitelist instructions and sample code

jeremiak:
This merges a long running pull request #328 and attributes the changes to @timwis.

This should also close #434 - thanks for opening that PR will but I wanted to make sure to give credit where it is due.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,17 @@ module.exports = {
 };
 ```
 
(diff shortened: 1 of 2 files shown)


timwis:
Thanks @jeremiak, you are truly a git wizard!

PR #439, opened 2016-06-28
Stores state in more place to deal with react-router weirdness

el-mapache:
* Better regex to exclude these branches from travis CI
* Saves the store's current state in the provider and top level app component so it propagates correctly through react-router
* Adds method to fetch content from arbitrary folders in github repos to siteActions
* Rearranges some folders for organization


assets/app/actions/siteActions.js
```diff
@@ -75,6 +76,23 @@ export default {
           });
         });
       });
-    })
+    });
+  },
+
+  fetchContent(site, path) {
+    function dispatchChildren(site, path, files) {
```
el-mapache: rename to dispatchChildContent

assets/app/components/site/SideNav/sideNav.js
```diff
@@ -16,28 +16,30 @@ class SideNav extends React.Component {
...
-          icon='settings'
-          linkText={sideNavPaths.SETTINGS}
-        />
-        <SideNavItem
-          href={this.getUrl(siteId, sideNavPaths.LOGS)}
-          icon='logs'
-          linkText={sideNavPaths.LOGS}
-        />
-      </ul>
+      <div className="usa-width-one-sixth" id="fool">
```
jeremiak: Is this `id` used?

assets/app/main.js
```diff
@@ -8,48 +8,26 @@ import siteActions from './actions/siteActions';
...
-}
-
-Provider.childContextTypes = {
-  state: React.PropTypes.object
-};
 
 store.subscribe(() => {
   const state = store.getState();
+
   if (!!!state.user) return;
```
jeremiak: Would just a single `!` work here?

assets/app/main.js
```diff
@@ -8,48 +8,26 @@ import siteActions from './actions/siteActions';
...
 
   render((
-    <Provider state={state}>
+    <Provider state={{get: store.getState}}>
```
jeremiak: Why not just use `state={ store.getState() }` here?

assets/app/actions/siteActions.js
jeremiak: I could probably go either way, but I like using fetch instead of dispatch here. I realize that the function is mostly dispatching an action but the action creator still calls the fetch method to make a network request and get the content, and that's why you'd want to call it in the first place.

assets/app/routes.js
```diff
@@ -15,18 +15,20 @@ const Stub = () =>
...
 export default (
   <Route path="/" component={App}>
-    <IndexRedirect to="sites" />
+    <IndexRedirect to="sites"/>
     <Route path="sites">
       <IndexRoute component={Dashboard}/>
       <Route path=":id" component={SiteContainer}>
-        <IndexRoute component={PagesContainer}/>
+        <IndexRoute component={SitePagesContainer}/>
+        <Route path="tree/:fileName" component={SitePagesContainer}/>
```
jeremiak: I love this distinction between the tree and actual edit routes, much clearer than before!

jeremiak:
Yo @el-mapache this looks pretty cool. I'm not sure I understand a few things.

1. Where does `context` come from when a component is created? Some of the constructors accept `props` and `context`, but I'm not sure why.

Questions spurned by this PR but not directly related:

1. What sort of naming scheme should we follow when naming/organizing the components?

assets/app/components/site/SideNav/sideNav.js
el-mapache: haha I have no idea, I haven't changed any of the original ids. It was helpful when I was searching the code...

assets/app/main.js
el-mapache: yes, yes it would

assets/app/main.js
el-mapache: App.js needs to have access to the store.getState function so it can call it on componentWillReceiveProps and update its own internal state. We can also pass the output of store.getState() to app.js and just set state with the entire context

```
componentWillReceiveProps(nextProps, nextContext) {
  this.setState(nextContext);
}
```

el-mapache:
`context` is a part of the React API that is generally used to store some global information components might care about. You can pass information down through a subtree of components without explicitly passing props.

Any component that cares about `context` just has to define a `contextTypes` property on the class, with the properties that component needs. `context` isn't great for mutable values as updating it doesn't trigger a call to `render`, which is why we have to store the output of the `context` in the `state` object of app.js.

Unfortunately, we _have_ to use it with react-router because there is no way to pass down props through the router.

https://facebook.github.io/react/docs/context.html

jeremiak:
Cool, I'm merging this so we can build off of it.

There's some weirdness here that we'll clean up in a bit:

0. No editor components so the route just says "hi"
0. Child directories aren't listing their children

Issue #171, opened 2015-09-16
Support renaming repositories


dhcole:
A user may at some point want to rename a repository. Doing so breaks Federalist, because webhook posts include the new repository name.

Changing the repository name in Federalist would publish the site to a new S3 prefix, which would break mapping to custom domains.

We should try to figure out a better way to support this.

For now, if a user wanted to rename a repository, these would be the steps:
1. Rename the repository on GitHub
2. Add the new repository as a new site in Federalist
3. Copy over any configuration and custom domain settings from the old site's settings
4. Update the CDN origin to map to the new site's S3 origin path
5. Once confirmed that the CDN origin change has propagated, delete the original site in Federalist

This process should be stable and should not result in downtime as Federalist doesn't delete the site content from S3 when a site is deleted in Federalist, though I'd recommend running this on a test site first.

cc @shawnbot

dhcole:
Probably the best way around this issue is captured in #161, which is to adjust Federalist to publish sites to `/sites/custom-domain.example.gov/` on the S3 bucket. This way, if a repo changes, simply adding a new repo with the same custom domain would change which repo publishes to the site.

shawnbot:
:+1: to that, @dhcole

wslack:
This just came up in a thread today. Right now a repo name change requires DNS config (the price we pay for predictable URLs) and the /sites/custom-domain.example.gov/ path is probably best, though we'll have to account for possible conflicts if multiple repos are set up with the same custom URL path.

shawnbot:
For posterity, here's what we ended up doing with College Scorecard:

1. Disabled webhooks on the old repo
1. Transferred the repo to RTI
1. Re-added the new repo
1. Copied the custom configuration over from the old one to the new one
1. Deleted the old repo

wslack:
@shawnbot you mean the config in the federalist GUI or something else?

shawnbot:
Sorry, yes: I meant the custom config field in the Federalist GUI.

wslack:
related to https://github.com/18F/federalist-docs/issues/39

wslack:
@el-mapache interested in this one?

wslack:
@jseppi this could also be a next project for you.

wslack:
To do this, the user would have to be an admin on the repo.

jseppi:
We should think through the interactions and user experience for this pretty carefully. My gut feeling is that this is kind of weird.

wslack:
[...]

PR #454, opened 2016-07-08
React new site

el-mapache:
* Adds alert banner
* Adds new site templates selector
* Adds ability to create a site from a selected template


el-mapache:
How is federalist currently handling sites that are deleted? If the user deletes the repo does the federalist app get a webhook notifications?

jeremiak:
@el-mapache Federalist doesn't handle sites being deleted at all right now. We should implement that, but probably improve the way we associate sites with Github repos per #171

jeremiak:
@el-mapache I made some changes, the most significant of which is to redirect to the site listing page after the site is successfully added. I'm going to merge this because the PR is old at this point and we can make any functionality changes in a subsequent PR.

PR #458, opened 2016-07-12
Add unit tests for a single Redux reducer. They live in `test/client`.

### LINKED ISSUE #457: Add front-end unit tests and code coverage tooling
This includes writing a minimal unit test to prove the tooling works.

brandocalrissian:
Add code coverage reporting for the client tests and code.

`npm run test:client` runs the client mocha tests.

`npm run coverage:client` creates the test coverage reports for the `app/`
code. Point your browser at the generated `coverage/index.html`

Add chai for assertions.

Add istanbul for code coverage.

Add proxyquire so we can have true unit tests (while we're still using a
transpiler -- will need to use something else when everything is native
ES2015).

Add a .babelrc to control how babel is invoked when running the tests
and coverage reports. We can use the same mechanism in the regular build
process (later).

gitignore the directory where we put the code coverage reports.

Fixes #457.


brandocalrissian:
Looking into travis failure.

brandocalrissian:
Looking into travis failure. It happens on npm install.

brandocalrissian:
The prosemirror build (triggered when we `npm install`) itself runs a particular version of babel that does not seem happy with the format of the .babelrc file included in this PR. Will need to either change the .babelrc to not have the standard name and trigger its config when we want it or ditch using .babelrc entirely.

el-mapache:
Have you tried just including the babel presets in the package.json file itself?

```
"babel": {
  "presets": ["es2015", "react"]
}
```

I'm not 100% sure if that does exactly the same thing as having a babelrc file, but it tell us if the problem is competing versions of babel, or the actual rc file itself

brandocalrissian:
Looks like we're using an older version of prosemirror, a version depending upon the old-school babel 5 package as the CLI. The latest version of prosemirror is updated to rely on babel 6. When we update to a newer version of prosemirror, there's a decent probability the build problem will go away. Eight months can be an eternity in tooling...

Will work around the current dependency as changing the editor is way out of scope for this task.

brandocalrissian:
Current travis failure:
Error: Cannot find module './components/SiteContainer' from '/home/travis/build/18F/federalist/assets/app'

Why? The file is called "siteContainer.js". We won't be able to reproduce the error on non-case-sensitive filesystems (e.g. OS X). Travis is running in Linux and a case-sensitive file system.

brandocalrissian:
Yes, this is an issue in the react-redux branch. See the latests travis output:

Error: Cannot find module './MediaThumbnail'

Our imports do not match the files on case-sensitive filesystems and travis can't find the files.

brandocalrissian:
Added issue to fix the case problem in #459.

brandocalrissian:
Merged in the fed-459 branch and now travis is failing here:
```sh
> federalist@0.0.0 browserify /home/travis/build/18F/federalist
> browserify assets/app/main.js -o assets/js/bundle.js -t [ babelify --presets [ es2015 react ] ]
Error: Cannot find module './components/site/pages/pagesContainer' from '/home/travis/build/18F/federalist/assets/app'
    at /home/travis/build/18F/federalist/node_modules/browserify/node_modules/resolve/lib/async.js:55:21
```

brandocalrissian:
5251cfd was a push --force and rewrote history.

jeremiak:
Woot, this now passes!

I'm going to merge this because it does set up a testing structure and therefore accomplishes the original goal. Since we changed [18f/prosemirror](https://github.com/18f/prosemirror), a build won't turn green until we rebase this against open PRs.

PR #462, opened 2016-07-15
Calls bootstrap callback correctly

el-mapache:
Only runs unit tests against the API and new React client side code

* Removes references to browser and integration tests
* Removes SauceConnect addon from `.travis.yml`


jeremiak:
Looks good to me!

PR #463, opened 2016-07-17
Update prosemirror library and wrap in component

jeremiak:
After all the work last week to deal with prosemirror build problems, I wanted to see how easy it was to upgrade the library. We originally had to fork it because we were using it before it was published to the npm registry.

It was easy to install the library and I wrapped it in a component for ease of use within React. I did the same for the codemirror library and made an editor container object to hold both of them.

I'm not sure I named the files correctly, since there seems to be kind of a mix of capitalization usages when it comes to components and directories. Maybe we should just switch everything to snake case while we still have a relatively small number of components so that its easier to think about? After that issue where travis is case sensitive, I'm sort of leaning towards snake case so we avoid it entirely.

Otherwise, I'm open to camel casing everything starting with lowercase letters but it should be consistent at least across the front end code.

Thoughts @el-mapache & @brandocalrissian?

### DIFF
--- a/assets/app/components/site/editor/prosemirror.js
+++ b/assets/app/components/site/editor/prosemirror.js
@@ -11,14 +11,16 @@ const menu = buildMenuItems(schema);
 class Prosemirror extends React.Component {
   constructor(props) {
     super(props);
-
+    this.state = {
+      doc: null
+    };
     this.toMarkdown = this.toMarkdown.bind(this);
   }
 
   componentDidMount() {
     this.editor = new prosemirror.ProseMirror({
       doc: defaultMarkdownParser.parse(this.props.initialMarkdownContent),
-      place: document.querySelector('#js-prosemirror-target'),
+      place: document.getElementById('js-prosemirror-target'),
       schema: schema,
       plugins: [
         menuBar.config({ content: menu.fullMenu })
@@ -28,6 +30,13 @@ class Prosemirror extends React.Component {
     this.editor.on.change.add(() => {
       const doc = this.editor.doc.toJSON();
       this.setState({ doc });
+    });
+  }
+
+  componentWillUnmount() {
+    const changeHandlers = this.editor.on.change.handlers;
+    changeHandlers.forEach((handler) => {
+      this.editor.on.change.remove(handler.fn);
     });
   }
 
--- a/assets/app/components/site/editor/prosemirror.js
+++ b/assets/app/components/site/editor/prosemirror.js
@@ -1,0 +1,55 @@
+
+import React from 'react';
+import prosemirror from 'prosemirror';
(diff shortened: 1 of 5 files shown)


assets/app/components/site/editor/prosemirror.js
```diff
@@ -0,0 +1,55 @@
...
+  constructor(props) {
+    super(props);
+
+    this.toMarkdown = this.toMarkdown.bind(this);
+  }
+
+  componentDidMount() {
+    this.editor = new prosemirror.ProseMirror({
+      doc: defaultMarkdownParser.parse(this.props.initialMarkdownContent),
+      place: document.querySelector('#js-prosemirror-target'),
```
el-mapache: Not a huge deal since we are only calling this one, but in general it is always better to use getElementById when you can

assets/app/components/site/editor/prosemirror.js
```diff
@@ -0,0 +1,55 @@
...
+      schema: schema,
+      plugins: [
+        menuBar.config({ content: menu.fullMenu })
+      ]
+    });
+
+    this.editor.on.change.add(() => {
```
el-mapache: I think we need to unbind from this even on unmount

assets/app/components/site/editor/prosemirror.js
```diff
@@ -0,0 +1,55 @@
...
+import { defaultMarkdownParser, defaultMarkdownSerializer } from 'prosemirror/dist/markdown';
+import { menuBar } from 'prosemirror/dist/menu';
+import { schema } from 'prosemirror/dist/schema-basic';
+
+const menu = buildMenuItems(schema);
+
+class Prosemirror extends React.Component {
+  constructor(props) {
+    super(props);
+
```
el-mapache: I guess yu don't technically have to, but we should toss a
```
this.state = {
  doc: null ( or '')
};
```
in the constructor

el-mapache:
Looks great. I feel pretty strongly against using snake case in javascript projects, but if everyone else wants that I will defer to the will of group

brandocalrissian:
I have no preference on casing other than I prefer consistency within a project whenever we can pull it off. I spent enough years in JavaLand to MakeThis or thisReallyEasyToRead. On the other hand, this_is_easy_to_read_comma_too.

el-mapache:
Great, lets just do snake_case in the interest of expediency. I will make a new ticket for that work.

jeremiak:
Good feedback @el-mapache, updated the code.

I'm cool with whatever case. If you feel strongly one way or the other Adam, let's go with your style and just not capitalize the first letter of any directory or js file?

el-mapache:
Looks good! Will merge once the build passes.

Issue #466, opened 2016-07-19
Add unit test coverage to all of reducers.js


brandocalrissian:
Now that we have the unit test plumbing in the react-redux branch, let's finish the unit test coverage for the reducers.

PR #467, opened 2016-07-20
Get to 100% test coverage for redux reducers

brandocalrissian:
Found a couple of things that might be issues while writing the tests (see commit log for 81d730d6dd55e0ce9ff679d975c86e079e61880b).

This PR is supposed to solve #466.


el-mapache:
@brandocalrissian I added the childDirectoriesMap object to the reducer, what is the issue you are running into? We are currently using it to build out the tree view in the `pages` view

brandocalrissian:
@el-mapache the line that confuses me is 

```js
      const currentMap = state.childDirectoriesMap || {};
```

I don't see how the state can ever have a childDirectoriesMap. The sites in the state, yes! The root sites reducer state? Um... I looked around and couldn't see how it's set.

assets/app/reducers.js
```diff
@@ -39,36 +39,21 @@ export function builds(state = [], action) {
...
-        if (s.id !== action.siteId) return s;
-        return Object.assign({}, s, {
-          files: action.files
-        });
-      });
+      return mapPropertyToMatchingSite(state, action.siteId, { files: action.files });
+    case siteActionTypes.SITE_DELETED:
+      return state.filter((site) => site.id != action.siteId);
     case siteActionTypes.SITE_CHILD_CONTENT_RECEIVED:
       const currentMap = state.childDirectoriesMap || {};
```
brandocalrissian: This is the puzzling line where I don't see how the "currentMap" can be anything other than {}.

el-mapache: ahh I see. Yes, that is not correct, I think its a bad copy/paste from some previous code, when I put the child contents in a separate reducer (for some reason).

Most of that work should get moved into the function that is applied to the list of sites currently in the state array.

brandocalrissian: Was really meant to be this:

```js
    case siteActionTypes.SITE_CHILD_CONTENT_RECEIVED:
      return state.map((site) => {
        if (site.id !== action.siteId) return site;

        const currentMap = site.childDirectoriesMap || {};
        const nextMap = {};
        nextMap[action.path] = action.files || [];

        return Object.assign({}, site, {
          childDirectoriesMap: Object.assign({}, currentMap, nextMap)
        });
      });
```
?

el-mapache: Yes, exactly!

brandocalrissian:
Done. We now have the desired behavior in the code for this PR.

el-mapache:
Perfect, thanks!

What are your thoughts on breaking up the reducers / tests into separate
files? Has the code reached a place where that might be advisable?

brandocalrissian:
I usually prefer a collection of files instead of one file with lots of independent things in it. I can split them up quickly now.

I'd also advise putting constants in the file that owns it instead of creating separate files filled with all the constants.

el-mapache:
I believe the main reason a lot of constants are put into a single file is
because other files can use the same constants. Reducers and and the action
creators (or utility files, or whatever) both need access to the them, for
example.

brandocalrissian:
split the reducers into separate files.

any other suggestions?

el-mapache:
No, this looks awesome! Thanks for getting started on the tests, spinning up a test harness is really important and also not super fun to do, so this work is really appreciated.

I think this can probably be merged once the build passes.

PR #469, opened 2016-07-28
Watch npm task

jeremiak:
The `npm run watch` task wasn't rebuilding client side code on changes for some reason. As we're doing work on the react refactor here, I've changed the script to only watch for client side code before restarting the build process


el-mapache:
Looks good! Will merge once the build is done

Issue #450, opened 2016-07-08
Refactor: Editor view needs a CodeMirror based editor to edit the front matter of a markdown document.


wslack:
If there are some fields configured to be displayed as regular UI form elements, those should be shown to the left with the markdown text field to the right.

PR #473, opened 2016-08-01
Save content from editor

jeremiak:
End feature will be a working editor interface that will commit markdown back to Github.

Refs #450 #451 #453

### DIFF
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -55,13 +55,16 @@ export default {
     }).catch(error => alertActions.httpError(error.message));
   },
 
-  createCommit(site, path, fileData) {
+
+  createCommit(site, path, fileData, message = false, sha = false) {
     const b64EncodedFileContents = encodeB64(fileData);
-    const commit = {
-      message: `Adds ${path} to project`,
+    const siteId = site.id;
+    let commit = {
+      message: (message) ? message : `Adds ${path} to project`,
       content: b64EncodedFileContents
     };
-    const siteId = site.id;
+
+    if (sha) commit = Object.assign({}, commit, { sha });
 
     github.createCommit(site, path, commit).then((commitObj) => {
       alertActions.alertSuccess('File added successfully');
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -124,7 +124,13 @@ export default {
           return files;
         });
       });
-    }).catch((error) => alertActions.httpError(error.message));
+    }).catch((error) => {
+      // TODO: make a generic catch handler that will only
+      // trigger an http error action for an actual http
+      // error.
+      throwRuntime(throwRuntime);
+      alertActions.httpError(error.message)
+    });
   },
 
   // todo rename to something like fetchTree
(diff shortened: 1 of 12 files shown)


jeremiak:
@el-mapache This is ready for review now

assets/app/actions/siteActions.js
```diff
@@ -55,13 +55,16 @@ export default {
...
 
-  createCommit(site, path, fileData) {
+
+  createCommit(site, path, fileData, message = false, sha = false) {
     const b64EncodedFileContents = encodeB64(fileData);
-    const commit = {
-      message: `Adds ${path} to project`,
+    const siteId = site.id;
+    let commit = {
+      message: (message) ? message : `Adds ${path} to project`,
```
el-mapache: Do we still need this fallback with the other default messages in place in the editor component?

el-mapache:
Does it matter right now that nested files can't be opened properly? It looks like requests are made to find all files at the root level of the repo.

assets/app/components/site/editor/editorContainer.js
```diff
@@ -85,37 +114,60 @@ class Editor extends React.Component {
...
-      return file.path === this.path;
-    });
-
-    if (currentFile && currentFile.content) {
-      return decodeB64(currentFile.content);
-    }
-    return '';
+  getComputedMessage() {
+    const newPage = this.props.route.isNewPage;
+    const hasMessage = this.state.message || this.state.message === '';
```
el-mapache: This can be ``!!this.state.message`

brandocalrissian:
The views are getting pretty beefy and lengthy with the management of internal state and whatnot, but it's consistent with how we're using react thus far.

I noticed a drop in code coverage in the reducer and am using the tests to clarify my understanding of the changes there-- will have a commit soonish.

el-mapache:
I think most of the views that manage internal state are ones that contain forms (or, rather, are the parent containers of forms), which is generally accepted as OK in terms of React best practices.

There is also an argument to be made to for storing form states in their own reducers, and having any setState calls in the containers update those reducers and have all form data flow back down into the form components as props. This results in a bit more boilerplate but buys into the redux pattern a bit more

brandocalrissian:
I back-filled the code coverage cases in the commits. I wonder if the extra cases are things we should check on the way in-- there are a few cases where we expect something to not be initialized until we try to update it (say, site files) instead of making sure the entity is initialized with empty, but well-formed, data (say, []). That would make the reducers simpler with fewer edge cases.

I'm not knowledgeable enough about the expected functionality to know if our intention is to run with partially initialized and progressively infilling data structures or if we're instead protectively coding around potentially illegal states.

If we happen to be running with the second option, can we prevent illegal states by rejecting insufficient data on the way in?

brandocalrissian:
As for stateful views, I'm for the simplicity of state always being state and views reflecting some subset of the state instead of conflating the two, but that's a different conversation out of the scope of this PR...

brandocalrissian:
@el-mapache and @jeremiak, if you're ok with the test backfill I threw in there, I'm 👍 to blessing/merging this PR.

el-mapache:
@brandocalrissian I don't know what of the two modes we are aiming to operate in, but I do know that part of the reason I was backfilling state is that I wasn't really sure at any given time, what the complete state of the app was supposed to be. There weren't really any model definitions in the backbone app, and enough things were mutating that I wasn't super comfortable making a initial schema based off of the network requests.

I think in a system with unknowns its a little easier to add in protective clauses as they come up. In general, my preference would be to always have a well-defined schema that provides defaults. This way we don't end up duplicating guard logic throughout the views.

brandocalrissian:
Not a problem. We can put figuring out valid states, protecting the system from encoding invalid states, and removing guard clauses onto our tech debt list. We don't have to do them all at once.

PR #475, opened 2016-08-05
references path so edit view loads correct file

el-mapache:
* References `path` property instead of `name` when generating an edit view link, github file path now resolves correctly


jeremiak:
Closed in favor of #476

PR #476, opened 2016-08-08
Jk edit tree fix

el-mapache:
* Nested file paths resolve correctly
* Refactored SITE_CHILD_CONTENTS_RECEIVED and the directory map to SITE_FILES_RECEIVED and flat array
* Fixed tests
* First attempt at standard data fetching pattern -- calls actions from both componentDidMount and ComponentWillReceiveProps to handle cases where component is loaded directly and navigated to.


jeremiak:
Looks good to me

PR #477, opened 2016-08-09
Moves methods to component, new names, fixes loading bug

el-mapache:
* Moves getPath and filesFromPath functions into pagesContainer component
* Renames functions for clarity
* Fixes bug where nested pages don't load properly


jeremiak:
Nice refactoring

PR #479, opened 2016-08-11
Pm image component

el-mapache:
* Adds image picker component
* Adds prosemirror image menu extension
* Adds prosemirror markdown parser/serializer extensions
* Adds ability to insert images into markdown documents


assets/app/actions/siteActions.js
```diff
@@ -55,6 +55,23 @@ export default {
     }).catch(error => alertActions.httpError(error.message));
   },
 
+  uploadFile(site, path, fileData, message) {
```
el-mapache: This action creator is very similar to the createCommit action below it. I'd like to be able to unify them, but I'm not sure of the best way to do that.

They diverge in two major ways
* the file data is already base64 encoded when uploading an asset
* we have different callback behavior in each of the two actions &emdash; the uploadFile action doesn't redirect after a successful upload, but the createCommit action does

jeremiak: Why is the file data already base64 encoded? Is there a way that encoding or decoding only happens in actions and components just work with things like human readable strings and file objects?

jeremiak: Also, I don't think the `createCommit` action should redirect. I think it should return a promise and then the consumer of `createCommit` can decide if it should redirect or not. I think we just did this in our pairing session...

jeremiak:
Hey @el-mapache I am not able to get this running on my machine. I can load the site, browser the site listing, browse the file listing but it errors out when I try to edit a file (in this case, [`README.md`](https://github.com/jeremiak/federalist-modern-team-template/blob/master/README.md)

Here's the start of the stack trace:
```js
Uncaught TypeError: content.match is not a function
initializeMD @ bundle.js:4695
documentStrategy @ bundle.js:4722
getStateWithProps @ bundle.js:2428
Editor @ bundle.js:2370
...
```

el-mapache:
yeah I think I needed to include a check for the existence of the file. Its in one of the other PRs I opened. This is why tests are important!

assets/app/actions/siteActions.js
el-mapache: I included the file decoding in the actual component. I left a comment in there about moving it to an action creator, wanted to get some feedback on if that seemed like a good idea. I believe this confirms that it would be!

el-mapache: Moved file parsing to the action creator. We still can't unify both of the functions for now, as the file is going to be run through a FileReader object, which encodes the contents to bas64 for us.

el-mapache: We can move the redirect action out of the createCommit in a future PR. This one is already pretty large and that functionality had been in the action for a few weeks prior to this code being added

el-mapache:
Added a test for the case of an invalid file being supplied to the documentStrategy function. Also, removed two tests that explicitly tested the private methods in that file. Showing that, given each type of file object input, the correct value is returned from the `documentStrategy` function is sufficient.

Thanks for taking the initiative and putting tests in there for that file!

jeremiak:
This looks cool! I don't see the bug I was reporting earlier but am now seeing the follow bugs:

1. I can upload a new image to Federalist and it shows up in the image gallery above the editor (🎉!) however when I click on the new image it triggers a browser native file picker UI where I can select a file from my local file system to upload.
2. The image URL is committed to Github ([ae1676b46105c1d102a4d7e180f2a739ee17e958](jeremiak/federalist-modern-team-template/commit/ae1676b46105c1d102a4d7e180f2a739ee17e958) still contains the path for getting the file off of Github. However, when this markdown is build into a page with Jekyll, the image path will actually reside at `{{ site.baseurl }}/assets/logo.png`

Let me know if that's confusing/weirdly worded

el-mapache:
1). It seems like I didnt check that use case, although I'm not sure why its any different from any of the other images, weird!
2). That is an oversight for sure, i wanted it to have the jekyll variables in there.

el-mapache:
@jeremiak what is the url supposed to look like? It is supposed to include both the github url and the variable? or just the variable?

jeremiak:
@el-mapache woot, the path looks good: https://github.com/jeremiak/federalist-modern-team-template/commit/62af73adf8a19047aa24133e37fbe98a71c8fc38

I'm going to merge this, but there is a weird small bug where the commit message isn't reflective of what the form field says. You'll notice that the commit message in my commit above is "Adds README.md to project" when it was an existing file and the commit message field said "Changes made to README.md"

jeremiak:
I am also seeing that when I upload an image, it isn't displayed in the image picker but I do get a success message that says "File was uploaded successfully" and it [commits the image to the repo](https://github.com/jeremiak/federalist-modern-team-template/commit/9f164cac1758eaa81b53b0d7bc8d6aef855f0c7b)

el-mapache:
@jeremiak I'm not seeing the image issue, images uploaded for me appear in the picker. I wonder if there is a race condition between the images getting uploaded to github and the image being available via the API

I can look at the message issue, haven't changed any of the logic around that in weeks though so its probably been broken for the duration

jeremiak:
@el-mapache Let me triple check I'm on the right branch and have a clear user agent cache before you go off bug hunting then

jeremiak:
This looks great, merging

Issue #448, opened 2016-07-08
Refactor: Each site should have a simplified page listing view that utilizes a `_navigation.json` file


wslack:

brandocalrissian:
How do we want to treat these nested views? How are they done in the non-react version?

![image](https://cloud.githubusercontent.com/assets/1403376/17486111/9abe25a2-5d5d-11e6-8ef9-5e61e6862262.png)

brandocalrissian:
Work is happening over here: https://github.com/18F/federalist/tree/bk-448

PR #481, opened 2016-08-12
Each site should have a simplified page listing view that utilizes a `_navigation.json` file

brandocalrissian:
This is the proposed fix for #448.

It handles having a _navigation.json file for listing simplification and splits out the two component paths (with/without _navigation.json).

It will keep nesting `ul`s as the _navigation.json specifies deeper and deeper child hierarchies.

There are some ugly CSS display issues, but it seems to be about the same as what we have on production. Propose we fix the display issues with another issue/PR.

Extracted some of the logic out of pagesContainer into a new pageList-- it renders the page list when the navigationJsonPageList doesn't apply.

Added some script npm targets for building the sourcemaps for easier client-side debugging.


assets/app/components/site/Pages/navigationJsonPageList.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PageListItem from './pageListItem';
+
+const navigationJsonContent = ({ site }) => {
```
el-mapache: This is probably nitpicky, and maybe only an issue that I have, but I really like the render function to be the last one in component.

It makes it more explicit that this component is essentially a function that returns some DOM.

Do other people have thoughts on this?

assets/app/components/site/Pages/pagesContainer.js
```diff
@@ -14,85 +16,65 @@ class Pages extends React.Component {
...
 
+const fetchFiles = (site, params) => {
+  siteActions.fetchFiles(site, getPath(params));
+};
 
-  getFilesByPath(files = [], startingPath = '') {
-    const isRoot = (startingPath === '');
-    const path = (!isRoot) ? `${startingPath}/` : '((?!/).)*$';
-    const startsWithPath = new RegExp(`^${path}`);
+const getPath = (routeParams) => {
```
el-mapache: Why move these back out again?

assets/app/components/site/Pages/navigationJsonPageList.js
brandocalrissian: I'm happy following any agreed convention. I tend to try to structure code where it starts at the highest level and references functions declared below as I find it more legible. My internal monologue goes, "I get what this does, now I need to understand what the function it's calling does because its name is dreadfully confusing... <scrolls down a bit> Ah, here it is."

assets/app/components/site/Pages/pagesContainer.js
brandocalrissian: Do we ever want anyone outside of the pagesContainer object to call pagesContainer.getFilesByPath() or pagesContainer.getPath()? If we prevent people from calling such functions, then nobody will ever accidentally call these functions. The only way we have real "private" members in our javascript class is by hiding them from the outside.

I'm also actively trying to put our components on a diet with a goal of having them responsible only for rendering a particular state represented by properties. This is intended to start extracting such functionality. For instance, why does the view component responsible for extracting "splat" and "fileName" from a map and determining what to give siteActions.fetchFiles()? Would it be better to push that logic down to the siteActions since pagesContainer is the only caller of siteActions.fetchFiles(), especially when we're going to cut more getPath() calls in pagesContainer.componentWillReceiveProps()? Moving out methods like these puts more of a "what is that doing here?" spotlight on the code.

el-mapache: I think in the case of React we don't need to worry about someone calling a method directly on a container, that is a huge anti-pattern and I don't think there would be a reason for a user to do that.

I don't hate the idea of moving the checks into actions; if we did that, though, would we start moving all that sort of business logic into the actions? For example, in my file picker component, would I want to move file size verification logic into an action?

el-mapache:
I guess my thinking is, if you are going to move these methods outside of the component, lets just go ahead and move them into either:

* an action creator
* a higher order component / controller function

brandocalrissian:
I wouldn't worry about designing for reuse that hasn't happened yet. Once we've determined something needs to be reused, then we can move the functionality to fit where it is needed: evolve the design based on actual, not predicted, needs.

el-mapache:
Oh ok, then I am little bit confused. If we aren't prepping the code to be reused, why move it out the component at all? Beyond someone accidentally calling the method on an instance of the component, which I don't think is particularly likely.

el-mapache:
The code looks fine, so feel free to merge!

assets/app/components/site/Pages/pagesContainer.js
brandocalrissian: Where is the logical line that makes the most sense today? What responsibility do we want the actions to have? In this case, I see `siteActions.fetchFiles(site, path)` and it makes sense to me, much more sense than `fetchFiles(site, hopeYouGetWhatYouWantInThisHash)`.

I guess what I'm getting at is where should this logic live? Why are we taking a generic `params` object and transform it into something we can actually use-- we don't do anything else with the `params` object.

That tells me we should change our interface (someday) in pagesContainer. It never ever ever ever really wants a `params` object. It really wants whatever it gets back from `getPath(params)` as that's what it does to transform `params` for use every single time.

In my perfect world, the well-formed, logically coherent path is passed to the pagesContainer instead of the `params` object. This path could be computed when the redux state is set and we'd never have to think about how the value is actually encoded or decode the form ever again in our view (or the action for that matter). That's where I'd want the logic to live.

PR #483, opened 2016-08-19
Compare paths, not number of files contained, to decide when to reload files for a site's path

brandocalrissian:
Deep-compare all of the files listed in the old and new properties to decide if we should trigger a fetch instead of using the count of files as an equality proxy.


el-mapache:
This is good, can be merged!

wslack:
@brandocalrissian do the honors?

brandocalrissian:
changed the base to staging (master + react-redux) and it checks out ok

PR #484, opened 2016-08-24
properly load yml files into codemirror

el-mapache:
* Shows yml files in codemirror instance

### DIFF
--- a/assets/app/util/documentStrategy.js
+++ b/assets/app/util/documentStrategy.js
@@ -1,0 +1,50 @@
+import { decodeB64 } from './encoding';
+
+const frontmatterDelimiterRegex = /^---\n([\s\S]*?)---\n/;
+
+const initializeYml = content => {
+  return {
+    frontmatter: content,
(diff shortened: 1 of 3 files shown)


assets/app/util/documentStrategy.js
```diff
@@ -0,0 +1,50 @@
+import { decodeB64 } from './encoding';
+
+const frontmatterDelimiterRegex = /^---\n([\s\S]*?)---\n/;
+
+const initializeYml = content => {
+  return {
```
el-mapache: There are a lot of `return {}` statements in this file. They all have different permutations so I'm not sure if there is a way to cut down their number

jeremiak: There are a lot of return statements but I find it pretty easy to read since its all singe functions that are mostly pure

jeremiak:
I added some tests to cover `documentStrategy`. Once it is green, I will merge this PR.

PR #486, opened 2016-08-24
Fed 485

el-mapache:
* Checks for JSON validity before attempting to show simplified tree view in pagesContainer


jeremiak:
Looks good to me!

Issue #445, opened 2016-07-08
Refactor: A site should have a simple media gallery to view all of their assets


wslack:

el-mapache:
@jeremiak Fairly certain this exists, I think it was a view you added pretty early on to the redux PR?

PR #488, opened 2016-08-25
Fed 445

el-mapache:
Addresses issue https://github.com/18F/federalist/issues/445

* Fixes up media gallery to use a grid.
* Fixes bug with image type detection method


jeremiak:
🚢

PR #492, opened 2016-08-28
Draft preview publish

el-mapache:
* User can create draft branches
* User can delete a draft branch
* User can merge a draft branch into the site's master branch

### DIFF
--- a/assets/app/actions/routeActions.js
+++ b/assets/app/actions/routeActions.js
@@ -1,0 +1,12 @@
+import { navigationTypes } from '../constants';
+import store from '../store';
+
+export default {
+  redirect(path) {
+    store.dispatch({
+      type: navigationTypes.UPDATE_ROUTER,
+      method: 'push',
+      arguments: [path]
+    });
+  }
+}
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -77,6 +77,18 @@ export default {
     }).catch(error => alertActions.alertError(error.message));
   },
 
+  createPR(site, head, base) {
+    return github.createPullRequest(site, head, base).then((pr) => {
+      return github.mergePullRequest(site, pr);
+    }).then(() => {
+      return github.deleteBranch(site, head);
+    }).then(() => {
+      this.fetchBranches(site);
+    }).then(() => {
+      return alertActions.alertSuccess(`${head} merged successfully`);
+    }).catch(error => alertActions.httpError(error.message));
+  },
+
   createCommit(site, path, fileData, message = false, sha = false) {
     const b64EncodedFileContents = encodeB64(fileData);
     const siteId = site.id;
@@ -89,7 +101,7 @@ export default {
 
     if (sha) commit = Object.assign({}, commit, { sha });
 
-    github.createCommit(site, path, commit).then((commitObj) => {
+    return github.createCommit(site, path, commit).then((commitObj) => {
       alertActions.alertSuccess('File committed successfully');
 
       store.dispatch({
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -222,11 +222,23 @@ export default {
 
   createDraftBranch(site, path) {
     const branchName = `_draft-${encodeB64(path)}`;
-    const sha = site.branches.find((branch) => {
-      return branch.name === site.defaultBranch;
-    }).commit.sha;
+    const existingDraftBranch = site.branches.find(branch => {
+      return branch.name === branchName;
+    });
+    let githubMethod;
+    let sha;
 
-    return github.createBranch(site, branchName, sha).then(() => {
+    if (existingDraftBranch) {
+      githubMethod = 'updateBranch';
+      sha = existingDraftBranch.commit.sha;
+    } else {
+      githubMethod = 'createBranch';
+      sha = site.branches.find((branch) => {
+        return branch.name === site.defaultBranch;
+      }).commit.sha;
+    }
+
+    return github[githubMethod](site, branchName, sha).then(() => {
       this.fetchBranches(site);
       return branchName;
     });
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -235,7 +235,7 @@ export default {
 
   deleteBranch(site, branch) {
     return github.deleteBranch(site, branch).then(() => {
-      this.fetchBranches(site);
+      return this.fetchBranches(site);
     }).catch((error) => alertActions.httpError(error.message));
   }
 }
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -193,6 +193,17 @@ export default {
         arguments: [`/sites/${site.id}`]
       });
     }).catch((error) => alertActions.httpError(error.message));
+  },
+
+  getBranches(site) {
+    return github.fetchBranches(site).then((branches) => {
+
+      store.dispatch({
+        type: siteActionTypes.SITE_BRANCHES_RECEIVED,
+        siteId: site.id,
+        branches
+      });
+    });
   }
 }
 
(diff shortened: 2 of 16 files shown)


el-mapache:
Includes fix to address issue https://github.com/18F/federalist/issues/496

el-mapache:
Addresses issue #492

assets/app/actions/routeActions.js
```diff
@@ -0,0 +1,12 @@
+import { navigationTypes } from '../constants';
+import store from '../store';
+
+export default {
+  redirect(path) {
+    store.dispatch({
+      type: navigationTypes.UPDATE_ROUTER,
```
brandocalrissian: I'm going to start pushing for action creators soon to encapsulate everything that makes a valid action. We don't have any yet, but this is a good place to start bringing it up. It's nice to have one place that tells you, "Give me a path and I'll give you a proper redux action for updating the router." This is also a separate concern from actually choosing the right action to create and send to the store for dispatch.

assets/app/actions/siteActions.js
```diff
@@ -77,21 +77,35 @@ export default {
...
+  createPR(site, head, base) {
+    return github.createPullRequest(site, head, base).then((pr) => {
+      return github.mergePullRequest(site, pr);
+    }).then(() => {
+      return github.deleteBranch(site, head);
+    }).then(() => {
+      this.fetchBranches(site);
+    }).then(() => {
+      return alertActions.alertSuccess(`${head} merged successfully`);
+    }).catch(error => alertActions.httpError(error.message));
```
brandocalrissian: Does it make sense to have any error recovery if any of these operations fail? I'm thinking there's nothing to really recover from (the PR got merged, but the branch couldn't be deleted for whatever reason so we'd need to manually address it later) but am happy to entertain options.

assets/app/components/site/editor/editorContainer.js
```diff
@@ -106,7 +133,63 @@ class Editor extends React.Component {
...
+    if (draftBranch) {
+      this.submitFile(draftBranch.name);
+    } else {
+      siteActions.createDraftBranch(site, path).then((branchName) => {
+        return this.submitFile(branchName);
+      }).then(() => {
+        const fileName = path.split('/').pop();
+        const branch = formatDraftBranchName(path);
+
+        routeActions.redirect(`/sites/${site.id}/edit/${branch}/${fileName}`);
```
brandocalrissian: We're encoding these paths multiple times in this file and the paths drive functionality. Can we extract lines like these to something that says "redirectToFileOnBranch" or a better name and reuse it instead of duplicating here and L64, L68 for instance? We'd get DRY to improve maintenance and have a nice name that would cause us to not have to think so much when we look at these lines.

assets/app/components/site/editor/editorContainer.js
```diff
@@ -106,7 +133,63 @@ class Editor extends React.Component {
...
+    }
+  }
+
+  deleteDraft() {
+    const { site } = this.props;
+    const { path } = this.state;
+
+    siteActions.deleteBranch(site, formatDraftBranchName(path)).then(() => {
+      alertActions.alertSuccess('Draft successfully deleted');
+      routeActions.redirect(`/sites/${site.id}`);
```
brandocalrissian: Yeah, at a code level like this file, I'd understand it more quickly we asked something to "displaySite" instead of the lower level "redirect to a particular URI we're crafting right here".

assets/app/components/site/editor/editorContainer.js
```diff
@@ -106,7 +133,63 @@ class Editor extends React.Component {
...
+  commitOrPR() {
+    const { site } = this.props;
+    const { path } = this.state;
+    const draftBranch = getDraft(path, site.branches);
+
+    if (!draftBranch) {
+      this.submitFile();
+    }
+
+    this.submitFile(draftBranch.name).then(() => {
```
brandocalrissian: can we come up with a good name for this block and call it instead of embedding the code within this function? We're named "commitOrPR", so I'm guessing the previous `submitFIle()` call is the commit path while the rest of this is the PR path. That's what I get when reading the following lines. I'd like to avoid reading and guessing because I'm very lazy and can only follow so much logic-- I'm happier when you tell me what the abstraction is called so I can move on to what I'm really looking for.

assets/app/reducers/sites.js
```diff
@@ -28,54 +30,72 @@ export default function sites(state = initialState, action) {
...
     return mapPropertyToMatchingSite(state, action.siteId, {
       files: newFiles.concat(leftBehind)
     });
+  }
 
   case siteActionTypes.SITE_DELETED:
     return state.filter((site) => site.id != action.siteId);
 
   case siteActionTypes.SITE_FILE_CONTENT_RECEIVED:
+    // get site we want to check files for from the list of user's sites
```
brandocalrissian: can we extract the function with this name instead of relying on the comment?

assets/app/reducers/sites.js
```diff
@@ -28,54 +30,72 @@ export default function sites(state = initialState, action) {
...
     const currentSite = state.find((site) => site.id === action.siteId);
+    // get list of files associated with the current site
```
brandocalrissian: likewise, can we extract a function with this name?

assets/app/util/branchFormatter.js
```diff
@@ -0,0 +1,16 @@
+import {encodeB64, decodeB64 } from './encoding';
+
+// format draft branch anme
```
brandocalrissian: i believe the function name and don't need this comment. I love how this code in this file has been extracted for reuse by others.

assets/app/util/branchFormatter.js
```diff
@@ -0,0 +1,16 @@
...
+export const formatDraftBranchName = (path) => {
+  return `_draft-${encodeB64(path)}`;
+}
+
+// determine if a path has a draft
```
brandocalrissian: yeah, don't need this comment either

brandocalrissian:
Had a few nitpick comments but it makes sense to me.

assets/app/actions/siteActions.js
el-mapache: I think if the delete branch failed we would want to have some mechanism to clean it later, although I'm not sure if that scenario is going to be common enough to warrant a change at this point. 

I do with you could make transactional calls to the github api, and they could just roll back your request if one of the operations failed.

assets/app/actions/routeActions.js
el-mapache: Are you saying this is an example if the type of action creator you'd like to see, or that this action creator should be changed?

assets/app/components/site/editor/editorContainer.js
el-mapache: I think there is a pattern in a few different that is essentially 'show a message, then redirect somewhere'. Would you see that kind of behavior ultimately ending up in an action creator? Or a wrapper component that passes down a `displayXResource` method as a prop?

assets/app/util/branchFormatter.js
el-mapache: ah yes, I think these were placeholder comments before  functions were written

assets/app/actions/routeActions.js
brandocalrissian: This is the type of action creator I'd like to see-- we do this in a few places and it's an easy thing to extract. I'm happy to run in and make this happen everywhere-- it's a quick quick quick refactor.

assets/app/actions/siteActions.js
brandocalrissian: Sounds reasonable. If the network drops in the middle of deleting the branch, then there's not much of a recovery opportunity. We can wait and see if something actually is a problem before trying to recover "nicely"-- in this case, manual recovery isn't complicated and possible.

brandocalrissian:
Looks good to me!

PR #495, opened 2016-08-31
Fed 494

el-mapache:
* Use the proper default commit message when saving changes to a file if a user hasn't updated
the message field


jeremiak:
👍 

Opened #496 for minor bug

PR #503, opened 2016-09-06
Use another buildpack

jeremiak:


brandocalrissian:
https://github.com/cloudfoundry/nodejs-buildpack#v1.5.19 does not appear to be on cloud.gov, so we're trying to use https://github.com/cloudfoundry/nodejs-buildpack#v1.5.18 to get the right version of node.

PR #504, opened 2016-09-06
Use autopilot to deploy

jeremiak:
Using before_deploy values lifted from the openfec project
and using the autopilot plugin to deploy automatically to
cloud.gov from travis


jeremiak:
Merging as we have to have the code be on `master` to test.

PR #516, opened 2016-09-16
fixes issue where code compiles `this` to undefined

el-mapache:
Seems like babel is forcing references to `this` to be undefined if the method doesn't live within a class or fat arrow function. Using `.call` and passing the `this` context the function should execute in was causing the 'save & submit' button to break.

The function in question had to be moved back into the class to work as expected.

### DIFF
--- a/assets/app/components/site/editor/editorContainer.js
+++ b/assets/app/components/site/editor/editorContainer.js
@@ -30,15 +30,6 @@ const alertAndRedirect = (message, uri) => {
   routeActions.redirect(uri);
 };
 
-const openPRWithHead = (branchName, site) => {
-  return this.submitFile(branchName).then(() => {
-    return siteActions.createPR(site, branchName, site.defaultBranch);
-  }).then(() => {
-    routeActions.redirect(`/sites/${site.id}`);
-  });
-}
-
-
 let insertFn;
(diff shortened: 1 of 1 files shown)


wslack:
hmmmm, interesting. @jmhooper are you able to do reviews of code given your staffing status?

brandocalrissian:
I suspect we could handle this `this` issue by binding this in the call to `openPRWithHead`. The comment (https://github.com/18F/federalist/pull/516#issue-177522744) seems to indicate that was a non-starter, but the commit is moving the function back into the Editor Component context. For now, let's move forward as the only issue is if we really want this function to be something externally visible for anyone who has a handle on an Editor. I'm happy to revisit this later when we can get some test coverage on it.

el-mapache:
@brandocalrissian yes, although it is not reflected in the PR, I did have a version that used openPRWithHead.call(this, all, the, args). I think the issue is that the fat arrow always looks at the surrounding context when deciding which scope `this` should be bound to. Since our files are evaluated in strict mode (possibly babel's default?), `this` will be undefined in the global scope. 

But yes, I'd like to get test coverage first, then refactor/move some of the methods that don't deal with rendering the view out of the component class.

PR #520, opened 2016-09-20
Fix preview link

el-mapache:
* Updates default s3 bucket computation to handle staging env
* Adds preview link to individual site item on `/sites`
* Adds preview link to `View this Website` buttons in the `pagesHeader` component

### DIFF
--- a/api/controllers/PreviewController.js
+++ b/api/controllers/PreviewController.js
@@ -16,20 +16,19 @@ module.exports = {
     if ((key).slice(-1) === '/') key = key + 'index.html';
 
     var object = s3.getObject({
-          Bucket: sails.config.build.s3Bucket,
-          Key: key
-        }).on('httpHeaders', function(statusCode, headers) {
-          var redirect = headers['x-amz-website-redirect-location'] ||
-            headers['X-Amz-Website-Redirect-Location'];
-          if (redirect) return res.redirect(redirect);
-          res.set(headers);
-        }),
-        stream = object.createReadStream().on('error', function(error) {
-          var file = key.split('/').pop().indexOf('.') !== -1,
-              notFound = error.statusCode === 404;
-          if (!file && notFound) return res.redirect(req.path + '/');
-          res.send(error.statusCode, error.message);
-        }).pipe(res);
+      Bucket: sails.config.build.s3Bucket,
+      Key: key
+    }).on('httpHeaders', function(statusCode, headers) {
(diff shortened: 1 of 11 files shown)


api/controllers/PreviewController.js
```diff
@@ -16,20 +16,19 @@ module.exports = {
...
-          if (redirect) return res.redirect(redirect);
-          res.set(headers);
-        }),
-        stream = object.createReadStream().on('error', function(error) {
-          var file = key.split('/').pop().indexOf('.') !== -1,
-              notFound = error.statusCode === 404;
-          if (!file && notFound) return res.redirect(req.path + '/');
-          res.send(error.statusCode, error.message);
-        }).pipe(res);
+      Bucket: sails.config.build.s3Bucket,
```
el-mapache: The indentation was bugging me here, so I changed it

api/models/Site.js
```diff
@@ -5,7 +5,11 @@
 * @docs        :: http://sailsjs.org/#!documentation/models
 */
 
-var DEFAULT_BUCKET = 'federalist.18f.gov.s3-website-us-east-1.amazonaws.com';
+var env = process.env.NODE_ENV;
+
+var DEFAULT_BUCKET = env === 'staging' ?
+  'federalist-staging.18f.gov.s3-website-us-east-1.amazonaws.com' :
+  'federalist.18f.gov.s3-website-us-east-1.amazonaws.com';
```
brandocalrissian: Do we want to make the bucket configuration come from the environment so we don't have this configuration/switch in code? If we always get the settings from the same environment variable, we don't need to know if we're dev/staging/production or whatever.

I'm fine keeping this for now as we had the production config here already; it's something to think about.

el-mapache: Yeah that is a valid point.

brandocalrissian: This is my only comment! Everything looks good!

I'm happy with merging this PR and making the "pull the bucket from the env" another (small) card/ticket/issue/task/whatever we're calling it.

el-mapache: I think we should go ahead and merge this, we'll still need some kind of mechanism for ensuring the user can get redirected to one of these URLs when the app is running locally (like manually setting env variables) and I don't want to have more moving parts to keep track of at this point

brandocalrissian: Sounds good. Merging.

PR #521, opened 2016-09-20
update robots file to disallow indexing of preview branches

wslack:


assets/robots.txt
```diff
@@ -4,5 +4,5 @@
 
 
 # To prevent search engines from seeing the site altogether, uncomment the next two lines:
-# User-Agent: *
-# Disallow: /
+User-Agent: *
+Disallow: /preview
```
brandocalrissian: Should this actually be `/preview/` with a trailing slash? I had to hunt down some docs as I haven't bothered with robots.txt for ages. Can we test this?

el-mapache:
@wslack merging this, fairly certain we don't need the trailing slash and if we do, we can always open another PR.

PR #524, opened 2016-09-26
Fixes draft branch bugs in editor component

el-mapache:
Various bug fixes to the editor component.

* Fixes bug where a user hitting the browser back button, when on a draft branch, is navigated to the `master` branch of that page

* Modifies `routeActions.redirect` to accept an optional `method` argument. This lets the programmer call arbitrary methods in the browserHistory API. Defaults to `push`

* Fixes base64 draft branch name computation by supplying the entire path (instead of just the filename) to the `formatDraftBranchName` function. Upon branch creation, the draft branch name is computed using the file's name and its relative path in the project (e.g. `pages/about.md`). Subsequent computations, however, only used the file's name (e.g. `about.md`). This led to improper redirects and a failure to find draft branches that corresponded with the file

* Removes integration test cruft from package.json


assets/app/components/site/editor/editorContainer.js
```diff
@@ -22,7 +22,7 @@ const propTypes = {
 };
 
 const redirectToFileOnBranch = (siteId, branch, filePath) => {
-  routeActions.redirect(`/sites/${siteId}/edit/${branch}/${filePath}`);
+  routeActions.redirect(`/sites/${siteId}/edit/${branch}/${filePath}`, 'replace');
```
brandocalrissian: Can we have two different functions in routeActions? One for push and one for replace? In this manner we can think about what people who call each function in routerAction really mean without having to expose the particular 'push' or 'replace' concept to outsiders.

Naturally, my first question as a client would be "What are the valid methods? Are push and replace the only ones? When would I use each of these methods? What happens if I have a typo in my method?"

el-mapache: We can have two different methods, but I believe the concepts of `push` and `replace` will still need to be communicated. The consumer of these methods will want to know whether or not their redirect is adding an entry to the browser's history or replacing the existing entry with a new one.

brandocalrissian: Whatever you think the proper names to communicate those concepts to callers to the methods works for me!

el-mapache: cool! let me know what you think about the updates I made yesterday

brandocalrissian: Oh! I should have looked closer. Github had both commits labeled as "7 days ago". These look great. Merging.

PR #525, opened 2016-09-26
Get 100% test coverage for app/actions...

brandocalrissian:
...and start extracting/refactoring along the way to make things easier to follow and maintain.

This was rebased from staging of September 23.

After this, our code coverage stats look like: 33.39% Statements (410/1228), 31.94% Branches (122/382), 50.39% Functions (129/256), 31.95% Lines (370/1158)


el-mapache:
Looks like a lot of good work! Thanks for digging into this and getting tests around the actions. Will comment on individual lines.

assets/app/actions/createDraftBranchName.js
```diff
@@ -0,0 +1,5 @@
+import { encodeB64 } from '../util/encoding';
+
+export default path => {
```
brandocalrissian: You are absolutely correct these one-off functions are not actions-- they are in this directory not because they share the same class of functionality as the "real actions" in this directory but because I didn't put them in a different directory when I extracted them. I didn't immediately have a good name for where to put them (utility seems a bit vague and grab-baggy) so I left them there.

I can be very lazy.

I'll put them somewhere and take a swing at better naming/organization.

assets/app/actions/actionCreators/alertActions.js
```diff
@@ -0,0 +1,39 @@
+import { alertActionTypes as types } from '../../constants';
+
+const authErrorType = types.AUTH_ERROR;
```
brandocalrissian: I like having constants located where they make sense and matter instead of extracted to some "always include me as a dependency" file. This is a step along the path toward moving constants to the code that is coupled with their definition.

I also like exporting the types as we can eventually write a test to ensure no two action types have a name collision.

Our tests should verify the actionCreator is well formed and meets our expectations.

assets/app/actions/siteActions.js
```diff
@@ -1,234 +1,132 @@
...
+    }).catch(alertError);
   },
 
   fetchSiteAssets(site) {
     const config = site['_config.yml'];
     const assetPath = (config && config.assetPath) || 'assets';
 
-    return github.fetchRepositoryContent(site, assetPath).then((assets) => {
+    const filterAssetsWithTypeOfFile = (assets) => {
       return assets.filter((asset) => {
```
brandocalrissian: It could be. Would make the tests cleaner, and a cleaner test is generally indicative of clearer code.

assets/app/actions/makeCommitData.js
```diff
@@ -0,0 +1,29 @@
+import { encodeB64 } from '../util/encoding';
```
brandocalrissian: Putting it into a private function makes the tests much messier as it increases the paths we need to test. It's not an "action", that's for sure (see earlier, concurring comment).

assets/app/actions/dispatchActions.js
```diff
@@ -0,0 +1,72 @@
+import { dispatch } from '../store';
```
brandocalrissian: The "actions" in siteActions, for instance, do much more than simply dispatch actions to the redux store. They also make multiple calls to github or the federalist APIs.

As part of refactoring after writing some tests, these "create an action and dispatch it to the redux store" functions emerged as a related cohesive set of logic. Pulling them out separated the business logic of the various API calls from the state changes spurred by dispatching actions to the redux store. Now the business logic siteActions do not need to understand what it means to create an action and dispatch it to the redux store.

The inconsistency comes from my not following through and performing the same transformations to the other actions (alertActions, buildActions, ...). That's confusing. The logical separation is between the non-redux-state-transforming logic and the redux-state-transforming bits.

Looking at where the code says it wants to go...

assets/app/actions/actionCreators/alertActions.js
el-mapache: This makes sense, but if the action type constants are going to be 100% coupled with the action creators, can we just remove the constants file entirely and have the creators be the canonical source of types?

I understand that we need to check that the type of action returned by the action creator matches the string literal of the constant, but it still feels weird to me to expose a property that is just a test hook.....I suppose the alternative is to define the constants again in the test files and that is certainly not what we want.

assets/app/actions/createDraftBranchName.js
el-mapache: Yes, the `util` directory is also lazy :( Is there a good way to define some small subset of folders we can group these functions into? It would be nice to come up with something that would be mostly obvious or easily documented so that future utility type functions could be easily added to the right folder.

Something like `services` for API interactions, `transforms` for data transforms, `filters` for things that take a list and reduce the contents. I don't know exactly what the naming conventions should be and I don't want to drift too far into angular land with them.

Does that seem like a reasonable approach?

assets/app/actions/dispatchActions.js
el-mapache: Cool, that makes sense to me. I actually think that the placement of business logic is, weirdly, not really discussed much in terms of React application architecture, this kind of separation between the composition and dispatching actions seems really helpful!

assets/app/actions/createDraftBranchName.js
brandocalrissian: If we can make a logical grouping like this, I'm all for it. We can always regroup/rename later as the design evolves.

assets/app/actions/actionCreators/alertActions.js
brandocalrissian: I support dropping any "this is a bag of constants" files whenever we can.

Exposing the type right now for testing seems questionable until it has real meaning in our application or real merit in tests, but copying the "ACTION NAME" constants in the test is a bit wonky too, I agree.

assets/app/actions/dispatchActions.js
brandocalrissian: I have a demo design app put together for a different client that's developed opinions about these very ideas here: https://github.com/18F/react-redux-event-demo

brandocalrissian: I expect to have some more clean up work to do on that project after I extend its functionality a touch, so I wouldn't say it's quite where it will end up. However, it's conceptually something I'm happy to get behind.

el-mapache: cool, I'll take a look at that

assets/app/actions/createDraftBranchName.js
el-mapache: ha, yes, I thought you had already seen that! I also didn't realize we left the duplicate code in that file...

PR #533, opened 2016-09-28
update origin caption

wslack:


wslack:
Assigning @jeremiak to make sure I'm not missing a use case, given how the container build works.

el-mapache:
I'd like to hear what @jeremiak has to say, but I think this is still confusingly worded. Why are we providing this value to the user at all if it's only going to be used on the 18F side?

I think we should mention both why the user might need to know about this url, and that the intention of the readonly form field is to provide an easy way for the user to copy/paste the URL.

wslack:
@el-mapache really, I just want it in a place where anyone can easily find it, but it doesn't really need to be visible to the user.

wslack:
made an edit!

el-mapache:
I'm just going to merge this for the time being, we can explore copy or UI changes at some point in the future if user's are still confused.

PR #538, opened 2016-10-03
Fed 532

el-mapache:
* Adds form fields to easily change page layout and title
* Adds stage-2 babel plugin to allow use of the `...` operator


assets/app/components/site/editor/configs/pageSettings.js
```diff
@@ -0,0 +1,100 @@
...
+
+const getPageLayoutsFromConfig = (config) => {
+  if (!config.defaults) {
+    return [];
+  }
+
+  return config.defaults[0].values.layout;
+};
+
+const buildBaseFormFields = (layouts, handleChange) => {
```
el-mapache: Not super happy about having this logic here, its making the component difficult to test. Not sure what the right abstraction is for this collection of private methods

brandocalrissian: As a start, can you extract the yaml reading code from the rest and test it independently?

jmhooper:
I like this. Pulled and took it for a test drive and it seems to work the way we want it to.

PR #545, opened 2016-10-05
skips creating a pr if one is already open for the draft in question

el-mapache:
* In the case of an existing PR, return and use it instead of attempting to create a new one

### DIFF
--- a/assets/app/actions/siteActions.js
+++ b/assets/app/actions/siteActions.js
@@ -143,7 +143,17 @@ export default {
   },
 
   createPR(site, head, base) {
-    return github.createPullRequest(site, head, base).then((pr) => {
+    return github.fetchPullRequests(site).then((openPrs) => {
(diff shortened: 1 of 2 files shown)


jmhooper:
This looks good to me. My only concern would be about the fact that `fetchPullRequests` doesn't seem to handle pagination There's the chance that a pre-existing PR could be back in the history somewhere and get duplicated.

I'm not sure what the backlog of PRs on these sites looks like, but I can't imagine it'd be a problem for many.

el-mapache:
That's true, none of the existing federalist methods handle pagination. I don't think its been an issue with users thus far, but if people start making large sites with federalist there could be issues further down the line.

jmhooper:
Yeah, I looked at the rest of the requests the API client handles and they didn't either. It should be fine. Just something to watch out for and consider if ends up becoming a problem down the line.

This looks 👍 to merge from what I can tell.

PR #549, opened 2016-10-12
Fed 547

el-mapache:
* Fetch navigation.json from s3
* Use navigation.json when available
* Fixes bug where retrieved single file isn't added to site files array


brandocalrissian:
Looks good. Merging. Saw most of this yesterday afternoon during pairing.

PR #553, opened 2016-10-18
Get some coverage of the navigationJsonPageList, fix a bug.

### LINKED ISSUE #552: Children in _navigation.json aren't recognized as draft paths
The draft logic only applies to the parent page. The children are not recognized as drafts.

brandocalrissian:
Found a bug in this component, so tests help.

Fixes #552.

### DIFF
--- a/assets/app/components/site/Pages/navigationJsonPageList.js
+++ b/assets/app/components/site/Pages/navigationJsonPageList.js
@@ -31,7 +31,7 @@ const emitPages = (pages, siteId, defaultBranch, branches) => {
         isPageDirectory={ false }
         hasDraft={ pathHasDraft(page.path, branches) }
       >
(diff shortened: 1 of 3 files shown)


brandocalrissian:
The travis build failed with "Cannot find module '../../../../../../assets/app/components/site/pages/navigationJsonPageList'"

Will fix. Didn't happen locally, so clearly I can't choose the one in front of me.

brandocalrissian:
Oh. Case sensitivity strikes again. On OS X, "pages" matches "Pages".

el-mapache:
There is another failure around the radio button test, if you change the expectation from this:
`expect(checkedInputWrapper.find('input[checked=""]')).to.have.length(1);`
 to this:
`expect(/checked/.test(checkedInputWrapper.html())).to.be.true;`
it will work. Not sure why it just started breaking

PR #560, opened 2016-10-24
10/24/2016 Merge staging-test back into staging

el-mapache:
* Merges the work we've done on `staging-test` back into staging.


el-mapache:
Just the last commit, but that _shouldn't_ affect anything except having `staging` built and deployed on travis instead of the `staging-test` branch

brandocalrissian:
Ah, I see. Seems good to me as I cannot really test without running everything through travis. Since it passed travis-ci, there's lots of confidence there. Merging.

PR #563, opened 2016-10-25
Merging GovCloud staging changes into staging

el-mapache:
* Port all changes made for gov cloud testing into the staging branch. Once this is merged, E/W staging will no longer be updated or supported


package.json
```diff
@@ -6,24 +6,33 @@
   "keywords": [],
   "dependencies": {
     "async": "^1.4.2",
-    "autoprefixer": "^6.1.2",
```
brandocalrissian: this seems to be causing the travis-ci failure

brandocalrissian:
travis failure is here:
> federalist@0.0.0 autoprefixer /home/travis/build/18F/federalist
> postcss --use autoprefixer -r assets/styles/styles.css
module.js:327
    throw err;
    ^
Error: Cannot find module 'autoprefixer'

Autoprefixer was removed from the package.json.

brandocalrissian:
Are we actually using autoprefixer with this code? Should we add it back or drop it from the build process?

el-mapache:
Yes that's correct, I stashed a couple of changes I made and haven't reconciled those with the branch quite yet.

Issue #559, opened 2016-10-21
Can't deploy staging to GovCloud


el-mapache:
This issue is arose when trying to deploy the staging configuration with the node devDependencies.

When doing a cf push without uploading node_modules, the build process was running out of disk space.

```
cp: cannot create regular file ‘/tmp/contents354112921/app/node_modules/babel-preset-es2015/node_modules/babel-plugin-transform-es2015-function-name/node_modules/babel-helper-function-name/node_modules/babel-traverse/node_modules/lodash/fp/times.js’: No space left on device
```

Cloud Foundry has a 1 GB limit on application uploads, it has a 2 GB disk space limit for a running app (It's not clear what limit is enforced in a build process) https://docs.cloudfoundry.org/devguide/deploy-apps/large-app-deploy.html

Attempting to build locally before pushing resulted in the following error:

```
       > fsevents@1.0.14 install /tmp/app/node_modules/fsevents
       > node-pre-gyp install --fallback-to-build

       node-pre-gyp ERR! Tried to download: https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.0.14/fse-v1.0.14-node-v46-linux-x64.tar.gz
       node-pre-gyp ERR! Pre-built binaries not found for fsevents@1.0.14 and node@4.2.3 (node-v46 ABI) (falling back to source compile with node-gyp)
       gyp ERR! UNCAUGHT EXCEPTION
       gyp ERR! stack Error: Cannot find module 'es5-ext/string/#/contains'
       gyp ERR! stack     at Function.Module._resolveFilename (module.js:337:15)
       gyp ERR! stack     at Function.Module._load (module.js:287:25)
       gyp ERR! stack     at Module.require (module.js:366:17)
       gyp ERR! stack     at require (module.js:385:17)
       gyp ERR! stack     at Object.<anonymous> (/tmp/app/node_modules/d/index.js:6:21)
       gyp ERR! stack     at Module._compile (module.js:435:26)
       gyp ERR! stack     at Object.Module._extensions..js (module.js:442:10)
       gyp ERR! stack     at Module.load (module.js:356:32)
       gyp ERR! stack     at Function.Module._load (module.js:311:12)
       gyp ERR! stack     at Module.require (module.js:366:17)
       gyp ERR! stack     at require (module.js:385:17)
       gyp ERR! System Linux 4.4.0-42-generic
[...]

PR #568, opened 2016-10-26
[ work in progress ] Debugging GovCloud deployment

rogeruiz:
Debugging branch for #559

This PR is a working progress for debugging the issues with #559. Commits here should be reverted or squashed once debugging is complete.

AC:
- [x] Asses the dependencies
- [x] Update the resources so that disk and memory quotas aren't hit
- [x] Document what the fixes are 18F/cg-site#494
- [ ] Help automate on CI


rogeruiz:
I was able to get this to deploy a few times running the following commands:

```
$ ./scripts/build-on-ci.sh && cf push -f staging_manifest.yml
```

@el-mapache let me know if this works for you too! I can clean up the history to remove the debugging bits, but this should work. I think using `node-pre-gyp` or `node-gyp` on :cloud:.gov is going to give you a hard time based on disk and memory constraints.

staging_manifest.yml
```diff
@@ -2,17 +2,18 @@
...
 - federalist-staging-rds
 - federalist-staging-redis
 - federalist-staging-s3
 - federalist-staging-env
 env:
-  NODE_ENV: staging
+  NODE_ENV: production
   SAILS_LOG_LEVEL: verbose
-  NPM_CONFIG_PRODUCTION: false
+  NPM_CONFIG_PRODUCTION: true
```
rogeruiz: When deploying on :cloud:.gov, you need to ensure that your running `NODE_ENV: production` and `NPM_CONFIG_PRODUCTION: true` otherwise you're going to be installing some `devDependencies`. The installation of certain libraries will hit the memory and disk limits of the container that it is being deployed to.

scripts/build-on-ci.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+set -u
+
+time rm -rf ./node_modules
+time npm install
+npm run build
+time rm -rf ./node_modules
+time npm install --production
```
rogeruiz: This double install process is what need to be happening on the CI. Since the CI should be doing the building of the project, it will need to run `npm install` without the production flag set. This way it can run `npm run build`. Once that runs, it needs to vendor the dependencies for the deployment, but `devDependencies` shouldn't exist in the `node_modules` directory. Because of this, you'll need to remove them and then reinstall them with the `--production` flag.

I would make sure that your CI isn't running with the `NODE_ENV=production` nor `NPM_CONFIG_PRODUCTION=true` variables set like that.

scripts/debug-directory-size.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo '==================================================';
+
+echo '==================================================';
+echo 'top Debug Info';
+echo '==================================================';
+
+top -b -n 3 | head -n 10;
+
+echo '==================================================';
```
rogeruiz: All of this file will be removed.

package.json
```diff
@@ -83,11 +67,12 @@
     "migrate:create": "grunt db:migrate:create",
     "migrate:up": "grunt db:migrate:up || true",
     "migrate:down": "grunt db:migrate:down",
-    "uswds": "npm run uswds:css && npm run uswds:images && npm run uswds:fonts",
-    "uswds:css": "cp node_modules/uswds/dist/css/uswds.css assets/styles/",
-    "uswds:images": "cp -r node_modules/uswds/dist/img/ ./assets/images/uswds",
-    "uswds:fonts": "cp node_modules/uswds/dist/fonts/* ./assets/fonts",
-    "codemirror": "cp node_modules/codemirror/lib/codemirror.css assets/styles/",
+    "debug:fileSizeInMB": "./scripts/debug-directory-size.sh",
```
rogeruiz: This needs to be removed.

package.json
```diff
@@ -55,13 +39,13 @@
...
     "underscore": "^1.8.3",
-    "uswds": "^0.9.0",
     "validator": "^3.39.0",
     "whatwg-fetch": "^1.0.0",
     "yamljs": "^0.2.8"
   },
   "scripts": {
-    "install": "npm run uswds && npm run codemirror && npm run migrate:up",
+    "preinstall": "npm run debug:fileSizeInMB",
+    "postinstall": "npm run debug:fileSizeInMB && npm run uswds && npm run codemirror && npm run migrate:up",
```
rogeruiz: The `debug:fileSizeInMB` needs to be removed as well.

staging_manifest.yml
```diff
@@ -2,17 +2,18 @@
 applications:
 - name: federalist-staging
 buildpack: https://github.com/cloudfoundry/nodejs-buildpack#v1.5.4
+stack: cflinuxfs2
 domain: fr.cloud.gov
 disk_quota: 2G
-memory: 512MB
+memory: 1G
```
rogeruiz: At `1G` there is no longer a memory issue. But I'm certain this was being caused by the `node-pre-gyp` installation of `node-sass`.

config/connections.js
```diff
@@ -90,9 +90,9 @@ module.exports.connections = {
...
-  memory: {
-    adapter: 'sails-disk',
-    inMemory: true
-  }
+  // memory: {
+    // adapter: 'sails-disk',
+    // inMemory: true
+  // }
 
 };
```
rogeruiz: This `connections.js` file needs to be refactored a bit. I think all these changes can be reverted.

staging_manifest.yml
rogeruiz: Which was being caused by the `staging_manifest.yml` not configuring the production environment properly. This was causing `npm install` to install all the dependencies like `node-sass` which were crashing the system with an out-of-memory error.

staging_manifest.yml
el-mapache: Noted, we will most liekly be changing the way sails uses env variables and user-provided services to configure the application at start up time

staging_manifest.yml
rogeruiz: No big deal, I would rely on a separate environment variable for staging configuration loading rather than production. `NODE_ENV` is very overloaded and triggers many side effects if it's not set properly.

scripts/build-on-ci.sh
el-mapache: 👍 I like the addition of the timing information

rogeruiz:
@el-mapache feel free to close and delete this branch once you have the changes we mentioned above in the repo as either documentation or code changes. You can `git cherry-pick` any of the commits from this branch locally on to whatever branch you're working on. Just be sure you do a `git fetch --all` before attempting to cherry pick any commits to make sure you download all the git refs.

el-mapache:
Closing, I've incorporated the work here https://github.com/18F/federalist/pull/579. Thanks for the assistance @rogeruiz

PR #578, opened 2016-10-31
Sets a value for origin in config/cors.js

el-mapache:
* Sets CORS origin to a value, per recommendation on security issue here: http://sailsjs.org/documentation/concepts/security/cors


brandocalrissian:
We might want to see if we can make this configurable by environment at some point (no need for prod/staging to share the same config), but this works for now.

PR #579, opened 2016-10-31
Govcloud deploy fixes

el-mapache:
Incorporates changes made in https://github.com/18F/federalist/pull/568

* Moves build dependencies into devDependencies to reduce size of installed node modules
* Adds APP_ENV environment variable to avoid overloading NODE_ENV. This was done to avoid having to coordinate merging this PR with renaming existing services on cloud.gov
* Remove config/env/staging.js; The production config file now handles staging and prod environment configs.


rogeruiz:
Is it safe to abandon #568 @el-mapache?

staging_manifest.yml
```diff
@@ -13,9 +13,10 @@ services:
 - federalist-staging-s3
 - federalist-staging-env
 env:
-  NODE_ENV: staging
+  NODE_ENV: production
+  APP_ENV: staging
   SAILS_LOG_LEVEL: verbose
-  NPM_CONFIG_PRODUCTION: false
+  NPM_CONFIG_PRODUCTION: true
```
rogeruiz: You may be able to refactor all this to only use `NPM_CONFIG_PRODUCTION` to true and leave your `NODE_ENV: staging` in order to not need those `APP_ENV` references. 💭 

source: https://devcenter.heroku.com/articles/nodejs-support
> When `NPM_CONFIG_PRODUCTION` is true, npm automatically runs all scripts in a subshell where NODE_ENV is 'production.’

Since it uses a version of Heroku buildpack, it can be safe to assume this works. I would test it.

rogeruiz: Although if the buildpack ever changes, you can be sure that the `NODE_ENV` changes will need to be introduced again. So maybe it's safer to use `APP_ENV` for Sails in the long-run.

el-mapache:
I did test that actually, Sails loads in production mode, probably because it uses `npm run start` to execute the app.js file.

staging_manifest.yml
el-mapache: I did test that! Sails boots up in production mode, probably because it uses `npm run start` to execute the app.js file.

brandocalrissian:
Can we please fix the branch conflicts to make sure someone like me doesn't mess it up? Thank you!

el-mapache:
Yeah, this is like the third time I've done it :D

config/env/production.js
```diff
@@ -47,12 +49,20 @@ if (dbURL) {
...
-if (AWSCreds) {
-  AWS.config.update({
-    accessKeyId: AWSCreds.access_key,
-    secretAccessKey: AWSCreds.secret_key,
-    region: AWSCreds.region || 'us-east-1'
+if (sqsKey && sqsSecret && AWS_S3_CREDS) {
+  module.exports.SQS = new AWS.SQS({
+    accessKeyId: sqsKey
+    secretAccessKey: sqsSecret,
+    region: 'us-east-1'
```
brandocalrissian: should we put this region in an environment variable while we're at it?

brandocalrissian:
I only have the one comment but everything else looks like an improvement.

config/env/production.js
el-mapache: The region is actually in an environment variable, I forgot to update production since we weren't using it!

brandocalrissian: ok, we can env it when it actually makes sense to env it-- no point in changing something when it is unclear when it will be able to have more than one value.

brandocalrissian:
looks good, merging.

PR #581, opened 2016-10-31
Load session secret from env in staging/production

jmhooper:
This commit adds code to the environment files that loads the session secret from the cloud foundry environment instead of using the hardcoded secret in `session.js`.

In development, the hardcoded secret will still be used.

I've already updated the user provided service in govcloud staging so there shouldn't be any configuration necessary in cloud foundry before this can be merged.


el-mapache:
Looks good

PR #582, opened 2016-10-31
[fed-566]: Adds HoC to gate render of pages component if site does not exist

el-mapache:
* Displays a helpful message instead of an error when the site being requested doesn't exist or is being built

### DIFF
--- a/.sailsrc
+++ b/.sailsrc
@@ -1,5 +1,9 @@
 {
   "generators": {
     "modules": {}
+  },
+  "hooks": {
+    "sockets": false,
+    "pubsub": false
   }
 }
--- a/.sailsrc
+++ b/.sailsrc
@@ -1,9 +1,5 @@
 {
   "generators": {
     "modules": {}
-  },
-  "hooks": {
-    "sockets": false,
-    "pubsub": false
   }
 }
--- a/assets/app/components/site/Pages/pagesContainer.js
+++ b/assets/app/components/site/Pages/pagesContainer.js
@@ -3,6 +3,7 @@ import isEqual from 'lodash.isequal';
 import PageList from './pageList';
 import PageListItem from './pageListItem';
 import NavigationJsonPageList from './navigationJsonPageList';
+import hasConfig from '../../higherOrder/hasConfig';
 
 import siteActions from '../../../actions/siteActions';
 
@@ -13,10 +14,6 @@ const propTypes = {
 class Pages extends React.Component {
   constructor(props) {
     super(props);
-  }
-
-  componentDidMount() {
-    fetchFiles(this.props.site, this.props.params);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -82,4 +79,4 @@ const hasNavigationJsonContent = (site) => {
 
 Pages.propTypes = propTypes;
 
-export default Pages;
+export default hasConfig(Pages);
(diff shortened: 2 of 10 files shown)


test/client/app/components/higherOrder/hasConfig.test.js
```diff
@@ -0,0 +1,34 @@
...
+import { shallow } from 'enzyme';
+import hasConfig from '../../../../../assets/app/components/higherOrder/hasConfig';
+
+const mock = React.createClass({
+  render() {
+    return <div></div>
+  }
+});
+
+describe.only('hasConfig()', () => {
```
brandocalrissian: we should remove the "only" here as it causes us to run only these two client tests.

el-mapache: yes, sorry about that

.sailsrc
```diff
@@ -1,5 +1,9 @@
 {
   "generators": {
     "modules": {}
+  },
+  "hooks": {
```
el-mapache: Put this is by mistake, want to test this functionality separately. I'll remove it

assets/app/components/site/Pages/pagesContainer.js
```diff
@@ -15,10 +16,6 @@ class Pages extends React.Component {
     super(props);
   }
 
-  componentDidMount() {
-    fetchFiles(this.props.site, this.props.params);
```
brandocalrissian: Do we no longer need to do this fetch the first time we mount the component? _starts digging around..._

test/client/app/components/higherOrder/hasConfig.test.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import hasConfig from '../../../../../assets/app/components/higherOrder/hasConfig';
+
+const mock = React.createClass({
```
brandocalrissian: FWIW, since everything goes through the babel JSX pipeline (including tests), we should be able to write
`const mock = <div></div>;` instead of having to write the React class boilerplate.

.sailsrc
brandocalrissian: Cool, thank you. Was the last comment I had and you beat me to it!

test/client/app/components/higherOrder/hasConfig.test.js
el-mapache: good point!

el-mapache: Looking at this further, what you suggested won't work, React expects either a string representing a tag like 'div' or an actual React component, not just JSX. Providing a string produces an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html)

Going to leave as it is for now!

brandocalrissian: Aha! I'm almost always using stateless components, so I haven't tried this when testing a full-blown React component honestly.

el-mapache: It does work!

brandocalrissian:
I'm not holding this back any further as I think it's fair to assume we no longer need to fetch files upon mounting the site container. I was going to verify this but got a bit swamped. The code looks good.

Issue #586, opened 2016-11-01
POST /build/status/:id/:token does not 404 if no build is found


jmhooper:
The path to update a build status will respond with a 200 if the token is correct regardless of whether or not there is a build to update. If no build is found for the given ID we should probably 404 here.

Also, the path may make more sense as:

```
POST /build/:id/status?token=:token
```

PR #587, opened 2016-11-01
Setup request spec test harness and test build API

jmhooper:
These commits configures a test environment for request specs that test the API and includes:

- Factories to create test models
- An ephemeral data store so broken test data ins't persisted
- A helper to spoof authenticated sessions.

This test environment is used to test the following actions in the build API:

- find: renders a list of builds
- findOne: renders a specific build
- status: updates a build's status

I opened #586 while writing these specs to be addressed later.


api/models/Site.js
```diff
@@ -75,17 +75,7 @@ module.exports = {
...
-          if (err) return done(err);
-          if (!site.users[0]) return done();
-          var build = {
-              user: site.users[0].id,
-              site: model.id,
-              branch: model.defaultBranch
-          };
-          Build.create(build, done);
-        });
+    this.startInitialBuild(model, done)
```
jmhooper: I made this change so I could stub the `startInitialBuild` callback instead of trying to stub `Site`'s `afterCreate` action. That is because of the way Waterline manages lifecycle callbacks which makes it very difficult to stub this callback directly.

PR #593, opened 2016-11-04
Stop redirects on edit page if page is new, put page in pages

el-mapache:
* Don't attempt to redirect when user is creating a new page
* Put newly created pages in the `pages` directory
* Call `Boolean` to make code less obtuse


brandocalrissian:
Yes, this is easier to follow.

PR #603, opened 2016-11-07
Adds unit tests for add site component

el-mapache:
* Adds unit tests for AddSite component

### DIFF
--- a/test/client/app/components/addSite/addSite.test.js
+++ b/test/client/app/components/addSite/addSite.test.js
@@ -1,0 +1,114 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { expect } from 'chai';
+import { stub, spy } from 'sinon';
+import templates from '../../../../../config/templates';
+import proxyquire from 'proxyquire';
+
+proxyquire.noCallThru();
+
+const mock = () => {
+  return function() {
+    return <div></div>;
+  };
+};
+
+const TemplateSiteList = mock();
+const LinkButton = mock();
+const AlertBanner = mock();
+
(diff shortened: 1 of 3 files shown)


test/client/app/components/addSite/addSite.test.js
```diff
@@ -0,0 +1,114 @@
...
+    wrapper.instance().onSubmitTemplate(templateRepo, siteName);
+
+    expect(cloneRepo.calledWith({repo: templateRepo}, templates[siteName])).to.be.true;
+    expect(cloneRepo.called).to.be.true;
+  });
+
+  it('delivers the correct props to its children', () => {
+    const bannerProps = wrapper.find(AlertBanner).props();
+    const templateListProps = wrapper.find(TemplateSiteList).props();
+    const linkButtonProps = wrapper.find(LinkButton).props();
```
brandocalrissian: Is the intent to test if the right props are given to the link button? These props aren't validated in the test.

el-mapache: Oops, yes, I left that in by mistake. I was going to test that those props were being correctly passed, but they are hardcoded, not computed. That made me think it was an unnecessary test. Does that seem reasonable?

brandocalrissian: Yeah, testing configuration is not a very interesting test as you're simply copying the same strings from one place to another. I vote for taking out the line.

el-mapache: word, updated!

brandocalrissian:
Looks good! Merging...

Issue #566, opened 2016-10-26
Add loading screen for site that either may not exist, or is still propagating


el-mapache:
When you fork a repo using Federalist, there is a delay of ~ 2 minutes before the site can be fetched by Federalist and displayed to the user. We should create a page that conveys this information to the user instead of redirecting them to a blank page.

PR #604, opened 2016-11-08
Adds loading components, does not rely on config file

el-mapache:
Checking for the existence of a config file is too brittle, and doesn't allow users who have sites without one to browse their files on federalist. 

This PR instead checks for the existence of a site, then progressively tries to load data from github.

Related to issue #566 

* Check for existence of site before attempting to retrieve its assets from github
* No longer checks for the existence of a config.yml file, which may not be present on sites that are not federalist templates.
* Adds SiteInvalid and SiteLoading actions
* Adds an isLoading component and renames hasConfig
* Fixes height 100% on homepage


assets/app/routes.js
```diff
@@ -4,15 +4,15 @@ import { Route, Link, IndexRedirect, IndexRoute, Redirect } from 'react-router';
...
+import SitePagesContainer from './bundles/pagesContainerBundle';
 import SiteEditorContainer from './components/site/editor/editorContainer';
 import SiteLogs from './components/site/siteLogs';
 import SiteMediaContainer from './components/site/siteMediaContainer';
 import SiteSettings from './components/site/siteSettings';
 import NewSite from './components/AddSite';
 import NotFound from './components/NotFound';
 
-import alertActions from './actions/alertActions';
+import store from './store';
```
brandocalrissian: I don't see how this is used.

brandocalrissian:
Cool. Merging.

PR #607, opened 2016-11-08
Adds vuln tracking from Snyk as a badget to the README

NoahKunin:
Looks like there may be some patching needs on `master`.


wslack:
@NoahKunin we're retiring `master` - the version getting the new ATO is in `staging`. Will make new issues for the relevant findings from the correct scan.

PR #619, opened 2016-11-15
Build queue refactor

jmhooper:
This commit refactors some of the logic for the SQS service and the Build model.

This refactor was started in response to issues that arose while I was trying to update npm modules to remediate vulnerabilities. It looked like most of the problems were occurring around the code that communicates with SQS and within the lifecycle callbacks for the Build model.

This commit cleans up both of those and covers them with some comprehensive tests to help make issues more discoverable.


wslack:
@jmhooper THANK YOU! I can tell this was a lot of work.

Do we care about Travis being unhappy?

npm ERR! Failed at the federalist@0.0.0 test:server script 'NODE_ENV=test mocha --timeout 5000 test/api/bootstrap.test.js test/api/unit/**/*.test.js test/api/requests/**/*.test.js'.
npm ERR! This is most likely a problem with the federalist package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     NODE_ENV=test mocha --timeout 5000 test/api/bootstrap.test.js test/api/unit/**/*.test.js test/api/requests/**/*.test.js

jmhooper:
Didn't see that. I'll investigate.

jmhooper:
Okay, I just amended the commit. The failure was an artifact of me breaking the app to confirm the failure case resulted in a failed test.

jmhooper:
Note that this also resolves #586

api/services/SQS.js
```diff
@@ -1,82 +1,81 @@
...
+    return ""
   }
+}
 
-};
+var sourceForBuild = (build) => {
+  return build.source || {}
+}
+
+SQS.buildMessageBody = (build) => {
```
jmhooper: Maybe we should rename this `messageBodyForBuild`

api/models/Build.js
```diff
@@ -1,12 +1,58 @@
-/**
-* Build.js
-*
-* @description :: TODO: You might write a short summary of how this model works and what it represents here.
-* @docs        :: http://sailsjs.org/#!documentation/models
-*/
+var afterCreate = (model, done) => {
```
jmhooper: We should probably consider removing the messaging logic from this lifecycle callback and instead rely on a build inside of a controller action where builds are created.

api/services/SQS.js
brandocalrissian: well, if we move this (since it's really about what the build container wants to see and not how we use SQS), then everyone should be happier.

brandocalrissian:
Peachy. 🍑-y. Emoji puns?

Issue #621, opened 2016-11-15
Unable to create sites via UI in staging


jmhooper:
I just tried to create some sites with the latest on staging an encountered some errors.

One the new site page, entering a site name under a template and clicking "Create site" does not create a site. The following is logged in the console:

```
Uncaught TypeError: (0 , _siteActions.cloneRepo) is not a function(…)
```

One the new site page, entering the repository and owner for an existing site and clicking "Submit repository-base site" does not create a site. The following is logged in the console:

```
Uncaught TypeError: (0 , _siteActions.addSite) is not a function(...)
```

PR #622, opened 2016-11-15
Fix webhook issue that was preventing site creation

jmhooper:
This commit fixes an issue creating webhooks that was occurring when users attempted to add repository based sites. The issues occurred because `GitHub.setWebhook(site, user, done)` was being passed a user object instead of an integer for the user's id.

This commit fixes the issues and adds tests to verify this works as expected in the future.


jmhooper:
Note that this issue is also related to #621. Without these changes users cannot create repository based sites.

PR #623, opened 2016-11-15
Import siteActions correctly in AddSite component

jmhooper:
The siteActions were not being properly imported in the AddSite component which was causing #621.

This commit fixes the problem by correctly importing it.


brandocalrissian:
Yeah, we can't destructure our default export (yet? oh, standards how they sometimes fly!) as we're really importing a named export... which looks like destructuring...

Issue #609, opened 2016-11-08
DOS vulnerability in ws


jmhooper:
Snyk found a potential DOS vulnerability in our `ws` module which is part of socket.io.

ref: https://snyk.io/test/github/18f/federalist/staging
ref: https://snyk.io/vuln/npm:ws:20160624

jmhooper:
This is fixed even though snyk may still show the issue for a while.

jmhooper:
This issue is actually still present. Adding `ws` with a given version to the dependencies in `package.json` does not update `ws` in the modules in which it is a child dependency.

jmhooper:
Here is a breakdown of our issue:

We use `sails-db-migrate` to handle our migrations which supports `sails@0.x.x`, but not `sails@1.x.x` which is the latest version.

We are looking at a vulnerability in `ws@1.1.0` which we get through `sails@0.x.x`. The latest version of `sails@0.x.x` depends on `sails-hook-sockets` which is [dependent](https://github.com/balderdashy/sails-hook-sockets/blob/master/package.json#L17) on `socket.io@1.4.8` which is a vulnerable version of socket.io because it's dependency tree contains `ws@1.1.0`. There are more recent versions of `socket.io` that do not have this issue.

The latest version of sails, `1.0.0-11`, is not dependent on `sails-hook-sockets`, and does not have the vulnerability, but upgrading breaks our which means to upgrade we may need to find a new solution for migrations which is a tough spot.

jmhooper:
I opened balderdashy/sails-hook-sockets#31. Going to mark this "blocked" for now and watch for movement there.

PR #626, opened 2016-11-17
Remediate ws vulnerability by disabling websockets entirely

jmhooper:
This commit entirely disables the `sockets` and `pubsub` Sails hooks so the DOS vulnerability in the `ws` modules is never exposed. We don't actually use websockets at this point so there is no real point in keeping this code around.

This commit also cleans up some old code paths that leaned on web sockets.

ref: https://gist.github.com/mikermcneil/8911041
ref: #609


wslack:
@jmhooper Travis is unhappy? 

```Failed at the federalist@0.0.0 test:server script 'NODE_ENV=test mocha --timeout 5000 test/api/bootstrap.test.js test/api/unit/**/*.test.js test/api/requests/**/*.test.js'.```

jmhooper:
I believe that is unrelated to this change. I opened #627 to deal with it.

brandocalrissian:
This PR makes sense. Might as well wait until #627 is fixed and this build goes green before merging. Would feel completely confident once we no longer have to say we "believe" we know the travis failure is unrelated to the change.

jmhooper:
This is rebased with the commit from #627 merged. Will wait on Travis then merge.

wslack:
@jmhooper needs to rebase!

brandocalrissian:
staging moved after the last rebase to cause this to get back out of sync

brandocalrissian:
good to go.

PR #624, opened 2016-11-16
Upgrade sails, grunt, and socket.io-redis

jmhooper:
This commit updates the following packages to remediate vulneratbilites:

- sails@0.11.0 -> sails@0.12.9
  - [handlebars@3.0.3 -> handlebars@4.0.6](https://snyk.io/vuln/npm:handlebars:20151207)
  - [hawk@1.1.1 -> hawk@3.1.3](https://snyk.io/vuln/npm:hawk:20160119)
  - [minimatch@0.2.14 -> minimatch@3.0.3](https://snyk.io/vuln/npm:minimatch:20160620)
  - [negotiator@0.5.3 -> skipper@0.6.2](https://snyk.io/vuln/npm:negotiator:20160616)
  - [request@2.40.0 -> request@2.68.0](https://snyk.io/vuln/npm:request:20160119)
  - [uglify-js@2.3.6 -> uglify-js@2.6.0](https://snyk.io/vuln/npm:uglify-js:20150824)
- grunt@0.4.2 -> grunt@1.0.1
  - [minimatch@0.2.14 -> minimatch@3.0.3](https://snyk.io/vuln/npm:minimatch:20160620)
- grunt-cli@0.1.13 -> grunt-cli@1.2.0
  - [minimatch@0.3.0 -> minimatch@3.0.2](https://snyk.io/vuln/npm:minimatch:20160620)
- socket.io-redis@0.1.4 -> socket.io-redis@1.0.0
  - [ms@0.6.2 -> ms@0.7.1](https://snyk.io/vuln/npm:ms:20151024)

This commit also bumps the node and npm versions which makes `npm install` run a bit snappier and will let us use some of the more modern features available in node v6.

ref #611
ref #608

PR #628, opened 2016-11-17
Update buildpack

jmhooper:
We updated to node 6.9.0 and npm 3.10.8 in #624. Our manifest was hardcoded to use the nodejs buildpack v1.5.4 which does not support those versions of node / npm. This commit updates the manifest to use buildpack v1.5.23 which does support node 6.9.0 and npm 3.10.8.


wslack:
@jmhooper are we expecting this to fail in Travis (and so should we just immediately merge?)

jmhooper:
Yes, this will auto fail b/c of the webhook API spec. It also will not deploy if the spec fails so the only way to get it to deploy from Travis would be to bypass that spec.

jmhooper:
Okay, this is working on staging now and should be ready to go. Now the changes include:

- Update buildpack
- Update travis config to use node v6
- Delete the staging environment config (we use the production config on staging; see the commit message for details)
- Restructure the production config to be a bit more readable
- Upgrade some node modules so everybody gets along

brandocalrissian:
if the tires have been kicked with these changes on staging, then we should be good to go once the rebase/conflict resolution is complete.

brandocalrissian:
meaning it's good for merging once we make sure the conflicts are nothing to worry about.

jmhooper:
Rebased and waiting.

brandocalrissian:
we're green. merging.

PR #630, opened 2016-11-17
Adds landing page to the templates

wslack:
Questions:
- [x] Are there any other places this needs to be added?
- [x] Where is the image for the thumbnail stored? (Was added [here](https://github.com/18F/federalist/commit/d948f150b8f212483b12dd75bdf59642ff1d4257))
- [x] Is there any needed refactoring to make adding other templates easier in the future?


jmhooper:
@wslack: I moved the commit with the thumb to this branch (you originally pushed it to master). Then I updated the `SiteTemplate` component to take up 1/4 of the screen instead of 1/3 so the templates still layout in a row.

If both of those changes look good to you then this can probably be merged.

jmhooper:
To actually answer your questions:

>  Are there any other places this needs to be added?

No, just in the config

>  Where is the image for the thumbnail stored? (Was added here)

You added it in the right place, just on the wrong branch

>  Is there any needed refactoring to make adding other templates easier in the future?

Seems straightforward to me, but if there was somewhere you saw room for improvement we can make improvements.

wslack:
@jmhooper should we merge despite Travis?

jmhooper:
If there's no hurry-up on this we can wait for #631 which makes that failing spec green again.

jmhooper:
Rebased this for you @wslack.

Issue #627, opened 2016-11-17
Failed webhook request spec


jmhooper:
We have a failing webhook request spec on staging:

```
    POST /webhook/github
Unhandled rejection AssertionError: expected [ Array(1) ] to have a length of 2 but got 1
    at factory.then.then.then.then.then.builds (/Users/jonathanhooper/Developer/federalist/web/test/api/requests/webhook.test.js:63:32)
    at tryCatcher (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/promise.js:510:31)
    at Promise._settlePromise (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/promise.js:567:18)
    at Promise._settlePromise0 (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/promise.js:612:10)
    at Promise._settlePromises (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/promise.js:691:18)
    at Async._drainQueue (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/async.js:138:16)
    at Async._drainQueues (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/async.js:148:10)
    at Immediate.Async.drainQueues (/Users/jonathanhooper/Developer/federalist/web/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:637:20)
    at tryOnImmediate (timers.js:610:5)
    at processImmediate [as _immediateCallback] (timers.js:582:5)
```

The spec looks like this:

```javascript
it("should create a new site build for the sender", done => {
  var site, user

  factory(Site).then(site => {
    return Site.findOne({ id: site.id }).populate("users")
  }).then(model => {
    site = model
    user = site.users[0]
    return Build.find({ site: site.id, user: user.id })
  }).then(builds => {
    expect(builds).to.have.length(1)

    var payload = buildWebhookPayload(user, site)
    signature = signWebhookPayload(payload)

    return request("http://localhost:1337")
      .post("/webhook/github")
      .send(payload)
      .set({
        "X-GitHub-Event": "push",
        "X-Hub-Signature": signature,
        "X-GitHub-Delivery": "123abc"
      })
      .expect(200)
[...]

PR #631, opened 2016-11-18
Refactor webhook controller

jmhooper:
This commit refactors the code in the WebhookController to make a couple of changes:

- Respond with an error if the user that triggered the webhook does not exist in Federalist
- Respond with an error if the site the webhook was triggered for does not exist in Federalist
- Don't respond with a 200 OK until a build has been created
- Break out some of the logic and refactor the code to make it more readable

ref #627

### DIFF
--- a/api/controllers/WebhookController.js
+++ b/api/controllers/WebhookController.js
@@ -1,78 +1,81 @@
-// Validation from https://github.com/rvagg/github-webhook-handler
-// Copyright (c) 2014 Rod Vagg, MIT License
-
-var crypto = require('crypto');
+var crypto = require('crypto')
 
 module.exports = {
-
   github: function(req, res) {
-    var sig = req.headers['x-hub-signature'],
-        event = req.headers['x-github-event'],
-        id = req.headers['x-github-delivery'],
-        secret = sails.config.webhook.secret,
-        payload = req.body;
-
-    // Ignore non-push events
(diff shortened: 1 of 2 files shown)


wslack:
> Respond with an error if the user that triggered the webhook does not exist in Federalist

Do we want this? A Federalist rebuild shouldn't require a Federalist user account (just need someone on the team with a Federalist account who can manage the Federalist settings)

jmhooper:
Let me check the docs on what user is the `sender` for a webhook and I'll get back to this. What I'm trying to avoid is creating new Federalist users who maybe should not have accounts. That's what we did before.

wslack:
100% - users should only exist if auth'd by the federalist-users or 18F github orgs

jmhooper:
Okay, checked the GitHub webhook docs and it looks like the `sender` field here refers to the user who runs `git push`. That means this change would cause the webhook endpoint to 400 on legitimate requests. Good catch @wslack.

jmhooper:
I pushed a change that goes back to the way it used to be where a user is created for the sender. This user is not able to authenticate though. They just have a record in the DB as someone who has triggered builds.

wslack:
@brandocalrissian this is needed to fix Travis! Can you check it?

brandocalrissian:
looks good to me once the merge conflicts are resolved. Woo hoo rebase!

PR #632, opened 2016-11-18
add date formatting

wslack:


assets/app/components/siteList/publishedState.js
```diff
@@ -12,7 +13,7 @@ const getLastBuildTime = (builds) => {
   });
   let last = sorted.pop();
 
-  return last.completedAt || 'forever ago';
+  return moment().format(last.completedAt)) || 'forever ago';
```
jmhooper: This needs to looks like this:

```javascript
moment(last.completedAt).format("will slack's formatting string")
```

brandocalrissian:
it looks like the tests are failing because their expected outcomes changed.

wslack:
YAY!

Issue #625, opened 2016-11-16
Merge Passport and User model


jmhooper:
As it stands, a user only ever has one Passport record. In fact, there are several potential error states that arise if a user has more than one Passport. This makes the relationship an obvious candidate for a refactor.

We can simplify a lot of the logic in the backend by merging the Passport model into the User model. This would carryover into some less dramatic changes in the frontend that would mostly increase readability.

The attributes we'd need to merge over to the User model would be

- githubUserID <currently Passport.identifier>
- githubAccessToken <currently Passport.tokens.accessToken>

The rest of the attributes on the Passport model have values that are hardcoded into the application code where they're set and would not need to be migrated.

wslack:
@jmhooper how much time do you think this will take?

jmhooper:
Tough to call because it involves writing a migration which I've never done with Sails, but should be PR ready within 4 hours.

PR #637, opened 2016-11-23
Merge Passport and User model

jmhooper:
This commit gets rid of the Passport model in favor of saving that data on the User model. This decision was made because the User model effectively has a `hasOne` relationship to the Passport model. This lead to lots of complicated and confusing code around managing that relationship.

This commit removes that model entirely by making the following changes:

1. Write a migration to move data from the `passport` table into the `user` table.
1. Remove Passport model code from the API
1. Remove Passport model code from the frontend
1. Change the `/v0/user` route to `/v0/me` to expose GitHub user data to the current user only.

ref #625


brandocalrissian:
much clearer, thank you.

PR #640, opened 2016-11-23
Use recursive promises in drop-passport migration

jmhooper:
The drop-passport migration was not working because it was trying to make several simultaneous writes to the user table. These writes were locking the database causing the migration to block.

This commit works around this problem by using recursive promises to make sure that writes happen serially. This means that the write will lock, perform it's actions, then unlock before the next write occurs.


jmhooper:
We should avoid merging until we get a review from @rogeruiz

PR #641, opened 2016-11-23
Consolidate AWS configs

jmhooper:
This commit consolidates the configs for AWS into 2 places depending on the service:

- SQS: `sails.config.sqs`
- S3:  `sails.config.s3`

This commit then updates the code so that these configs are used instead of the configs that were spread out all of the place before. This commit also drops some dead code I found while making these changes.


brandocalrissian:
looks good!

Issue #639, opened 2016-11-23
"<PublishedState /> displays the datetime of the most recent build" fails because of timezones


jmhooper:
This spec passes on CI because CI uses UTC, but locally where there's timezones it fails because the hour is out of sync:

```
  1) <PublishedState /> displays the datetime of the most recent build:

      AssertionError: expected 'This site was last published at September 4th 2015, 10:11:23 am.' to equal 'This site was last published at September 4th 2015, 3:11:23 pm.'
      + expected - actual

      -This site was last published at September 4th 2015, 10:11:23 am.
      +This site was last published at September 4th 2015, 3:11:23 pm.
      
      at Context.<anonymous> (test/client/app/components/siteList/publishedState.test.js:38:41)
```

The string that is being passed into the component is `2015-09-04T15:11:23.000Z` and this works on Travis since that is 3:11 UTC, but fails locally for me since that is 10:11 CT.

PR #642, opened 2016-11-23
Use moment to format datetime in publishedState component test

jmhooper:
This commit updates the publishedState component's test to use moment to format the date string it expects the component to render. This is done to make sure the date string is displayed in the current timezone and not in UTC which is what it was hardcoded to do.

With this change the test should pass regardless of the timezone it is run in.

### DIFF
--- a/test/client/app/components/siteList/publishedState.test.js
+++ b/test/client/app/components/siteList/publishedState.test.js
@@ -1,18 +1,21 @@
 import React from 'react';
+import moment from 'moment'
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
 import PublishedState from '../../../../../assets/app/components/siteList/publishedState';
 
 const NO_BUILDS_MESSAGE = 'Please wait for build to complete or check logs for error message.';
 const PUBLISHED_BASE = 'Please wait for build to complete or check logs for error message.';
-const MOST_RECENT_BUILD = 'This site was last published at September 4th 2015, 3:11:23 pm.';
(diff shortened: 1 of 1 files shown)


jmhooper:
The issue this address is #639

test/client/app/components/siteList/publishedState.test.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
+import moment from 'moment'
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
 import PublishedState from '../../../../../assets/app/components/siteList/publishedState';
 
 const NO_BUILDS_MESSAGE = 'Please wait for build to complete or check logs for error message.';
 const PUBLISHED_BASE = 'Please wait for build to complete or check logs for error message.';
-const MOST_RECENT_BUILD = 'This site was last published at September 4th 2015, 3:11:23 pm.';
+const FORMATTED_MOST_RECENT_BUILD_TIME =  moment("2015-09-04T15:11:23.000Z").format('MMMM Do YYYY, h:mm:ss a')
```
brandocalrissian: can you extract this string representing the build time as a constant to make it absolutely clear that this value matches the value a few lines later?

jmhooper: You mean the string `"2015-09-04T15:11:23.000Z"`?

brandocalrissian: Precisely!

jmhooper:
Pushed an amended commit

wslack:
this LGTM but will defer to @brandocalrissian to merge

Issue #597, opened 2016-11-04
Consider removing `populate` action from User API


jmhooper:
The User API currently allows the `populate` action provides by sails which creates routes like:

```
GET/POST             /v0/user/:id/site
GET/PUT/DELETE /v0/user/:id/site/:id
GET/POST             /v0/user/:id/build
GET/PUT/DELETE /v0/user/:id/build/:id
```

We get that same behavior from the Site API and the Build API since they apply filter policies that filters the responses by ones associated with the current user.

It appears that we prefer the actions provided the Build and Site API over the populated User API responses. If we aren't going to use these API actions, and we have no plans to use them in the future, they should probably be removed.

PR #643, opened 2016-11-23
Remove unused routes from policies

jmhooper:
This commit removes routes from the policies that aren't used. These include:

- Site.fork
- Site.lock
- Site.unlock
- User.populate

ref #597


brandocalrissian:
groovy.

PR #646, opened 2016-11-25
Fix GitHub service

jmhooper:
The `setWebhook` method in the GitHub service was anticipating a user object, but was receiving a user ID. This was causing a problem setting the webhook since `userId.githubAccessToken` is obvious undefined.

This commit loads the user from the user ID before attempting to set the webhook.


brandocalrissian:
good to go

Issue #644, opened 2016-11-25
Clicking "logs" when in logs page flips order of table


wslack:
https://github.com/18F/federalist/pull/613 may have introduced this issue. The logs table shouldn't flip.

PR #647, opened 2016-11-25
Sort site builds array in build log

jmhooper:
The build log was reversing the site's array of builds in place. This meant every time `render` was called on the component it would reverse the order of the list of builds.

This commit fixes the issue by using the `sort` method to make sure that:

1. A copy of the build array is modified
2. The build array is always sorted correctly regardless of how the API renders it (since Sails does not promise to render a sorted array)

ref #644


brandocalrissian:
cool

PR #649, opened 2016-11-28
Give GitHub OAuth token write:repo_hooks scope

jmhooper:
This commit adds a scope to the GitHub OAuth token so that it can create webhooks on repositories.


brandocalrissian:
we can roll this back if we validate this does not fix the problem in staging.

PR #651, opened 2016-11-28
Don't stub `GitHub.registerSite` in integration tests

jmhooper:
One of the weaknesses of our integration tests was the fact that they did not cover the GitHub service.
In the integration tests we stubbed the GitHub service's API to prevent it from sending HTTP requests to GitHub which would 404.

This commit fixes that by adding a request mock for the Webhook API endpoint and using that in the specs.

This needs to be used in the factories when one-off sites are created, and in tests where we want to verify that a webhook is created successfully.
It also needs to be used in requests that create Sites.

### DIFF
--- a/api/models/Site.js
+++ b/api/models/Site.js
@@ -65,9 +65,10 @@ module.exports = {
   },
 
   registerSite: function(values, done) {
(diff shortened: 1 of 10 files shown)


brandocalrissian:
got it.

Issue #359, opened 2016-05-25
Improve Dev Onboarding


wslack:
cc @jeremiak and @el-mapache 

Capturing Questions
1. What environment variables do I need for Docker builder?
2. How does the staging environment?

Anything preceded by a -E flag is an environment variable. You will also need your Github OAuth token for access to the repo in question.
1.

jmhooper:
Curious about what else we could add to this. Right now we have:

- [Running Federalist Locally](https://federalist-docs.18f.gov/pages/how-federalist-works/running-federalist-locally/)
- [Deploying Federalist](https://federalist-docs.18f.gov/pages/how-federalist-works/deploying-federalist/)
- [cloud.gov Setup](https://federalist-docs.18f.gov/pages/how-federalist-works/cloud-gov-setup/)

In addition, the [How Federalist Works](https://federalist-docs.18f.gov/pages/how-federalist-works/) doc has been significantly cleaned up. I've also made sure the README's for each of our repositories are up to date.

Anything else we want to add before closing this?

jmhooper:
At the very least, I've gone through and added documentation to answer the questions that I had when starting on the project that haven't been resolved by cleaning up the codebase.

wslack:
I'm happy to put it in Done? pending retro convo today.

PR #653, opened 2016-11-30
Update README to reflect how Federalist actually works

jmhooper:
This commit makes some changes to the README with developer onboarding in mind.

For the most part it changes the setup instructions to account for some recent changes to how the environment is configured.

ref #359

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,51 +9,56 @@ Federalist is a unified interface for publishing static government websites. It 
 
 ## Getting started
 
-To run the server, you'll need [Node.js](https://nodejs.org/download/) and [Ruby](https://www.ruby-lang.org/en/documentation/installation/) installed. The setup process will automatically install Jekyll and its dependencies based on the `github-pages` gem.
-You'll also need [nvm](https://github.com/creationix/nvm).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,22 +122,50 @@ organizations: [
...
+
+In development, Federalist uses the staging S3 bucket.
+To connect to the bucket, get the credentials for the staging bucket and add them to the S3 config in your `config/local.js`.
+
+```
+s3: {
+  accessKeyId: "ABC123",
+  secretAccessKey: "456def",
+  region: "us-gov-west-1",
+  bucket: "789ghi",
```
wslack: Can we mark these as fake in a comment just to avoid confusion?

jmhooper: I'm not sure about a comment. I'd prefer to rework the statement above to provide clarity:

> To connect to the bucket, get the credentials for the staging bucket and add them to the S3 config in your `config/local.js`.

jmhooper: Okay, I added some sentences above the examples to clarify that they are example values only and a note about where to get the actual values.

PR #656, opened 2016-11-30
Update .nvmrc to reflect current node version

jmhooper:
The .nvmrc fell behind the actual node version.
This commit updates it to 6 which is the version we currently use.


wslack:
This PR apparently broke our cloud.gov deploy and has been reverted for now in `staging`.

PR #657, opened 2016-11-30
Refactor GitHub service

jmhooper:
This commit refactors the GitHub service to return Promises instead of using callbacks. This makes the asynchronous logic easier to follow.

This commit also updates the places where the GitHub service is used, in some cases pulling logic that is a concern of the service out of other parts of the app and into the service, and also removing the callback logic.

### DIFF
--- a/api/services/GitHub.js
+++ b/api/services/GitHub.js
@@ -72,7 +72,22 @@ const handleWebhookError = (error) => {
 }
 
 module.exports = {
-  setWebhook: function(site, user) {
+  checkPermissions: (user, owner, repository) => {
+    return githubClient(user.githubAccessToken).then(github => {
+      return getRepository(github, {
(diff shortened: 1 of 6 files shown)


jmhooper:
Let's actually move this to WIP while I work on some tests for the GitHub service.

jmhooper:
Added the tests. This is no longer WIP.

api/services/GitHub.js
```diff
@@ -1,27 +1,95 @@
-/*
- * Service for interfacing with GitHub
- */
-var url = require('url'),
-    Client = require('github'),
-    github = new Client({ version: '3.0.0' });
+var Github = require("github")
+var github = new Github({ version: "3.0.0" })
```
jmhooper: Amended the commit and its gone.

PR #658, opened 2016-11-30
Add "Restart" button the build logs

jmhooper:
This PR adds a restart button to the build logs. This button communicates with the API via a restart endpoint that was also added in the PR. This endpoint creates a new build with the same site, user, and branch as the previous. It then renders this build back to the frontend where it is appended to the build log.

### DIFF
--- a/test/api/requests/build.test.js
+++ b/test/api/requests/build.test.js
@@ -16,10 +16,9 @@ describe("Build API", () => {
   var buildResponseExpectations = (response, build) => {
     Object.keys(build).forEach(key => {
       expect(response[key]).to.not.be.undefined
-
       if (typeof build[key] === "number" && typeof response[key] === "object") {
         expect(response[key].id).to.equal(build[key])
-      } else if (build[key].toISOString) {
+      } else if (build[key] && build[key].toISOString) {
         expect(response[key]).to.equal(build[key].toISOString())
       } else {
         expect(response[key]).to.equal(build[key])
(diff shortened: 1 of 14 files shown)


jmhooper:
Screenshot:

image

test/api/requests/build.test.js
```diff
@@ -146,6 +145,102 @@ describe("Build API", () => {
...
+          user: build.user,
+          site: build.site,
+        }).sort("createdAt DESC")
+      }).then(builds => {
+        buildResponseExpectations(response.body, builds[0])
+        done()
+      })
+    })
+
+    it("should respond with a 403 if the original build is not associated with the user", done => {
```
wslack: Is this saying that only the original user can trigger a rebuild?

jmhooper: Only the user that created the build. Looking at it now, perhaps it should change to a user associated with the site's build.

wslack: I think it's misleading if users can all see buttons to do rebuilds that they themselves can't trigger.

jmhooper: Yeah, I agree. Cooking up that change right now.

wslack: Shouldn't any user with the ability to see a site in Federalist (with write access) be able to trigger a build?

jmhooper: Yep ^

jmhooper: Amended the commit that adds the API endpoint so that now it has this behavior.

Good catch

brandocalrissian:
Looks good. Very clear. Verified code coverage, too.

PR #662, opened 2016-12-01
Bump nvmrc back up to node v6

jmhooper:
Revert the reversion


wslack:
verified that there's no venerable app, merging.

PR #664, opened 2016-12-02
Fix option to cf login

jmhooper:
When we updated to cf-cli 6.22.2 we did not update the syntax for cf login which changed from:

```
cf login --u username --p password
```

to:

```
cf login -u username -p password
```


wslack:
we could have a staging-for-staging version deployed but that feels redundant..... :P

jmhooper:
Before I open PRs for these I short circuit the Travis config so I can verify that it works: https://travis-ci.org/18F/federalist/builds/180807264

Then I drop the commit that changes the config and open a PR.

Issue #665, opened 2016-12-05
Cloned sites cancel and start another build


jmhooper:
When a site is created from a template (endpoint `/v0/sites/clone`, the build that is created with the site is cancelled, and a new one that pushes the new site to GitHub is created.

We should come up with a more straightforward way to manage builds so that we don't have to start and then immediately cancel a build.

This work should also consider that fact that there are 2 endpoints that create a site: `/v0/sites/clone` and `/v0/users/add-site`. These could likely be combined into a single site creation endpoint.

PR #667, opened 2016-12-06
Refactor site creation code

jmhooper:
This PR refactors the API endpoints that are used to create sites.

Prior to this commit, sites were created using `/v0/site/clone` to create a site from a template, and `/v0/user/add-site` to create a site with an existing repo on GitHub.
This commit changes the API so that both actions can be performed using `/v0/site`.

To create a site from a template, a post request can be sent to `/v0/site` with the following payload:

```json
{
  "template": "team",
  "owner": "siteOwner",
  "repository": "siteRepository",
  "defaultBranch": "master",
  "engine": "jekyll"
}
```

The template will be cloned pushed to the new repo.
The site will then be built.

To create a site from an existing GitHub repo, a post request can be sent to `/v0/site` with the following payload:

```json
{
  "owner": "siteOwner",
  "repository": "siteRepository",
  "defaultBranch": "master",
  "engine": "jekyll"
}
```

The site will be cloned from the provided repo and then the site will be built.

Ref #665

At this time, this PR covers the API, but work still needs to be done for the frontend to be able to handle these changes.


jmhooper:
Pushed a commit that integrates this with the frontend. This is no longer WIP and is ready for review.

api/controllers/SiteController.js
```diff
@@ -1,49 +1,139 @@
...
+      })
+
+      if (existingSite) {
+        return existingSite
+      } else {
+        return createAndBuildSite(req)
+      }
+    }).then(model => {
+      site = model
+      site.users.add(req.user.id)
```
brandocalrissian: Would it make more sense to move this "update the site to add the current user as a user" logic to `createAndBuildSite`-- then we would also seem to be able to remove the following `then` clause as `createAndBuildSite` can return the same values we'd get in the next call to `Site.findOne(site.id).populate("users").populate("builds")`

api/controllers/SiteController.js
```diff
@@ -1,49 +1,139 @@
...
+      return site.save()
+    }).then(result => {
+      return Site.findOne(site.id).populate("users").populate("builds")
+    }).then(site => {
+      res.send(site)
+    }).catch(err => {
+      if (err.code === "400") {
```
brandocalrissian: Can we extract these common lines of error handling code? There's a duplicate set below (will flag).

api/controllers/SiteController.js
```diff
@@ -1,49 +1,139 @@
...
-        });
-      });
-    });
+      })
+    }).then(build => {
+      return Site.findOne(siteId).populate("users").populate("builds")
+    }).then(site => {
+      res.send(site)
+    }).catch(err => {
+      if (err.code === "400") {
```
brandocalrissian: This is where the duplication is mentioned in line 109 above.

api/controllers/SiteController.js
jmhooper: The issue with that is `createAndBuildSite` is only called if the site does not already exist. That is in the block above:

```javascript
if (existingSite) {
  return existingSite
} else {
  return createAndBuildSite(req)
}
```

If this logic is moved into `createAndBuildSite`, then a user who attempts to add an existing site will never be added to the site.

The way I was trying to make this read is, "find or create a new site, then add the user to that site, then..."

jmhooper:
Went through and addressed everything. This should be ready for another look over.

api/controllers/SiteController.js
brandocalrissian: Ah, yes. Still feels like we could make this better. Since I don't have a constructive suggestion, I'll stop my whining.

PR #670, opened 2016-12-07
Fix SiteContainer so that build logs and branches appear correctly

jmhooper:
This commit fixes 2 problems involving the site container. These
problems worked together to make the site data as it was displayed in
the container look incorrect.

The first issue was in the site container's code itself. The container
was importing an action creator and not an action to replace the router
history. This meant that if the container loaded without site data
instead of re-directing to the site list as it was intended, it rendered
its children without data fetched from GitHub.

The second issue was deep in the PublishedDate component which was
calling `pop()` on the build array for each site. This was removing the
most recent build from each site so the build logs were rendering
without that build.

### DIFF
--- a/assets/app/components/siteContainer.js
+++ b/assets/app/components/siteContainer.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Link } from 'react-router';
 
-import { replaceRouterHistory } from '../actions/actionCreators/navigationActions';
+import { replaceHistory } from '../actions/routeActions';
 import siteActions from '../actions/siteActions';
 
 import SideNav from './site/SideNav/sideNav';
@@ -27,7 +27,7 @@ class SiteContainer extends React.Component {
     if (currentSite) {
       siteActions.fetchSiteConfigsAndAssets(currentSite);
     } else {
-      replaceRouterHistory('/sites');
+      replaceHistory('/sites');
     }
   }
 
--- a/assets/app/components/siteList/publishedState.js
+++ b/assets/app/components/siteList/publishedState.js
@@ -12,7 +12,7 @@ const getLastBuildTime = (builds) => {
       let bCompletedAt = new Date(b.completedAt);
       return aCompletedAt > bCompletedAt;
     });
-    let last = sorted.pop();
+    let last = sorted.slice().pop();
     return last.completedAt
   }
 };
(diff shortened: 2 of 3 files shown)


brandocalrissian:
IN-ter-es-ting... will review

assets/app/components/siteList/publishedState.js
```diff
@@ -12,7 +12,7 @@ const getLastBuildTime = (builds) => {
       let bCompletedAt = new Date(b.completedAt);
       return aCompletedAt > bCompletedAt;
     });
-    let last = sorted.pop();
+    let last = sorted.slice().pop();
```
brandocalrissian: This has changed the functionality but the PublishedState test has not been changed. This seems to indicate our test for PublishedState wasn't as thorough as we wanted-- it looks like our test happened to only test if the last build was represented in the render and this hid the bug you found. Can we please add another test to exercise this new functionality to verify we don't happen to run into it again and further document the code with tests?

assets/app/components/siteContainer.js
```diff
@@ -27,7 +27,7 @@ class SiteContainer extends React.Component {
     if (currentSite) {
       siteActions.fetchSiteConfigsAndAssets(currentSite);
     } else {
-      replaceRouterHistory('/sites');
+      replaceHistory('/sites');
```
brandocalrissian: good catch! some day we should get some test coverage on this component. it's never surprising to find bugs in code that's missing tests.

jmhooper:
I added tests.

I'm a bit concerned that this kind of test won't address the deeper problem here. A test to verify that the PublishedDate component doesn't mutate it's data will prevent this problem in this one part of the code, but we can still mutate the state from within any other component.

[Immutable data management is a prerequisite for using redux](http://redux.js.org/docs/recipes/reducers/PrerequisiteConcepts.html#immutable-data-management). While we COULD add tests to all of our components to verify that they aren't mutating their data, that's a lot of overhead in our tests. If we're serious about catching / preventing problems where the state is mutated by reducers or components, my suggestion would be to use something like [immutable.js](https://facebook.github.io/immutable-js/) for our state.

brandocalrissian:
we shouldn't make tests to verify we aren't modifying state we've received. that's waaaaaaay too much trouble and introduces more manual pain, something we'll likely mess up in the future.

it would be better to use immutable.js to enforce this behavior. I've enforced this in the past through convention (don't modify state. ever. i mean it. stop it. seriously. it's not funny.), but it might not be enough in this case.

thank you for the test mod

Issue #671, opened 2016-12-07
Site container redirects to site list when refreshed


jmhooper:
When the site container is refreshed, it redirects to the site list. This is because it is rendered before the `siteActions.fetchSites` action resolves triggering the `SITES_RECEIVED` action. Because of that it is rendered without site data and redirects to `/sites`.

We should build out the `SiteContainer` component so that it waits for site data before deciding whether it should render data or redirect.

wslack:
@jmhooper I am having a hard time following the impact to the user - what is the negative outcome in the status quo?

jmhooper:
If you got to a site and select any of the links in the sidebar then refresh, instead of refreshing the page it redirects you to the site list. So you refresh and then wind up on a totally different page.

This did not happen in the past because of a bug, the fix for which was recently merged.

wslack:
@jmhooper gotta love bug fixes that unleash other issues. Thanks!

PR #674, opened 2016-12-08
Prevent site container from redirecting if site data is not present

jmhooper:
The SiteContainer component would redirect to the SiteList if a user
refreshed while viewing it. This was because the container was mounted
without any site data, causing the container to think a user was
attempting to look at a site which was not in their site list.

This commit fixes that by checking that site data exists before
attempting to set a current site. This means that the container will
wait until data is present before redirecting.

The downside of this solution is that a user without any sites will not
see a redirect since they do not have site data to begin with.

In the future, I think a better solution would be to change the state
such that it is aware of whether or not site data has been fetched. That
way instead of checking the `storeState.sites` array, the SiteContainer
could check that sites have been downloaded directly and then know
whether to proceed with rendering site content. That change will require
some big architectual change though.

### DIFF
--- a/assets/app/components/siteContainer.js
+++ b/assets/app/components/siteContainer.js
@@ -21,7 +21,23 @@ class SiteContainer extends React.Component {
   }
 
   componentDidMount() {
-    const { storeState, params, routeParams } = this.props;
+    const { storeState } = this.props;
+    if (storeState.sites.length) {
+      this.downloadCurrentSiteData();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { storeState } = this.props;
+    const prevStoreState = prevProps.storeState;
+
+    if (storeState.sites.length && !prevStoreState.sites.length) {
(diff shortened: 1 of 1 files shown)


jmhooper:
This is in regards to #671 btw

assets/app/components/siteContainer.js
```diff
@@ -21,7 +21,23 @@ class SiteContainer extends React.Component {
...
   componentDidMount() {
-    const { storeState, params, routeParams } = this.props;
+    const { storeState } = this.props;
+    if (storeState.sites.length) {
+      this.downloadCurrentSiteData();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { storeState} = this.props;
```
brandocalrissian: nitpick -- missing space before brace in `{ storeState}`

brandocalrissian: This is my only comment. everything else looks good since we don't have any coverage on this pesky component yet.

PR #676, opened 2016-12-09
Cleanup policies

jmhooper:
This PR will cleanup our policy code which is kind of hard to understand / follow at this point. I also plan to add tests for the policies in this PR.

Right now I've put one commit up here which removes some policies.


jmhooper:
This should be ready to go.

This PR turned out to be pretty busy. Among the changes you'll see here:

- Remove policies that are used in just one controller action and put that logic in the controller
- Refactor the filterCurrentUser policy
- Remove the logic pertaining to preview URLs in the policies and move it to a service class. That commit also cleans up the logic in the preview controller which fits in that aforementioned service class.

api/policies/filterCurrentUser.js
```diff
@@ -1,56 +1,57 @@
...
+      })
+
+      if (currentUserIndex >= 0) {
+        next()
+      } else {
+        res.forbidden("Forbidden")
+      }
+    }).catch(err => res.forbidden("Forbidden"))
+  } else {
+    res.badRequest()
```
brandocalrissian: Nifty. We should catch these errors when they're thrown in development in the normal scheme of things, like an assertion.

Issue #584, opened 2016-11-01
Rename build container environment's PREFIX variable


jmhooper:
The `PREFIX` variable is sent in the build container environment to a build container which causes issues with NVM. Details about this problem are laid out in #570

Per my comments there, we should consider changing the name of the `PREFIX` container env variable to something like `SITE_PREFIX` in the web app and in the build container.

Here's where it is set in the web app: https://github.com/18F/federalist/blob/c0302467f5e63b4c5dd70f2c603bc4cb1d42da00/api/services/SQS.js#L49

wslack:
this is holding for our retirement of E/W

wslack:
Deployed in staging!

PR #680, opened 2016-12-13
Rename PREFIX to SITE_PREFIX in build message

jmhooper:
This commit changes the PREFIX variable in the build message to
SITE_PREFIX. This is in response to the change in
18F/federalist-docker-build#23 in which the build container begins
expecting SITE_PREFIX instead of PREFIX.

This change is necessary because using PREFIX in the environment results
in a collision with npm.

ref #584
ref 18f/federalist-docker-build#23


jmhooper:
This should be merged and deployed alongside 18F/federalist-docker-build#23

PR #682, opened 2016-12-19
Build container logging

jmhooper:
This PR prepares the web application to receive updates about the build from the build container and display them to the user.

This PR is WIP. At the moment, it extends the API such that build containers can communicate output to the the backend, and the frontend can fetch the build logs from the backend. This PR will be ready when the frontend is able to display these records to the user.

### DIFF
--- a/api/policies/buildCallback.js
+++ b/api/policies/buildCallback.js
@@ -3,7 +3,7 @@
  * Requests must have a `token` param that matches config and an `id`
  */
 module.exports = function (req, res, next) {
-  if (req.param('token') !== sails.config.build.token || !req.param('id')) {
+  if (req.param('token') !== sails.config.build.token) {
     return res.badRequest();
   }
   next();
--- a/api/policies/buildCallback.js
+++ b/api/policies/buildCallback.js
@@ -1,7 +1,3 @@
-/**
- * Build Callback Policy
- * Requests must have a `token` param that matches config and an `id`
- */
 module.exports = function (req, res, next) {
   if (req.param('token') !== sails.config.build.token) {
     return res.badRequest();
--- a/assets/app/components/site/siteBuildLogs.js
+++ b/assets/app/components/site/siteBuildLogs.js
@@ -1,0 +1,65 @@
+import React from "react";
+import buildLogActions from "../../actions/buildLogActions";
+
+class SiteBuildLogs extends React.Component {
+  componentDidMount() {
+    buildLogActions.fetchBuildLogs({ id: this.props.params.buildId });
+  }
+
+  buildLogs() {
+    if (!this.props.buildLogs) {
+      return [];
+    }
+
+    return this.props.buildLogs.filter(log => {
+      return log.build === parseInt(this.props.params.buildId);
+    }).sort((a, b) => {
+      return new Date(a.createdAt) - new Date(b.createdAt);
+    });
+  }
+
+  render() {
+    if (this.buildLogs().length > 0) {
+      return this.renderBuildLogsTable()
+    } else {
+      return this.renderEmptyState()
+    }
+  }
+
+  renderBuildLogsTable() {
+    return (
+      <table>
+        <thead>
+          <tr>
+            <th>Source</th>
+            <th>Output</th>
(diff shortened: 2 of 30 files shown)


jmhooper:
This is no longer WIP, but we should hold off on merge until corresponding changes have been made to 18F/federalist-docker-build.

api/policies/buildCallback.js
```diff
@@ -3,7 +3,7 @@
  * Requests must have a `token` param that matches config and an `id`
  */
 module.exports = function (req, res, next) {
-  if (req.param('token') !== sails.config.build.token || !req.param('id')) {
```
brandocalrissian: Now the above comment is incorrect--- which is right, the comment or the code?

Code wins. Flawless victory. Might I suggest removal of the entire comment?

jmhooper: Amended the commit

assets/app/components/site/siteBuildLogs.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import buildLogActions from "../../actions/buildLogActions";
+
+class SiteBuildLogs extends React.Component {
+  componentDidMount() {
```
brandocalrissian: I would prefer we move away from fetches triggered from view lifecycle actions, but we repeat this idiom many times in the application.

It's better to have the data load call not part of the view so the view is just a view of given state. As such, this view will flash with a "no build logs" until it gets the data from the Federalist API.

I'm definitely not going to hold back this merge for a change because we have accepted this design pattern in the client many times.

jmhooper: I'm curious about how to avoid triggering fetches in lifecycle callbacks?

For reference, a pattern I'm using to working in is [Presentation/Container components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0). I've used that pattern in this situation by having a stateful container component that triggered the fetch and then updates the props of stateless presentational components appropriately.

To avoid the situation where we display an empty state while waiting on data I'd configure the state to look like this:

```javascript
buildLogs: {
  state: "pending", // ["pending", "fulfilled", "rejected"],
  data: [],
  error: null,
}
```

When the `componentWillMount` method is called on the container I dispatch an action that will update the state to pending, trigger the fetch, and then dispatch and action to update the state to fulfilled / rejected with either data or an error. 

I didn't use that here because our state only seems to save the data fetched from the API and I wanted to stay consistent with the rest of the code in the frontend. 

The above pattern has worked for me, but still depends on lifecycle callbacks to trigger fetches, so I'm curious about other methods.

brandocalrissian: "Something changed! We need to show the site build logs!"

At the very moment when we know we're going to want to have a siteBuildLogs component, we also know we want to fetch the data needed by that component. We should fetch the data then instead of relying on the React lifecycle.

If we know enough to change the redux state to signify we're waiting on data, then we know enough to trigger the fetch whose response will populate the redux state through a state transition. The closer we can align the causality of these two events (the redux "pending" transition and the data fetch), the easier it'll be to follow the dependency.

If the fetching bit is triggered in a "container" component whose only purpose is to trigger the fetch of the data, then we're looking at a path where we...
  * Alter the redux state to put us in the "we're about to care about the build logs" state
  * A particular container component is instantiated by React as the state is passed down the render tree
  * React calls componentWillMount on the container component
  * Fetch happens
  * Response triggers the Redux state transition with the new data

Is the container responsible for holding the components that will render when the data to be fetched? That's just one more thing for it to be responsible for.

Does it really need to be that complicated?

If we trigger the fetch at the moment when we know we need the data, our path becomes
  * Alter the redux state to put us in the "we're about to care about the build logs" state
  * Fetch happens
  * Response triggers the Redux state transition with the new data

Then nothing depends on the Redux component lifecycle. Components are simpler and easier to test. Containers can have the singular responsibility of bridging the Redux state into the React props tree.

I'd use the React lifecycle to trigger things that really really really depend on that lifecycle, e.g. "we need to wait to run this code until after we're certain our components have mounted in the DOM."

Issue #695, opened 2016-12-28
Frontend breaks on build logs


jmhooper:
The build logs component breaks on the build logs for `18f/joining-18f`: https://federalist-staging.fr.cloud.gov/sites/18/builds

The console output there looks like this:

```
bundle.js:4237 Uncaught TypeError: Cannot read property 'username' of undefined
    at getUsername (https://federalist-staging.fr.cloud.gov/js/bundle.js:4237:14)
    at https://federalist-staging.fr.cloud.gov/js/bundle.js:4322:24
    at Array.map (native)
    at SiteBuilds (https://federalist-staging.fr.cloud.gov/js/bundle.js:4320:28)
    at https://federalist-staging.fr.cloud.gov/js/bundle.js:59063:16
    at measureLifeCyclePerf (https://federalist-staging.fr.cloud.gov/js/bundle.js:58832:12)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (https://federalist-staging.fr.cloud.gov/js/bundle.js:59062:14)
    at ReactCompositeComponentWrapper._constructComponent (https://federalist-staging.fr.cloud.gov/js/bundle.js:59037:21)
    at ReactCompositeComponentWrapper.mountComponent (https://federalist-staging.fr.cloud.gov/js/bundle.js:58945:21)
    at Object.mountComponent (https://federalist-staging.fr.cloud.gov/js/bundle.js:65339:35)
```

jmhooper:
The frontend throws the error in this spot:

```javascript
const getUsername = (site, id) => {
  const user = site.users.find(user => user.id === id);
  return user.username;
};
```

It appears that this can occur when a build has a user that is not included in `site.users`.

2 things to do from here:

- Handle the exception where a build may have a user that is not included in `site.users` in the frontend code above
- Investigate why a user for a build on `18f/joining-18f` is not included in `site.users`.

jmhooper:
It looks like when a webhook is received for a user that is not a part of Federalist, they are created [here](https://github.com/18F/federalist/blob/4bfbb16f87152881231d4239b501ad9236fa74a8/api/controllers/WebhookController.js#L61), but they are never added to the site. This would explain how user can be associated with a build for a site, but not with the site itself.

It is unclear if this is regression. [The E/W codebase does not seem to handle this case either](https://github.com/18F/federalist/blob/d948f150b8f212483b12dd75bdf59642ff1d4257/api/controllers/WebhookController.js#L43).

[...]

PR #696, opened 2016-12-28
Handle situation where a build's user is not in site.users array

jmhooper:
The build logs were crashing in staging because there was a build with a user that was not present in the `users` array on the SiteBuild component's `site` prop. This was causing the following error:

```
Cannot read property "username" of undefined.
```

The source of the error was the `getUsername` method. This commit fixes that method so that it can handle this exception case, but does not fix the root issue, that users building sites are not being rendered in the API response.

ref #695


jmhooper:
Let's hold off on review for this until I write some tests to verify this change.

jmhooper:
I amended the amended commit. This should be ready for review at this point.

Issue #536, opened 2016-09-30
links to get help not visible when signed in


afeld:
Logged-out:

<img width="722" alt="screen shot 2016-09-30 at 2 55 31 pm" src="https://cloud.githubusercontent.com/assets/86842/19003259/05d6659a-871e-11e6-8f56-60eb42d03ea0.png">

Logged-in:

<img width="694" alt="screen shot 2016-09-30 at 2 55 47 pm" src="https://cloud.githubusercontent.com/assets/86842/19003264/0c1ad5d0-871e-11e6-82b7-5c2552fdc0b0.png">

Would be great to have the links to "Help" and "Contact us" visible somewhere, even if they're in a footer or something.

PR #703, opened 2016-12-30
add Documentation and Contact Us link, switch up toolbar links in staging

wslack:
After this is merged, move https://github.com/18F/federalist/issues/536 to Done?

### DIFF
--- a/assets/app/components/nav.js
+++ b/assets/app/components/nav.js
@@ -13,13 +13,13 @@ const getSecondaryNav = (loggedIn, username) => {
   const authLinks = [
     <Link to="/sites">{username}</Link>,
     <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
-    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>
-    <a href="/logout">Log out</a>
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>,
+    <a href="/logout">Log out</a>,
   ];
   const unAuthLinks = [
     <a href="/auth/github">Log in</a>,
-    <a href="https://federalist-docs.18f.gov" target="_blank">Help</a>,
-    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>
+    <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>,
   ];
 
   return (loggedIn ? authLinks : unAuthLinks).map((link, index) => {
--- a/assets/app/components/nav.js
+++ b/assets/app/components/nav.js
@@ -12,6 +12,8 @@ const propTypes = {
 const getSecondaryNav = (loggedIn, username) => {
   const authLinks = [
     <Link to="/sites">{username}</Link>,
+    <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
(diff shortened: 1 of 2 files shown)


assets/app/components/nav.js
```diff
@@ -12,6 +12,8 @@ const propTypes = {
 const getSecondaryNav = (loggedIn, username) => {
   const authLinks = [
     <Link to="/sites">{username}</Link>,
+    <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>
```
jmhooper: Looks like there's a syntax error from a missing comma at the end of this line.

Also, note that ES6 gives us [trailing commas](http://dontkry.com/posts/code/trailing-commas.html), so if you wanted you could also stick a comma to the end of line 17 below to reduce diff noise and make issues with commas more obvious in the future.

assets/app/components/nav.js
```diff
@@ -12,6 +12,8 @@ const propTypes = {
 const getSecondaryNav = (loggedIn, username) => {
   const authLinks = [
     <Link to="/sites">{username}</Link>,
+    <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
```
jmhooper: In the unauthenticated view, this link is titled "Help", but here it is titled "Documentation". Not asking for a change here, just wanted to make sure we realize and are okay with that.

assets/app/components/nav.js
wslack: edits done!

assets/app/components/nav.js
wslack: changed the other "help" to "documentation"

jmhooper:
I built this and the root URL was still showing "Help" instead of "Documentation". Looks like it may also need to be updated [here](https://github.com/18F/federalist/blob/staging/assets/index.html#L58).

wslack:
Should we delete some of the code on the other page if its not being used?

jmhooper:
We probably can, but I'm not sure that it is not being used. Let's remove it and see what happens.

wslack:
you or me?

jmhooper:
Alright, so it looks like `unAuthLinks` is only used if the `Nav` component's `isLoggedIn` prop is false. That prop is passed through the `Header` component:

```javascript
const Header = ({ isLoggedIn = false, username = null }) =>
  <header className="usa-site-header" role="banner">
    <Disclaimer />
    <Nav isLoggedIn={ isLoggedIn } username={ username } />
  </header>
```

Which has it's props passed down from the `App` component where the logic to set `isLoggedIn` lives:

```javascript
render() {
  const storeState = this.state;

  return (
    <div>
      <Header
        isLoggedIn={ !!storeState.user }
        username={storeState.user.username}
      />
      <!-- ... -->
    </div>
  );
}
```

What's interesting about this, it that `main.js`, the entry point for the frontend has a check for the presence of `storeState.user` and does not render if that is false:

```javascript
store.subscribe(() => {

  if (!state.user) return;

  render(/* ... */);
});
```

What this means is this code is in fact never used since `isLoggedIn` is only false if there is no user in the state, and if there is no user in the state the frontend will not render.

wslack:
Neat. Syntax wise, is it better just to comment it out with a note to refer to the HTML

jmhooper:
I think we should remove it completely if it is dead code. I started the work and it looks like removing the logic for  `isLoggedIn` breaks some tests. With that in mind I advise merging this and moving that work to another branch.

jmhooper:
The PR to remove the `isLoggedIn` logic is open at #704

PR #704, opened 2017-01-03
Remove unauthenticated logic from Header and Nav

jmhooper:
The array `unAuthLinks` is only used if the `Nav`
component's `isLoggedIn` prop is false. That prop is passed through the
`Header` component:

```javascript
const Header = ({ isLoggedIn = false, username = null }) =>
  <header className="usa-site-header" role="banner">
    <Disclaimer />
    <Nav isLoggedIn={ isLoggedIn } username={ username } />
  </header>
```

Which has it's props passed down from the `App` component where the
logic to set `isLoggedIn` lives:

```javascript
render() {
  const storeState = this.state;

  return (
    <div>
      <Header
        isLoggedIn={ !!storeState.user }
        username={storeState.user.username}
      />
      <!-- ... -->
    </div>
  );
}
```

What's interesting about this, it that `main.js`, the entry point for
the frontend has a check for the presence of `storeState.user` and does
not render if that is false:

```javascript
store.subscribe(() => {

  if (!state.user) return;

  render(/* ... */);
});
```

What this means is this code is in fact never used since `isLoggedIn` is
only false if there is no user in the state, and if there is no user in
the state the frontend will not render.

With the above in mind, this commit removes the logic that renders the
unauthenticated links from the Nav / Header components.

### DIFF
--- a/assets/app/components/app.js
+++ b/assets/app/components/app.js
@@ -41,7 +41,6 @@ class App extends React.Component {
     return (
       <div>
         <Header
-          isLoggedIn={ !!storeState.user }
           username={storeState.user.username}
         />
         {children && React.cloneElement(children, {
--- a/assets/app/components/header.js
+++ b/assets/app/components/header.js
@@ -4,22 +4,18 @@ import Disclaimer from './disclaimer';
 import Nav from './nav';
 
 const propTypes = {
-  isLoggedIn: React.PropTypes.bool,
   username: React.PropTypes.oneOfType([
     React.PropTypes.string,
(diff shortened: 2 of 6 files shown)


jmhooper:
This is WIP to investigate failing tests.

wslack:
Hmmm `npm ERR! Failed at the federalist@0.0.0 test:client script 'mocha --compilers js:babel-register --recursive ./test/client'.`

jmhooper:
I've reviewed and updated the failing tests. This should no longer be WIP.

Issue #706, opened 2017-01-04
X-Frame-Options: DENY breaks legitimate iframes


shawnbot:
We just recently turned on public branch previews for [Marigold](https://github.com/18F/marigold), which uses a tool called [Fractal](http://fractal.build) to display front end component libraries. Fractal uses iframes to sandbox component previews from the app's chrome, and we quickly discovered that this feature is defeated on Federalist preview URLs because they inherit cloud.gov's `X-Frame-Options: DENY` header, which tells browsers to deny any request from an iframe.

I think the solution here is to set `X-Frame-Options: SAMEORIGIN` on preview URLs, and preserve the header for the rest of the Federalist app to defend against click-jacking. I'll submit a patch for this shortly.

#### References
* [initial report](https://github.com/18F/marigold/pull/24#issuecomment-270505262)
* [escalation to Federalist](https://github.com/18F/federalist/issues/428#issuecomment-270507240)
* [cloud.gov proxy nginx config](https://github.com/18F/cg-secureproxy-boshrelease/blob/e2ee94368183332e95b078b94bb87d3fcc1c6fd6/jobs/secureproxy/templates/config/nginx.conf.erb#L85-L115)
* [Federalist preview proxy code](https://github.com/18F/federalist/blob/master/api/controllers/PreviewController.js#L22-L25)

wslack:
Relevant slack convo: https://gsa-tts.slack.com/archives/federalist/p1481152486000133

wslack:
Currently erroring during sprint review demo; moving back to "In Progress."

wslack:
This was done but now we have a bug! Opened new issue for that.

PR #707, opened 2017-01-04
Set X-Frame-Options: SAMEORIGIN for previews

shawnbot:
This is a potential fix for #706 that sets the `X-Frame-Options` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) to `SAMEORIGIN` so that hosted sites can display legitimate `<iframe>` elements referencing pages in the same domain. 

From a compliance perspective this may be a no-no because this _technically_ allows different Federalist sites to load one another's previews, but I think the risk of click-jacking is pretty low if we assume that most (all?) Federalist sites are entirely static. (Sites that authenticate with other services via AJAX would still, theoretically, be vulnerable.)

cc @jmhooper @NoahKunin

### DIFF
--- a/api/controllers/PreviewController.js
+++ b/api/controllers/PreviewController.js
@@ -22,6 +22,7 @@ module.exports = {
           var redirect = headers['x-amz-website-redirect-location'] ||
             headers['X-Amz-Website-Redirect-Location'];
           if (redirect) return res.redirect(redirect);
+          headers['x-frame-options'] = 'SAMEORIGIN';
           res.set(headers);
         }),
         stream = object.createReadStream().on('error', function(error) {


wslack:
If this messes with security for all sites I'm a little :thinking: but you're right that all sites are fully static (those that aren't will need additional security, methinks)

NoahKunin:
No issue from Infra. It's the lack _of_ the header that's prohibited. Any setting that is intentional and has business value is allowed, especially `SAMEORIGIN`.

wslack:
@shawnbot do you want this merged into `master` or `staging` (the codebase going though ATO?) I would think staging makes more sense (and so long as we aren't supporting customers with whatever you are deploying, it's fine compliance wise)

jmhooper:
@shawnbot: I took this for a test drive locally and it looks like it works 👍

I'm going to merge this into staging and look at it there.

jmhooper:
Looks good on staging

wslack:
@jmhooper safe to put on master, or should we just send this PR to staging?

jmhooper:
It _shouldn't_ be a problem, but I'd prefer to wait to see if @shawnbot needs it in E/W production first

wslack:
@shawnbot do you need in current production, given that Federalist-in-GovCloud is a few weeks out?

brandocalrissian:
Can cherry-pick the commit from staging that fixes the test that's failing in travis here?

brandocalrissian:
From the rest of the comments, seems this is good to go. I'm too concerned about the failing test in this case as it is unrelated to the changed code.

brandocalrissian:
should be merged with make_deploys_work_again (which is really our master production branch -- will rename at some point.

brandocalrissian:
Oh, look! I can change the base branch to make_deploys_work_again. Now we have some conflicts...

brandocalrissian:
Magic! I'm not fast enough!

jmhooper:
@shawnbot: This failed to deploy from CI b/c of expired deployer creds, but I've deployed it manually for you.

Issue #648, opened 2016-11-25
Audit and cleanup routes


jmhooper:
We have some routes that aren't very "RESTful", e.g. `/build/status/:id/:token`. We should survey the routes we have and fix issues like that where they appear.

jmhooper:
Here's a quick list of the routes that we expose:

# Application Routes

##### React

- GET /site(/*)?

##### Authentication

- GET /auth/github
- GET /auth/github/callback
- GET /logout

##### Webhooks

- GET /webhook/github

##### Previews

- GET /preview/:repo/:owner/:branch(/*)?

# API Routes

##### Builds

- POST /v0/build
- GET  /v0/build/:id
- POST /v0/build/:id/status/:token
- POST /v0/build/:id/restart

##### Build logs

- POST /v0/build/:build_id/log/:token
- GET  /v0/build/:build_id/log

##### Sites

- POST   /v0/site
- GET    /v0/site
- GET    /v0/site/:id
- PUT    /v0/site/:id
- DELETE /v0/site/:id

##### Users

- GET /v0/me
- GET /v0/user/:id
- GET /v0/user/usernames

Note that Sails does not provide an easy way to audit the routes the application exposes and leans very heavily on "shadow routes", so this list was built by hand. I'm pretty certain it includes everything though. Ref: https://github.com/balderdashy/sails/pull/2659#issuecomment-77417170

wslack:
And are all these calls done without any auth?

jmhooper:
All of them except for `GET /auth/github` and `GET /auth/github/callback` require session auth.

wslack:
awesome, thanks

jmhooper:
Here is a slimmed down definition for our API using the swagger spec. We may want to consider putting a more complete version of this in our public directory so that we can use `/swagger.yaml` with some of the other tools swagger provides.

From this point we can make changes to this spec to describe the API's desired state and work from the deltas between this specs and the desired one.

```yaml
swagger: '2.0'
info:
  title: Federalist API
  description: API for interacting with Federalist
  version: "0.0.0"
host: federalist-staging.fr.cloud.gov
schemes:
  - https
basePath: /v0
produces:
  - application/json
paths:
  /build:
    get:
      summary: Fetch all of the user's builds
      responses:
        200:
          description: An array of builds
  /build/{id}:
    parameters:
    - name: id
      in: path
      description: The id of the build
      type: integer
      required: true
    get:
      summary: Fetch a build by its ID
      responses:
[...]

Issue #205, opened 2015-10-15
Handle error messages consistently


dhcole:
We're handling error messages in a few different ways. Sometimes they're just strings; sometimes there's node Error objects.

On the client, sometimes we surface errors from the `error` query param, sometimes from the response of an error ajax response. Sometimes the server redirects (previews) and sometimes it doesn't (403 access error on any other page).

We should come up with and implement a consistent error reporting plan.

dhcole:
Another example: user interface errors: https://github.com/18F/federalist/pull/240/files#diff-fea22b01f3d3a16912a825704e0da8e0R28

wslack:
We want:

- [ ] A consistent way to communicate error messages
- [ ] A way to make sure that error messages for lacking admin permissions display correctly.

related to #648.

PR #711, opened 2017-01-10
Use consistent API error response format

jmhooper:
This commit uses the following response structure for every API error response:

```json
{
  "status": 403,
  "message": "You are not authorized to do that thing"
}
```

### Error response handlers

This change heavily modifies the response helpers for errors. For generic errors, an error response handler was added which attempts to determine what type of error has occurred and pipe it to the correct response helper. This done by:

1. *Checking the status:* If the error has a status code that matches a supported HTTP error status code, then the error will be piped to the error handler for that status code. Additionally, if the error is simply a number, (e.g. `throw 404`) a generic error for the corresponding status code will be used.
1. *Checking if the error is a Waterline validation error:* Validation errors from Waterline result in a 400.
1. *Creating a generic 500:* If the error response handler cannot make a determination, it renders a 500 with a generic message.

```javascript
// Render a 404
res.error(404)

// Render a 403
res.error({
  status: 403,
  message: "You need admin privileges to do that."
})

// Render a 500
Promise.resolve().then(() => {
  undefined.property
}).catch(err => {
  res.error(err)
})
```

### Controllers

This commit modifies the way errors are handled in controllers to use the new error response handlers.

### Tests

This commit adds the error responses to the Swagger file, which adds a JSON schema for errors. This commit uses that JSON schema in the tests to verify that errors match the expected format.

### DIFF
--- a/api/controllers/BuildController.js
+++ b/api/controllers/BuildController.js
@@ -7,9 +7,7 @@ const verifyUserIsAuthorizedToRestartBuild = (user, build) => {
     const index = user.sites.findIndex(site => site.id === build.site)
 
     if (index < 0) {
-      const error = new Error("Unauthorized")
-      error.statusCode = 403
-      throw error
+      throw 403
     }
   })
 }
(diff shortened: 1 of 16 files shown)


jmhooper:
Note that this commit uses `jmhooper-swagger` as its base. That is to keep noise from that branch out of the diff. The base for this commit should be changed to staging after the changes from `jmhooper-swagger` are applied to staging.

jmhooper:
Ref #205
Ref #648

Issue #713, opened 2017-01-10
"Cannot read property 'type' of undefined" masks frontend errors


jmhooper:
When errors occur the frontend is not displaying the correct error message and is instead displaying "Cannot read property 'type' of undefined". Here's a screenshot of the site list after I attempted to add a site that I did not have admin access to:

![image](https://cloud.githubusercontent.com/assets/963654/21811833/a6f4c582-d716-11e6-9601-830f115d0c2a.png)

My suspicion is that this is because we're dispatching `undefined` instead of an action at some point. This can be seen in the logs thanks to our redux logging middleware. When the reducer attempts to determine the actions type (`action.type === "SOME_ACTION`) this error occurs.

PR #715, opened 2017-01-10
Quit dispatching undefined actions

jmhooper:
The `alertActions` and `dispatchActions` were calling `dispatch` with a function that itself called dispatch instead of an action creator. In effect, this is what was happening:

```javascript
dispatch(dispatch(action))
```

Since `dispatch` returns `undefined`, the result was the dispatch of an undefined action. This caused the Redux store to throw an error when it attempted to inspect the action. Specifically, `Cannot read property 'type' of undefined` when it attempted to inspect the action's type.

This commit fixes the problem by removing the outer `dispatch` call.

Ref #713

### DIFF
--- a/assets/app/actions/alertActions.js
+++ b/assets/app/actions/alertActions.js
@@ -34,6 +34,10 @@ export default {
   },
 
   update(isStale) {
-    dispatch(this[isStale ? 'clear': 'setStale']());
+    if (isStale) {
+      this.clear()
+    } else {
+      this.setStale()
+    }
   }
 };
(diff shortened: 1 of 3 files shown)


assets/app/actions/alertActions.js
```diff
@@ -34,6 +34,10 @@ export default {
   },
 
   update(isStale) {
-    dispatch(this[isStale ? 'clear': 'setStale']());
```
brandocalrissian: Wow. That was complicated and completely obfuscated!

PR #716, opened 2017-01-11
Disable blueprint

jmhooper:
This commit disables blueprint which automatically generates routes and actions in favor of being explicit about routes and actions. The advantages of this are:

1. More clarity about how the app works / what's exposed
1. Better interoperability to another JS web app framework

This commit also removes the `/v0/user/:id` route since it was effectively identical to `/v0/me` and was never used by the frontend.

This commit also adds rules to the JSON schemas to verify that certain properties aren't in response body for users, functionality that was previously wrapped in the tests for `/v0/user/:id`, even though that doesn't make sense.


jmhooper:
Ref #717

config/routes.js
```diff
@@ -1,56 +1,49 @@
...
+  // Frontend
+  "get /sites(/*)?": "MainController.index",
+
+  // Previews
+  "get /preview/:owner/:repo/:branch(/*)?": "PreviewController.proxy",
+
+  /**
+    Webhooks
+  */
+  "post /webhook/github": "WebhookController.github",
```
brandocalrissian: Would it make sense to version the webhook endpoint? Eh... It's internal as far as the "clients" are concerned.

config/routes.js
```diff
@@ -1,56 +1,49 @@
...
+  // Auth
+  "get /auth/github": "AuthController.github",
+  "get /auth/github/callback": "AuthController.callback",
+  "get /logout": "AuthController.logout",
+
+  // Frontend
+  "get /sites(/*)?": "MainController.index",
+
+  // Previews
+  "get /preview/:owner/:repo/:branch(/*)?": "PreviewController.proxy",
```
wslack: @jmhooper this might be the source of the issue?

PR #718, opened 2017-01-17
Remove authorizations from sails policies

jmhooper:
This PR removes the code that authorizes users in the API from the Sails policies. Instead it introduces a concept of an "Authorizer".

Authorizers provide functions that map to API actions. These functions take a user and a record as an argument and return a Promise that resolves if the user is authorized for the given action, and rejects otherwise.

This change will more easily interpolate to a web application framework outside of Sails.

This borrows from the concept of policies in Rails, but unfortunately Sails has already borrowed the term "policies" for middlewares mapped to API actions.


jmhooper:
As of now, this change has only been applied to the Build API actions, and is WIP until applied to other API actions.

jmhooper:
I've added an authorizer for the User and Site API, so this commit should be ready to go.

PR #728, opened 2017-01-25
Use Postgres database in the tests

jmhooper:
The tests were running using an in-memory database instead of a postgres database. This was going to create problems when migrating away from Waterline because we'll have to leave behind the in-memory adapter. Additionally, using the in-memory store potentially covers up problems that may occur when we use a SQL based database.

This commit adds a connection to the test configuration to use a postgres database named `federalist-test`. This commit also moves around the testing npm scripts so that the test database is migrated before the tests run.

This commit adds a database chore to drop all records the database is storing before running the tests so that records from previous tests won't cause interference.

The changes in this commit revealed some issues in the test / application logic that were also fixed.


api/models/Site.js
```diff
@@ -48,6 +48,12 @@ module.exports = {
       obj.siteRoot = `http://${s3Config.bucket}.s3-website-${s3Config.region}.amazonaws.com`
       obj.viewLink = obj.domain || [obj.siteRoot, 'site', obj.owner, obj.repository].join('/')
 
+      for (key in obj) {
+        if (obj[key] === null) {
+          obj[key] = undefined
+        }
+      }
```
brandocalrissian: I'm curious: why do we need to scrub these representations and transform null values to undefined? What does it mean semantically? I presume it's related to our JSON stringification...

It's done here and with the User model.

jmhooper: Prior to this commit, the in-memory store did not enforce any kind of schema, so values that were not set were undefined. With the Postgres adapter, those become null. When Sails serializes a JSON response, it filters undefined, but not null.

This caused problems with the JSON schemas which expected values to be undefined or of a particular type. It also caused some weird issues in the frontend where keys for values that weren't present where expected to be missing.

PR #729, opened 2017-01-26
Add Sequelize

jmhooper:
This PR will switch out Waterline with Sequelize. At this time, the PR is WIP with only the basic config and foundational bootstrap code to make the switch.


api/modelss/build-log.js
```diff
@@ -0,0 +1,24 @@
...
+    output: {
+      type: DataTypes.STRING,
+    },
+    source: {
+      type: DataTypes.STRING,
+    },
+  }, {
+    tableName: "buildlog",
+    classMethods: {
+      associate,
```
brandocalrissian: Is there a reason to define the associations in classMethods like this instead of as described in the [Sequelize docs](http://docs.sequelizejs.com/en/latest/docs/associations/#creating-elements-of-a-belongsto-has-many-or-hasone-association)? I don't see anything in the docs indicating we have to do something different to define associations for [legacy tables](http://docs.sequelizejs.com/en/latest/docs/legacy/).

Is there an advantage to doing it this way? It seems to rely on the reader knowing/figuring out how classMethods work on top of things like `BuildLog.belongsTo(...)` instead of simply defining the relationship inline.

Everything else in this PR looks good. I cross-referenced the types with our sails models.

jmhooper: When you setup these associations, you need all of the models defined. This does not work when the modules are described in multiple files because you end up with circular requires and models that are being associated before they are defined.

There is an [issue for this](https://github.com/sequelize/sequelize/issues/4439) and [supposedly the team is working on a fix for the next version](https://github.com/sequelize/sequelize/issues/6524). In the meantime, the suggestion in the open issue is the follow [this pattern which is laid out in the Express sample project](https://github.com/sequelize/express-example/blob/master/models/task.js#L8-L16).

brandocalrissian: Require circular dependency. Got it. Thank you for the explanation!

jmhooper:
I've merged all of the commits regarding the sequelize migration into this branch. I'll continue to merge sequelize PRs into this one when they are approved, and keep this one up to date with staging. When the migration is done, then we can move the entire thing into staging.

PR #730, opened 2017-01-29
Update buildpack and node versions.

jmcarp:
I noticed that the node buildpack in use here was released in 2015. This updates the manifest to use the current buildpack from cloud.gov. I also updated the node versions in use so that the application will deploy with the latest buildpack--but I didn't actually deploy, so please verify before merging. Also, it looks like travis is running tests on node v6.9.4, so you all might want to either update your cloud foundry deployment to use the same version or downgrade travis to use the older version.


wslack:
Hey @jmcarp; thanks for this.

 `master` is not the code we've been working on to deploy in GovCloud (it's the old code we had on E/W). We're going to retire it after finishing the move to GovCloud. Should we invest the time in `master`? I think the live code is a bit fragile.

jmcarp:
Probably not! I would apply this change to whatever branch you're using and change the default branch.

wslack:
Fair point re default branch. Thanks!

jmcarp:
The context is that I noticed a different project using an outdated buildpack, then searched the 18f gh org for others, and wound up here (among other places). I would still apply these changes or similar to staging, since IMO it's best to stick to the cached buildpacks if at all possible--that way cloud.gov is responsible for updates, not you (although you still have to restage to apply updates).

jmhooper:
This looks good to me, but needs to be changed up a bit for staging. I'm going to cherry-pick and amend the commit and then push directly to staging.

jmhooper:
Closed via https://github.com/18F/federalist/commit/0a45310d9d0717726209f1235b5cb32f87894f98

PR #731, opened 2017-01-30
[HOLD] Testing harness for Sequelize models

jmhooper:
This commit sets up a testing harness for the Sequelize models. Changes include:

- Rewrite a factory helper that creates Sequelize models instead of Waterline models
- Add pending tests for Sequelize models
- Use Sequelize models to clean the database instead of Waterline models

### DIFF
--- a/test/api/bootstrap.test.js
+++ b/test/api/bootstrap.test.js
@@ -3,9 +3,10 @@ const Sails = require('sails')
 
 let sails
 
-const _cleanDatabase = (sails) => {
-  const promises = Object.keys(sails.models).map(key => {
-    return sails.models[key].destroy({})
+const _cleanDatabase = () => {
+  const models = require("../../api/modelss")
+  const promises = Object.keys(models).map(name => {
+    return models[name].destroy({ where: {} })
   })
   return Promise.all(promises)
 }
@@ -19,7 +20,7 @@ before(function(done) {
     sails = server;
     if (err) return done(err);
 
-    _cleanDatabase(sails).then(() => {
+    _cleanDatabase().then(() => {
       done(null, sails);
     }).catch(err => {
       done(err)


test/api/support/ffactory.js
```diff
@@ -0,0 +1,98 @@
...
+
+  return Object.assign({
+    owner: repository.owner,
+    repository: repository.name,
+    engine: "jekyll",
+    users: users
+  }, overrides)
+}
+
+let repositroyNameStep = 1
```
brandocalrissian: repositroy --> repository typo fix

test/api/support/ffactory.js
```diff
@@ -0,0 +1,98 @@
+const Promise = require("bluebird")
```
brandocalrissian: Since we've moved to node v6 (yea!), can we start moving to using native [es2015 promises](http://node.green/) instead of bluebird? Mocha's happy with whatever promise implementation we throw at it...

jmhooper: We also use some of Bluebird's collection helpers that are implemented by Node's native Promises in v6. Things like `Promise.props` and `Promise.map`.

brandocalrissian: I'm not excited about using a proprietary Promise to get some syntactic sugar, but I'm not going to be dogmatic about it. This chunk of code is pretty meta, so making it easier to read is a good thing.

jmhooper: I can look for mixins to replace the helpers we use. I'd like to see us using the native promises as well.

test/api/support/ffactory.js
brandocalrissian: I vote we fix this typo and approve this PR. There are more important things to do than removing the bluebird dependency; it's not related to the de-sails-i-fi-ca-tion.

PR #732, opened 2017-02-01
[HOLD] Implement Sequelize model logic and remove Waterline models

jmhooper:
This commit adds the logic and tests for the Sequelize models. It also contains a significant rewrite of the factory helper to leverage some of Sequelize's power. Finally, this commit removes all of the old Waterline model code.

Note that this commit is not integrated with the rest of the application. It should not be merged until a changeset that integrates it is added on top.


api/modelss/build-log.js
```diff
@@ -5,19 +5,50 @@ const associate = ({ BuildLog, Build }) => {
   })
 }
 
+const beforeValidate = (buildLog) => {
+  return saniztizeBuildSecrets(buildLog)
+}
+
+const saniztizeBuildSecrets = (buildLog) => {
```
brandocalrissian: saniztizeBuildSecrets -> sanitizeBuildSecrets

I know this is from the original, but might as well fix typos on the way...

PR #733, opened 2017-02-03
Apply Sequelize changes to the rest of the application

jmhooper:
The following need to be updated to use Sequelize models instead of Waterline models.

- [x] Authorizors
- [x] Controllers
- [x] Models
- [x] Policies
- [x] Responses
- [x] Services


jmhooper:
Everything is cleaned up and should be ready to go.

test/api/unit/authorizers/build.test.js
```diff
@@ -6,76 +6,84 @@ const authorizer = require("../../../../api/authorizers/build.js")
...
+      const site = factory.site({ users: Promise.all([user, otherUser]) })
       Promise.props({ user, otherUser, site }).then(({ user, otherUser, site }) => {
         return authorizer.create(user, { user: otherUser.id, site: site.id })
+      }).then(() => {
+        done(new Error("Expected authorization error"))
       }).catch(err => {
         expect(err).to.equal(403)
         done()
-      })
+      }).catch(done)
```
brandocalrissian: Why do we need these trailing `catch(done)` calls? Is this an expected path in the test?

brandocalrissian: I don't understand why a test could/should go through multiple done() paths.

jmhooper: Mocha expects async tests to call `done(err)` when there is a failure.

Trailing promises with `.catch(done)` catches an exception that is generated if the promise is rejected which tells mocha to stop the tests and report the error.

Otherwise, we have to wait for the test to timeout. The error that mocha reports is that the test timed out, and we have to scroll to where the test executed and look for where "Unhandled promise rejection" is output.

brandocalrissian: Ah, yes. Mocha. Cool.

PR #747, opened 2017-02-09
Hide site editor components

jmhooper:
This commit hides the site editor components by removing their routes from the route definitions and removing the links from the site side bar nav.

This commit does not remove the code associated with them that is now dead code.


wslack:
Merging after CI is good.

PR #754, opened 2017-02-13
Cleanup dead code after deleting the editor

jmhooper:
This commit pulls out all of the dead code that was orphaned when the editor was disabled. This commit also updates the tests so they don't fail without the removed files.

I tested this pretty thoroughly, but we may want to watch on staging for a while before merging into master.


package.json
```diff
@@ -7,42 +7,29 @@
...
-    "lodash.isequal": "^4.4.0",
-    "mime": "^1.3.4",
     "moment": "^2.10.3",
     "newrelic": "^1.22.0",
-    "node-pre-gyp": "^0.6.30",
     "passport": "^0.2.1",
     "passport-github": "^0.1.5",
-    "passport-local": "^1.0.0",
     "pg": "^4.4.0",
     "prosemirror": "^0.8.3",
```
brandocalrissian: Seems like we can get rid of prosemirror too, right?

jmhooper: Yes indeed. I think there may be some scripts for it that we can get rid of now

jmhooper:
I just pushed a commit that gets rid of prosemirror and codemirror.

Issue #691, opened 2016-12-23
Consolidate logic that interacts w/ GitHub


jmhooper:
Currently, the code that handles interaction with GitHub is spread across both the frontend and the backend. In some places this makes sense, for example content management on the frontend. In other places it does not make sense, for example repo creation on the frontend when adding a site, then webhook creation on the backend.

We should examine the ways that our app communicates with GitHub and move communication involving site creation and administration into the backend and content management onto the frontend.

wslack:
Closing as complete!

PR #762, opened 2017-02-17
Add SiteCreator service

jmhooper:
Creating sites is a very complex process, and prior to this commit, that logic was spread across multiple modules. This led to a number of issues:

 - No clear way to provide an end-to-end test of site creation.
 - Site creation logic lives in both the frontend and the backend.
 - Some authenticated actions (such as creating a repository) are triggered from the frontend which means the API must deliver a GitHub access token to the frontend.

 This commit adds a SiteCreator service which handles all of the activities necessary to site creation. This commit also cleans up the tests of the modules that used to handle site creation since that is now tested in the SiteCreator tests.

 ref #691


jmhooper:
This is currently work in progress. As it stands, the backend work is done, but I still have to remove the site creation logic from the fronted.

jmhooper:
I amended the commit to contain the work needed on the frontend. This is no longer WIP.

api/services/SiteCreator.js
```diff
@@ -0,0 +1,134 @@
...
+  siteParams = _paramsForNewSite(siteParams)
+
+  if (template) {
+    return _createSiteFromTemplate({ siteParams, template, user })
+  } else {
+    return _createSiteFromExistingRepo({ siteParams, user })
+  }
+}
+
+_checkExistingGithubRepository = ({ user, owner, repository }) => {
```
jmhooper: Habit. I can switch remove the underscores.

api/services/SiteCreator.js
```diff
@@ -0,0 +1,134 @@
...
+  return GitHub.getRepository(user, owner, repository).then(repository => {
+    if (!repository) {
+      throw {
+        message: `The repsoity ${owner}/${repository} does not exist.`,
```
jmhooper: Classic

Issue #763, opened 2017-02-17
Replace Sails w/ Express


jmhooper:
This is an action item from consideration on #690

PR #764, opened 2017-02-21
Extract sails

jmhooper:
This PR begins the process of extracting Sails's configuration management by replacing it with something that works the same, but without Sails. The PR adds a `configg` directory, and any configs that are not needed are moved there. An `index.js` file was added to that directory so it could be required and then used like `sails.config`.

When we are done extracting Sails, we can rename the `configg` directory to `config`.

This PR also removes the sails-db-migrate module. This module depends on the configs for connecting to the database be in Sails's config. Since I did not want to have them in 2 places just to support migrations, I tacked on a commit that replaces sails-db-migrate with our own implementation.


wslack:
ref #763

jmhooper:
Hold for #768

brandocalrissian:
This is ready whenever. I _think_ we're ready now since we merged staging into master earlier today.

wslack:
Should @brandocalrissian review again? (I can't tell)

jmhooper:
I'll stand by to see if he feels like he needs to. I pushed 2 small commits to fix issue that arose while rebasing. It looks like this also needs another fix b/c the test database is not getting migrated.

brandocalrissian:
If the tests pass for this, this is approved by me. We've reviewed all the different contributions and I checked out the changes made after the rebase.

PR #771, opened 2017-02-22
Proxy objects in S3 bucket that match a prefix

jmhooper:
The logic in the S3Proxy service class was assuming that objects where the path prefix matched more than one bucket object were directories. This was not the case.

An example of an exception is a file named `uswds.min.js` and `uswds.min.js.map` in the same directory. They'll both match against `uswds.min.js`. The result of this is the S3Proxy attempting to proxy from `uswds.min.js/index.html`.

This commit solves this problem by doing the following:

1. Finding an object with a key that is an exact match
2. Checking the size of the object and resolve with a directory if the object has a size of 0

### DIFF
--- a/api/services/S3Proxy.js
+++ b/api/services/S3Proxy.js
@@ -60,7 +60,12 @@ const resolveKey = (path) => {
       if (err) {
         reject(err)
         return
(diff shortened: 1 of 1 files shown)


wslack:
👏 👏 👏

PR #773, opened 2017-02-23
Use jekyll as the build engine for template sites

jmhooper:
The template sites are all jekyll sites, but default build engine for sites is the "static" build engine. Because of this, the "jekyll" option needs to be set explicitly in the SiteCreator. This commit does that.

ref #722

### DIFF
--- a/api/services/SiteCreator.js
+++ b/api/services/SiteCreator.js
@@ -81,6 +81,7 @@ const createSiteFromExistingRepo = ({ siteParams, user }) => {
 
 const createSiteFromTemplate = ({ siteParams, user, template }) => {
   let site = Site.build(siteParams)
+  site.engine = "jekyll"
   const { owner, repository } = siteParams
 
   return site.validate().then(error => {
--- a/test/api/unit/services/SiteCreator.test.js
+++ b/test/api/unit/services/SiteCreator.test.js
@@ -281,6 +281,23 @@ describe("SiteCreator", () => {
         }).catch(done)
       })
 
+      it("should use jekyll as the build engine", done => {
+        const siteParams = {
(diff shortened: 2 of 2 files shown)


test/api/unit/services/SiteCreator.test.js
```diff
@@ -281,6 +281,23 @@ describe("SiteCreator", () => {
         }).catch(done)
       })
 
+      it("should use jekyll as the build engine", done => {
+        const siteParams = {
+          owner: crypto.randomBytes(3).toString("hex"),
+          repository: crypto.randomBytes(3).toString("hex"),
+          template: "microsite",
```
wslack: Why only the microsite here? Shouldn't this apply to all sites?

wslack:
Well, that's no good. Suggest merging this into prod when fixed given the importance of the bug.

test/api/unit/services/SiteCreator.test.js
jmhooper: This is a test case. It is set for all sites in `SiteCreator.js` above.

api/services/SiteCreator.js
```diff
@@ -81,6 +81,7 @@ const createSiteFromExistingRepo = ({ siteParams, user }) => {
 
 const createSiteFromTemplate = ({ siteParams, user, template }) => {
   let site = Site.build(siteParams)
+  site.engine = "jekyll"
```
brandocalrissian: Would it be better conceptually to default all sites to use the jekyll engine?

brandocalrissian: No, then we'd be coupling the site creation code with our assumptions around how templated sites are created. Not good.

PR #775, opened 2017-02-24
Lowercase owner/repo before looking up site for webhook

jmhooper:
Webhook requests were 400ing because the webhook payload included owner and repo names that had uppercase letters. All of our sites have lowercase names, so they were not being found in response to the webhook.

This commit adds a test and fixes the issue in the controller.


wslack:
should we merge this to master immediately?

jmhooper:
@wslack: Yeah, let's merge as soon as we see a successful build on staging.

PR #781, opened 2017-03-01
Reduce instance memory from 512MB to 256MB

jmhooper:
The Federalist web application uses about 150MB of RAM. With this in mind, this commit scales the memory the app is allocated in cloud.gov from 512MB to 256MB, which should free up some memory in our space, while leaving overhead for memory spikes.

This commit also bumps the instance count in production to 5.


brandocalrissian:
Restarted the travis-ci run as it looked to be a temporary failure condition and it's easy to get a sanity check cleared.

PR #782, opened 2017-03-01
Extract sails

jmhooper:
This PR has the code for extracting sails. At the moment it builds off of the PR that extracts the sails config, and has some code to get unit tests green and to get authentication working.

At this point, the remaining work involves getting the rest of the request specs green.

### DIFF
--- a/config/log.js
+++ b/config/log.js
@@ -1,3 +1,3 @@
 module.exports = {
-  level: process.env.SAILS_LOG_LEVEL || 'info',
+  level: process.env.LOG_LEVEL || "info",
 }
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The following environment variables are set on the Cloud Foundry environment in 
 
 - `NODE_ENV`: The node environment where the app should run. When running in Cloud Foundry this should always be set to production, even for the staging environment
 - `APP_ENV`: The application environment in which the app should run. Valid values are `production` and `staging`.
-- `SAILS_LOG_LEVEL`: Sets the log level for sails.
+- `LOG_LEVEL`: Sets the log level for the app.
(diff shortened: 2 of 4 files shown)


jmhooper:
The tests for this branch are passing, and the app appears to be working. There are still some TODOs, but due to the size of this branch, I'll move those to another branch.

This PR should be held until PRs that do the following are ready

- [ ] Quit using global objects previously provided by Sails
- [ ] Setup proper request and error logging

jmhooper:
I removed the `HOLD` on this since it is going into `jmhooper-extract-sails-config` instead of `staging`, and `jmhooper-extract-sails-config` is held.

brandocalrissian:
AHA! Will look at it.

config/log.js
```diff
@@ -1,29 +1,3 @@
...
-  *                                                                          *
-  * You may also set the level to "silent" to suppress all logs.             *
-  *                                                                          *
-  ***************************************************************************/
-
-  level: process.env.SAILS_LOG_LEVEL || 'info'
-
-};
+module.exports = {
+  level: process.env.SAILS_LOG_LEVEL || 'info',
```
brandocalrissian: Does SAILS_LOG_LEVEL have any meaning anymore after this PR?

jmhooper: We still use it, but we should rename it to just `LOG_LEVEL`.

brandocalrissian: Ah, yes, in the manifest, right where it belongs.

brandocalrissian: Agreed!

jmhooper:
Renamed the env var. See now that this is approved. It's a big PR so I'm gonna do a manual merge.

PR #783, opened 2017-03-01
added support e-mail

wslack:
Is this the right place to add support e-mail for logged in users only?

cc @saracope

### DIFF
--- a/assets/app/components/nav.js
+++ b/assets/app/components/nav.js
@@ -12,7 +12,8 @@ const getSecondaryNav = (username) => {
   const links = [
     <Link to="/sites">{username}</Link>,
     <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
-    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>,
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank">File Issue</a>,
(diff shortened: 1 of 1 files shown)


assets/app/components/nav.js
```diff
@@ -12,7 +12,8 @@ const getSecondaryNav = (username) => {
   const links = [
     <Link to="/sites">{username}</Link>,
     <a href="https://federalist-docs.18f.gov" target="_blank">Documentation</a>,
-    <a href="https://github.com/18F/federalist/issues/new" target="_blank">Contact us</a>,
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank">File Issue</a>,
+    <a href="mailto:federalist-support@gsa.gov" target="_blank">Get Help</a>,
```
jmhooper: This looks like it should work. I don't know that you need to `target="_blank"` though?

wslack: yeah I just copy-pasted. Didn't we used to haev different links for logged in and not logged in users?

jmhooper: We do. Unauthed users will see [these](https://github.com/18F/federalist/blob/staging/assets/index.html#L57-L59)

wslack: great!

wslack: Made the change you asked for

wslack:
Confirmed this is working on staging.

PR #787, opened 2017-03-03
Replace sails.log and morgan with winston

jmhooper:
This commit consolidates all logging functionality under winston, the
same logging module we use in federalist-builder. It uses the same
configurations to set the logging level, but uses a winston based
logging middleware, and uses winston's logging functions in places where
logging happens.

### DIFF
--- a/api/controllers/preview.js
+++ b/api/controllers/preview.js
@@ -1,9 +1,10 @@
+const logger = require("winston")
 const S3Proxy = require("../services/S3Proxy")
 
 module.exports = {
   proxy: function(req, res) {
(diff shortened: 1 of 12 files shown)


brandocalrissian:
Didn't notice the different base branch. Will reconfigure to merge to jmhooper-extract-sails

brandocalrissian:
Will merge manually as I can't see a quick way to do it via the github ui...

brandocalrissian:
Now this branch is merged with jmhooper-extract-sails as intended since the remove-globals branch has merged with the extract-sails branch as well.

Issue #759, opened 2017-02-14
Federalist doesn't allow custom url's or config to be removed in settings


wslack:
If I ever want to remove my custom URL, the settings page doesn't let me, even though the command appears to succeed.

wslack:
Apparently this also affects the custom config box. Sorry @nickbristow.

PR #792, opened 2017-03-06
Recognize empty strings in update site request body

jmhooper:
The Site API was not updating attributes for a site if the new value was an empty string. This was because values where being set as such:

```javascript
{
  config: req.body.config || site.config,
}
```

Since the empty string evaluates to false, this was falling back to the value on the site. This commit fixes the problem by creating a params object that is the result of merging the site and request body:

```javascript
const params = Object.assign(site, req.body)
{
  config: params.config,
}
```

Ref #759

### DIFF
--- a/test/api/requests/site.test.js
+++ b/test/api/requests/site.test.js
@@ -207,7 +207,7 @@ describe("Site API", () => {
       }).catch(done)
     })
 
-    it("sould add a user to an existing site for an existing repository", done => {
+    it("should add a user to an existing site for an existing repository", done => {
       let user, site
       const userPromise = factory.user()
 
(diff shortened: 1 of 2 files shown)


test/api/requests/site.test.js
```diff
@@ -545,5 +545,40 @@ describe("Site API", () => {
...
+            domain: "",
+          })
+          .set("Cookie", results.cookie)
+          .expect(200)
+      }).then(response => {
+        validateAgainstJSONSchema("PUT", "/site/{id}", 200, response.body)
+        return Site.findById(site.id)
+      }).then(site => {
+        expect(site.config).to.equal("")
+        expect(site.domain).to.equal("")
```
brandocalrissian: Would be nice to have a line here to validate we're keeping the site values if not given in the request e.g.
`expect(site.defaultBranch).to.equal("foo")`
where `foo` is set in our site setup code above.

jmhooper: Ah, good thought. I'll make that into a separate example.

jmhooper:
I amended the last commit with a new example in the request specs.

Issue #799, opened 2017-03-07
Sites created from a template do not respect the templates default branch


jmhooper:
The template config specifies branches for some of the templates, but Federalist ignores those and defaults to master. We should respect the value specified in the template config.

PR #800, opened 2017-03-07
Use correct default branch for template sites

jmhooper:
Not all of the templates use `master` as their default branch, but prior to this commit, we used `master` as the default branch for every site created from a template. This commit adjusts the SiteCreator so that is looks up a template and uses the branch specified by the template as the default branch.

Ref #799


api/services/SiteCreator.js
```diff
@@ -82,6 +82,7 @@ const createSiteFromExistingRepo = ({ siteParams, user }) => {
 const createSiteFromTemplate = ({ siteParams, user, template }) => {
   let site = Site.build(siteParams)
   site.engine = "jekyll"
+  site.defaultBranch = templateForTemplateName(template).branch
```
brandocalrissian: Can we guarantee the template will have a branch? It seems we're trusting the client to have good data if I read the expected call stack properly.

If we can't guarantee a valid branch name, would it make sense to default it to the old `master` default?

jmhooper: We provide the template, the client just provides the template name. If the client attempts to provide a template name for a template that does not exist, we error before we get here: https://github.com/18F/federalist/pull/800/files#diff-6eb3478dc1e9d3885c6e6431de4931f5R134

jmhooper: That is unless the question about how this responds to a malformed config? In that case, it would fallback to the Site model's default with is `master`.

brandocalrissian: I meant the data in the template object referenced here seems to come from the front end. I tend to question whatever comes in from an HTTP request.

I very well could be misunderstanding the call stack here.

jmhooper: The template object is loaded based on the template name which is what comes from the frontend. If it's unable to load a template, it'll throw an error before it gets here.

brandocalrissian: ok cool. I didn't find the loading of the template on the back-end. We're good to go.

PR #802, opened 2017-03-07
Add USWDS template; shrink & optimize theme thumbnails

### LINKED ISSUE #798: Add USWDS template to Federalist template
https://github.com/18F/federalist/blob/staging/config/templates.js

@jmhooper, do we need to do some config to force the rows to work better? That set of four is responsive right now....and I would understand if we want to remove #1 or #2 of the current options.

shawnbot:
This adds the [U.S. Web Design Standards template](https://github.com/18F/federalist-uswds-template), which is in turn based on the new [Jekyll theme](https://github.com/18F/jekyll-uswds/), to the front of the line in the list of template options in the Federalist add site page.

I've also shrunk and optimized all of the other template thumbnail PNGs, because they were _huge_. They're 500px wide now, which is probably still to big, but they're at least 1/4 the size in bytes.

Fixes #798.


wslack:
Awesome! Will defer to @jmhooper for review.

jmhooper:
image

We used to have rows of 3, but we went to rows of 4 b/c the landing page template looked lonely by itself. Thoughts on going back to rows of 3, since it looks lonely again?

jmhooper:
Here's what rows of 3 looks like:

image

jmhooper:
Ah, just noticed the converstaion in #798 where we said we wanted to do rows of 3 and to swap them around some. I'll work on a commit to do that. It's kind of involved since our current setup is not all that conducive to multiple rows.

jmhooper:
@shawnbot handled the work in the comment above in #806, so no need to hold for that anymore

config/templates.js
```diff
@@ -1,32 +1,40 @@
 module.exports.templates = {
```
jmhooper: We'll need to reorder these so that the ones we recommend (uswds, team, landing-page) are on the top row.

wslack:
I'm good but someone should check my wording edits.

jmhooper:
All green! Pushing the button...

wslack:
This will likely go live next week unless we have a bug fix to merge into `master` before then. Thank you @shawnbot! Feel free to book the time spent working specifically on the Federalist merge to Federalist.

PR #806, opened 2017-03-09
Refactor template grid

shawnbot:
This is a refactoring of the template grid on the TemplateSiteList component that:

1. Switches the grid size from 4 to 3 cells wide, because 4 was really not nice
1. Accommodates (theoretically) any number of templates, even when the count is less than the number of available "slots"

Here's a preview with the current 4 templates:

image

I've also tested it with 1, 2, 3, and 5 templates—no problem!

I made one other unrelated change in 668e81a to resolve a warning about `storeState.user.id` having the wrong prop type: it was `string`, but it was getting a number, so I changed it to `number`. No more console warnings!

Tests pass locally, but I haven't looked at the coverage for this view yet, so it's hard to say whether I've broken anything.

This is the first time I've committed any React or JSX code, so please let me know if I've done anything egregious or weird. I ran `eslint` on the whole AddSite directory, and it didn't report any warnings or errors. 😬

### DIFF
--- a/assets/app/components/AddSite/TemplateSiteList/index.js
+++ b/assets/app/components/AddSite/TemplateSiteList/index.js
@@ -5,9 +5,33 @@ import TemplateSite from './templateSite';
 const propTypes = {
   templates: React.PropTypes.object.isRequired,
   handleSubmitTemplate: React.PropTypes.func.isRequired
-}
+};
+
+const MAX_CELLS_PER_ROW = 3;
+const CELL_WIDTHS = ['', 'whole', 'half', 'third', 'fourth'];
+
+/**
+ * Create a two-dimensional array of values that represent one or more rows
+ * of values, with `perRow` values per row.
+ *
+ * @param {array<*>} values
+ * @param {integer} perRow
+ * @return {array<array<*>>}
+ */
+const createRowsOf = (values, perRow) => {
+  let row = [];
+  const rows = [row];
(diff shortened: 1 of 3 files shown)


assets/app/components/AddSite/TemplateSiteList/index.js
```diff
@@ -27,31 +51,50 @@ class TemplateList extends React.Component {
...
+    // fill the space with them
+    const cellsPerRow = Math.min(templateKeys.length, MAX_CELLS_PER_ROW);
+    // generate a two-dimensional array of template keys
+    const templateRows = createRowsOf(templateKeys, cellsPerRow);
+    // i.e. 'whole', 'half', 'third', or 'fourth'
+    const cellSize = CELL_WIDTHS[cellsPerRow];
+
+    let index = 0;
+
+    const templateGrid = templateRows.map((row, rowIndex) => {
```
jmhooper: Not egregious, but a refactor point down the line may be to pull this logic out into an external `renderTemplateGrid` function. Then we can invoke it as `{this.renderTemplate(templates)}` below.

shawnbot: Yeah, that makes a lot of sense. I've been thinking in USWDS terms about how we'd describe a "flowing grid" like this and make components that implement the pattern. Maybe this is the genesis of... general-purpose USWDS React components?? 😱 @juliaelman

Issue #677, opened 2016-12-09
(Possible) Bug: Redis issue causes homepage to load slowly and prevents login


wslack:

jmhooper:
This is the second time that this issue has occurred, so we're going to move from the assumption that it is a one-time problem.

Some background on this issue:

We use a redis instance as our session store. We believe this is happening because our app cannot connect to the redis instance to inspect the session for the request. In the time before the request for the session from redis times out, the site hangs. This also prevents auth from happening since we can't authenticate a session if we can't read/write from the session store.

There's 2 pieces of evidence that we use to support the hypothesis that the issue is with the connection to the session store:

0. In the past, scaling the app to 1 instance and using the in-memory session store instead of redis has solved the problem. Switching back to redis has caused it to re-appear.
0. The last time we saw this issue, we solved it by provisioning a new redis instance and switching over to that one

jmhooper:
Okay, it's starting to look like the issue described above is not what's actually happening. I just restarted the app with `cf restart federalist-staging` and it started working.

jmhooper:
We're looking at another incident of this bug today. I attempted to fix the site by restarting Federalist, but no luck.

jmhooper:
I scaled the app down to 1 instance and removed the Redis session store config (Sails will fallback to an in-memory store) here: https://github.com/18F/federalist/commit/1633da08e3570da480d8c9fbc46c801c51af4f94

That commit also short circuits CI to deploy to staging without needing to be merged. That build is here: https://travis-ci.org/18F/federalist/builds/183698496

That change was deployed successfully, and the app is now back up. That is consistent with the piece of evidence posted above and supports the thought that this is an issue with the redis instance:

> In the past, scaling the app to 1 instance and using the in-memory session store instead of redis has solved the problem. Switching back to redis has caused it to re-appear.

wslack:
We think this is fixed, but are not sure.

wslack:
Kubernetes had a defect that got rescheduled, but cloud.gov thinks they have resolved core issue.

wslack:
We're going to let this sit for a few weeks to see if it recurs.

wslack:
Haven't seen additional reports, closing

wslack:
Happened again yesterday!

wslack:
Happened again today on `federalist-prod`. We're wondering if there's any in-memory storage happening. @jmhooper is looking at this.

wslack:
no update yet

PR #807, opened 2017-03-10
Use connect-session-sequelize as the session store instead of connect…

jmhooper:
…-redis

We've had trouble with our redis connection in cloud.gov resulting in some downtime. This commit remediates by switching off of the redis session store and instead using the postgres database as the session store.

This commit adds a connect addapter and configures it to connect to our database.

Ref #677

### DIFF
--- a/app.js
+++ b/app.js
@@ -20,16 +20,12 @@ const bodyParser = require('body-parser')
 const methodOverride = require('method-override')
 const expressWinston = require("express-winston")
 const session = require("express-session")
-const RedisStore = require("connect-redis")(session)
+const PostgresStore = require("connect-session-sequelize")(session.Store)
 const responses = require("./api/responses")
 
 const app = express()
-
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -20,16 +20,12 @@ const bodyParser = require('body-parser')
...
 
 const app = express()
-
-if (config.redis) {
-  config.session.store = new RedisStore(config.redis)
-} else {
-  config.session.store = new session.MemoryStore()
-}
+const sequelize = require("./api/models").sequelize
+config.session.store = new PostgresStore({ db: sequelize })
```
brandocalrissian: It's a little odd to see something called "db" given an ORM instance as they're not the same thing, but that's what they want in https://github.com/mweibel/connect-session-sequelize. There's nothing we can do about it.

This was the only confusing thing I saw in here.

PR #810, opened 2017-03-13
add to deletion confirm message

wslack:
adds to error message so that users know we are deleting from S3


wslack:
endorsed. Merge at will!

PR #811, opened 2017-03-13
Give branch previews their own tab

jmhooper:
This PR has the work to do 2 things:

1. Move branch previews to their own tab in the SiteContainer
2. Make branch preview options reflective of what is published to S3 instead of what is on GitHub

Currently, this PR has one commit which does the work on the API. That commit needs some clean up at the moment. Once that is done, the remaining work will be to add the new tab to the SiteContainer.


jmhooper:
Just pushed the commit to show these in the frontend. This should be ready for review now.

jmhooper:
Yep

jmhooper:
Changed the base and rebased. This one should be good to go.

frontend/components/site/sitePublishedBranches.js
```diff
@@ -0,0 +1,56 @@
...
+    if (!this.props.publishedBranches) {
+      return []
+    }
+    return this.props.publishedBranches.filter(branch => {
+      return branch.site.id === parseInt(this.props.params.id)
+    })
+  }
+
+  render() {
+    if (this.publishedBranches().length > 0) {
```
brandocalrissian: Do we care about the edge case where we show "loading published branches..." when there are no published branches yet?

brandocalrissian: We should be able to tell when we've fetched them or not based on if `props.publishedBranches` exists.

brandocalrissian: Oh, sorry, no. Not with the reducer like it is now.

test/frontend/actions/publishedBranchActionsTest.js
```diff
@@ -0,0 +1,48 @@
...
+      "../util/federalistApi": {
+        fetchPublishedBranches: fetchPublishedBranches,
+      },
+      "../store": {
+        dispatch: dispatch,
+      },
+    }).default
+  })
+
+  it("fetchPublisehdBranches", done => {
```
brandocalrissian: `Publisehed` --> `Published`

Issue #812, opened 2017-03-14
Respect the path component of custom domains


jmhooper:
If you specify a custom domain with a path the BASEURL variable in the container is set to an empty string, rather than the path. For example, a site with the custom domain `https://page.18f.gov/acqstack-journeymap` should have the baseurl `'journeymap'`, but instead has the baseurl `''`.

This bug is present in both E/W and GovCloud.

PR #813, opened 2017-03-15
Respect path component of custom domains

jmhooper:
Prior to this commit, Federalist did not respect the path component of custom URLs. This caused problems for sites that were not displayed at the root for their domain, e.g. `www.opm.gov/opm-site`. This commit modifies the service that sends builds to the build scheduler to send the path component for custom domains.

ref #812

### DIFF
--- a/test/api/unit/services/SQS.test.js
+++ b/test/api/unit/services/SQS.test.js
@@ -78,13 +78,28 @@ describe("SQS", () => {
 
     context("building a site's default branch", () => {
       it("should set an empty string for BASEURL in the message for a site with a custom domain", done => {
-        factory.site({ domain: "www.example.com", owner: "owner", repository: "repo", defaultBranch: "master" }).then(site => {
-          return factory.build({ site: site, branch: "master" })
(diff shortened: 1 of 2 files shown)


test/api/unit/services/SQS.test.js
```diff
@@ -101,6 +101,18 @@ describe("SQS", () => {
         }).catch(done)
       })
 
+      it("should respect the path component of a custom domain when setting BASEURL in the message", done => {
```
brandocalrissian: It's not clear to me this new test covers all of the cases within `baseURLForCustomDomain` in `SQS.js`

jmhooper: 🤔 hmm, good call I think. I'll change these up to give us better coverage.

jmhooper: I amended the commit. This test and the test without a path component both have an array of domains they iterate over to make sure that all return the same correct result.

brandocalrissian:
Cool. Looks great.

Issue #804, opened 2017-03-08
302 for preview requests without a trailing slash


jmhooper:
When S3 gets a request to a URL that ultimately resolves to an index document, and the request does not have a trailing slash, it renders a 302 with the location set to the path with the trailing slash.

For example, `/search` redirects to `/search/`. This behavior clips query params, which can cause unexpected behavior in production that does not appear in the previews.

We should adjust the S3Proxy service so it behaves like S3 in this case.

PR #821, opened 2017-03-21
Replicate S3 redirect behavior on preview URLs.

jmhooper:
When S3 gets a request with a trailing slash, it renders the object with the key plus `index.html`. For example, `/about/` will render the object at `/about/index.html`. When the request does not have a trailing slash, S3 will redirect to that path with a trailing slash. In the about example, `/about` renders a 302 redirect to `/about/`. That behavior is setup by us in the build container [here](https://github.com/18F/federalist-docker-build/blob/fa0c7f37a7ce8e554f430de7e2f05cb31c9a24de/publish.sh#L31-L34).

Prior to this commit, preview URLs did not behave like this. A request to `/about` would render the object at `/about/index.html`, bypassing the redirect. This created some issues. For example, the redirect on S3 clips query params. `/search?query=asdf` would redirect to `/search/`. These sorts of issues were not discovered until sites were in production because the behavior was not mirrored on preview URLs.

This commit configures the S3Proxy to behave like S3 with regard to URLs that don't have a trailing slash.

Ref #804

### DIFF
--- a/api/services/S3Proxy.js
+++ b/api/services/S3Proxy.js
@@ -24,17 +24,21 @@ const findSite = (req) => {
     if (!site) {
       throw httpError({
         message: `No such site: ${owner}/${repository}.`,
-        code: 404,
+        status: 404,
       })
     }
     return site
   })
 }
 
-const httpError = ({ message, code }) => {
+const httpError = ({ message, status }) => {
   const error = new Error(message)
-  error.code = `${code}`
+  error.status = status
   return error
+}
+
+const pathIsPreviewRoot = (path) => {
+  return path.split("/").length <= 4
 }
 
 const pipeS3ObjectToResponse = ({ key, res }) => {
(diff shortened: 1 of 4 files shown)


api/services/S3Proxy.js
```diff
@@ -51,8 +51,18 @@ const pipeS3ObjectToResponse = ({ key, res }) => {
 }
 
 const resolveKey = (path) => {
+  if (path.split("/").length <= 4) {
```
brandocalrissian: I have to think way too much about what this means. Can we encode this condition by naming it what we expect it to represent?

jmhooper:
> We're getting into some dangerous territory with the branch complexity without any tests.

Agreed. I'll work on a commit to finally add some tests for this.

jmhooper:
Just pushed a commit that adds tests and addressed the comment. This should be ready for re-review.

Issue #117, opened 2015-08-13
when I go to a federalist site and am prompted to authenticate, I should be sent to the site after authentication


ultrasaurus:
when I go to a federalist site and am prompted to authenticate, I should be sent to the originally requested URL after authentication

Steps:
1. Go to a site (branch) that requires authentication
2. Login
Result: I see Federalist page asking me if I want to create a site
Expected:  I should be re-directed to the URL I originally tried to go to

dhcole:
Across the site, if you access a page without signing in first, #303 will take users back to the home page. The next step to accomplish this (and similar issues on other pages) would be to capture the user's intended URL in a cookie or local storage, and then after successful authentication, redirect the user back to that route. Or we could try passing that path as a query parameter back to the API controllers `/auth` routes, which would need to then use it to change the `redirect_uri` sent to github so github sends the user back directly to the desired page after authentication.

PR #824, opened 2017-03-24
Redirect to preview after sign in

jmhooper:
Prior to this commit, users who were not signed in attempting to look at a preview were redirected to the home page. In order to see the preview, they would have to login, then go back to see the preview.

This commit smooths that process out by redirecting back to previews after sign in. It does this by saving the URL the user was attempting to visit in the session, then redirecting to the sign in path. After sign in, it redirects back to the URL.

Ref #117

### DIFF
--- a/test/api/requests/auth.test.js
+++ b/test/api/requests/auth.test.js
@@ -1,5 +1,6 @@
 const expect = require("chai").expect
 const cookie = require('cookie')
+const crypto = require("crypto")
 const nock = require("nock")
 const request = require("supertest-as-promised")
 const config = require("../../../config")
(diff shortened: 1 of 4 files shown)


test/api/requests/auth.test.js
```diff
@@ -190,5 +191,37 @@ describe("Authentication request", () => {
...
+    it("should redirect to a redirect path if one is set in the session", done => {
+      const sessionKey = crypto.randomBytes(8).toString("hex")
+      const sessionBody = {
+        cookie: {
+          originalMaxAge: null,
+          expires: null,
+          httpOnly: true,
+          path: "/"
+        },
+        authRedirectPath: "/path/to/something",
```
brandocalrissian: I get this string matches the string below in L#222, but can we make this a const value instead of trusting the two different string literals to match?

Issue #333, opened 2016-03-18
Add text area in settings to allow --future and other configs for previews only


gboone:
(Edit: the solution [we will pursue](https://github.com/18F/federalist/issues/333#issuecomment-270668812) is to add a "Preview configs" text area in the setting which contains configs that are only added to preview branches. The configs for drafts or future could be added there.)

Hey team, I was working on publishing some blog posts for 18f.gsa.gov that are slated for future publication. Right now we have to give them a phony, earlier-than-now `date` if we want to see them on the federalist preview branch and then re-date them when it's time to publish. It'd be great if there were an option to allow a branch to publish these posts in the preview URL even if they're not "published" yet.

wslack:
This is a jekyll constraint, I think, so Federalist can't fix it. Will confirm with team (and please pushback if I'm wrong!)

jmhooper:
We could add the `--drafts` option to the `jekyll build` command in the build container if we're on a preview branch.

ref: https://jekyllrb.com/docs/drafts/

wslack:
@jmhooper ooooo - if there's a way to do that, that sounds neato. I don't like the discrepancy between a preview branch and live branch, though.... @gboone what do you think?

gboone:
I think that's the right track. Maybe `--future` rather than (or in addition to) `--drafts`? 

You've got a good point about having a discrepancy between preview and live. I could see some site owners being confused why something is happening in one place and not the other. You could document it well but I'm not sure it's the precedent you want to set. 

To get around that, you could have a system where the site owners define some settings in their `_config.yml` that only apply in preview. Something like:

```
preview:
  future: true
```

wslack:
I like that idea; I wonder how hard it would be to implement.

jmhooper:
There's a number of ways it could be done. If we add it to the config.yml then we'd need to parse the yaml in the build container and apply it to overall configuration there which may not be the most straightforward way to do it.

Another option to consider may be to add a "Preview configs" text area in the setting which contains configs that are only added to preview branches. Then the configs for `drafts` or `future` could be added there. That could also be useful for other things.

Finally, we could add a check box, similar to how we handle public previews, that controls whether preview branch builds get the `--drafts` or `--future` option.

My vote is for the option with the preview configs in settings, but I am open to other ideas.

wslack:
I concur with the preview configs and will edit the issue accordingly.

shawnbot:
[...]

PR #827, opened 2017-03-27
Use separate custom configs for site and preview builds

jmhooper:
This commit adds a new, separate text area which, in addition to the pre-existing config text area, can be used to set separate custom configs for site and preview builds. Site builds get the configs in the site configuration text area, and preview builds get the configs in the preview configuration text area.

Ref #333


wslack:
cc @shawnbot to check since this *should* allow your Rubygem to go the way of the dodo (I think)

PR #841, opened 2017-03-31
Merge after 2017-03-31 Retro

jmhooper:


brandocalrissian:
Looks like travis failed because of the "too many CI builds going at the same time and the venerable app is already there..."

From travis log...

> Renaming app federalist-staging to federalist-staging-venerable in org gsa-18f-federalist / space staging as gsa-18f-federalist_deployer...
> error: Server error, status code: 400, error code: 100002, message: The app name is taken: federalist-staging-venerable

wslack:
huh. let's discuss in AM?

package.json
```diff
@@ -106,7 +106,7 @@
     "watchify": "^3.2.2"
   },
   "engines": {
-    "node": "~6.9.0",
-    "npm": "~3.10.8"
+    "node": "6.x.x",
```
wslack: what is impact of setting these to "x"

jmhooper: It means that they can take on any value. So the version can be 6.9.5 or 6.10.1

wslack:
this is still pending the bugfix for the issue we found in staging, I think

PR #842, opened 2017-04-03
Update node version to 6.x.x

jmhooper:
This commit upgrades our node version to 6.x.x. I mistakenly thought the 6.9.x family was the LTS, but it turns out it is 6.x.x. As a result of the LTS moving to 6.10.1, cloud.gov dropped support for our node version.

Also, when I was debugging this, I noticed that moving off sails made our deps skinny enough that we may not have to vendor them, so I added `node_modules` to our `.cfignore`.


.cfignore
```diff
@@ -52,6 +52,7 @@ lib-cov
 *.out
 *.pid
 npm-debug.log
+node_modules/
```
brandocalrissian: cool

PR #845, opened 2017-04-05
Remove bluebird promises

jmhooper:
This commit removes bluebird promises in favor of using native promises. The only function we used that was provided by bluebird was `Promise.props` so this commit pulls in the `promise-props` module to add that.

### DIFF
--- a/test/api/bootstrap.test.js
+++ b/test/api/bootstrap.test.js
@@ -1,3 +1,5 @@
+Promise.props = require("promise-props")
+
 const AWSMocks = require("./support/aws-mocks")
 
 const _cleanDatabase = () => {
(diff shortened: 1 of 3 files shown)


test/api/bootstrap.test.js
```diff
@@ -1,3 +1,5 @@
+Promise.props = require("promise-props")
```
brandocalrissian: promise-props is not in the package.json under devDependencies

jmhooper: Oops, good catch

jmhooper:
Amended the commit to include the `promise-props` module as a dev dep in the `package.json`.

PR #850, opened 2017-04-05
Select cf creds based on environment

jmhooper:
This commit changes the way credentials are set during CI deploys so that different environments have different credentials. It also removes the credentials from the Travis config, setting them in Travis env vars instead so they can be rotated without changing the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,7 @@ before_script:
 env:
   global:
   - CF_API="https://api.fr.cloud.gov"
-  - CF_USERNAME="gsa-18f-federalist_deployer"
   - CF_ORGANIZATION="gsa-18f-federalist"
-  # CF_PASSWORD for gsa-18f-federalist_deployer user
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,10 +16,6 @@ before_script:
 env:
   global:
   - CF_API="https://api.fr.cloud.gov"
-  - CF_USERNAME="gsa-18f-federalist_deployer"
-  - CF_ORGANIZATION="gsa-18f-federalist"
```
brandocalrissian: As long as CF_ORGANIZATION is set in the environment, too, we should be good.

jmhooper: It is not. I think in my mind I was deleting `CF_PASSWORD`.

jmhooper: Another good catch to start the day 🎣

jmhooper:
Amended the commit to put the `CF_ORGANIZATION` var back.

Issue #851, opened 2017-04-07
Sites can be added twice if the name has an uppercase character


jmhooper:
If you have a Federalist site named `18f/federalist-site`, and you attempt to add a Federalist site named `18F/federalist-site`.

wslack:
Done!

PR #852, opened 2017-04-07
Lowercase owner and repo before looking for existing site at site creation

jmhooper:
The check to make sure a site was not added twice by looking for the owner / repo name was broken if the name of the site to be added contained uppercase characters. This was because we lowercase the name of the site / repo when adding then, but not when checking for their existence.

This commit lowercases the name of the owner and repo before checking for their existence.

ref #851

### DIFF
--- a/api/services/SiteCreator.js
+++ b/api/services/SiteCreator.js
@@ -73,7 +73,7 @@ const createSiteFromExistingRepo = ({ siteParams, user }) => {
   const { owner, repository } = siteParams
 
   return Site.findOne({
-    where: { owner, repository },
+    where: { owner: owner, repository: repository },
     include: [ User ],
   }).then(model => {
     site = model
@@ -129,8 +129,8 @@ const paramsForNewBuildSource = (templateName) => {
 }
 
 const paramsForNewSite = (params) => ({
-  owner: params.owner,
-  repository: params.repository,
+  owner: params.owner ? params.owner.toLowerCase() : undefined,
+  repository: params.repository ? params.repository.toLowerCase() : undefined,
   defaultBranch: params.defaultBranch,
   engine: params.engine,
 })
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -48,7 +48,7 @@
     "watch:client:js": "watchify frontend/main.js -o public/js/bundle.js -t [ babelify --presets [ es2015 react stage-2 ] ]",
     "test": "npm run test:prepare && npm run test:server && npm run test:client",
     "test:prepare": "NODE_ENV=test npm run migrate:up && npm run build",
-    "test:server": "NODE_ENV=test mocha --timeout 6000 test/api/bootstrap.test.js test/api/unit/**/*.test.js test/api/requests/**/*.test.js",
+    "test:server": "NODE_ENV=test mocha --timeout 1000 test/api/bootstrap.test.js test/api/unit/**/*.test.js test/api/requests/**/*.test.js",
```
wslack: why this change, out of curiosity?

jmhooper: That's a sneaky change I made b/c I was frustrated waiting on the 6 second timeout. Now that Sails is in our past most tests run in under 100ms and all should be under 1 second.

api/services/SiteCreator.js
```diff
@@ -73,7 +73,7 @@ const createSiteFromExistingRepo = ({ siteParams, user }) => {
   const { owner, repository } = siteParams
 
   return Site.findOne({
-    where: { owner, repository },
+    where: { owner: owner.toLowerCase(), repository: repository.toLowerCase() },
```
brandocalrissian: I'm a little weirded out we understand the owner and repo are always lowercased, but this file doesn't also contain the code enforcing said behavior. This is a split responsibility which would be easier to maintain if it were unified.

Is that an easy possibility?

jmhooper: Hmm, we could do it as soon as the params enter the `SiteCreator` 🤔

brandocalrissian: That seems reasonable! Anything that lets us see "we lowercase the owner/repo coming in when creating and looking for sites.

jmhooper:
I've amended the PR to move the code that downcases the the owner / repo name into the function that gets the site's params.

PR #853, opened 2017-04-10
Allow sites to specify an image path for template images

jmhooper:
This commit adds a `thumb` prop to the template configuration. Here, a link to a thumbnail image for the template can be set. This will allow template maintainers to link to external template URLs and avoid having to PR Federalist to change the thumbnail when the template changes.

This commit should be followed up with commits to link to template thumbs in the templates' respective repos so they can be maintained there.


wslack:
Do we want to move the links right now or just merge? I'm ok with just merging.

jmhooper:
I was thinking we could merge, and then PR the template repos. Then we can PR this again to delete the thumbnails. That way we don't have a PR open that could potentially break the thumbnails.

wslack:
SGTM

PR #854, opened 2017-04-10
Fix linter

jmhooper:
This commit reconfigure ESLint so that it works

-  Changes up ESLint rules
- Upgrade ESLint
- Handle peer dependencies
- Update the eslintignore
- Fix the `lint` npm script.


jmhooper:
Right now the linter can find 572 problems. Running it with `--fix` can get it down to 98, but some of the decisions it makes look questionable so that rules may need to be fixed up some.

brandocalrissian:
I'm not able to run the linter-- when I `npm run lint` locally, it's totally confused by es2015 syntax. What am I doing wrong?

jmhooper:
@brandocalrissian: What version of node are you using? It should be v6.

brandocalrissian:
Node v6.9.0. I nuked the node_modules just for fun and that didn't help either.

jmhooper:
Weird. It works on my machine, but I was able to recreate the error in a Docker container. I could fix it by using the `babel-eslint` parse. I'll amend the commit.

jmhooper:
@brandocalrissian: Commit amended.

brandocalrissian:
Specifying the babel-eslint parser made it work.

PR #855, opened 2017-04-11
Fix intermittent GitHub auth test failure

jmhooper:
There was an intermittent GitHub auth test failure that I was able to track down to an issue with the nock for the GitHub access token request. It appears that the `passport-oauth` strategy, which the `passport-github` strategy is built on occasionally sends an additional `redirect-uri` param. This appears to depend on the URL that `supertest` assigns the app when it sends the request. When this happens, the request body for the nock does not match.

This commit changes the request body matching the access token nock to use a function instead of an object so an exact match is not required and the additional params will no longer break the test.

### DIFF
--- a/test/api/support/githubAPINocks.js
+++ b/test/api/support/githubAPINocks.js
@@ -12,10 +12,15 @@ const accessToken = ({ authorizationCode, accessToken, scope } = {}) => {
   }
 
   return nock("https://github.com")
-    .post("/login/oauth/access_token", {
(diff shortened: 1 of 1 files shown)


jmhooper:
Here's an example of one of these intermittent failures: https://travis-ci.org/18F/federalist/builds/220685829

jmhooper:
Oh yeah. Luckily yesterday the sun and the planets must have been in the right configuration or something for it to fail more consistently than it has in the past.

Issue #801, opened 2017-03-07
Unauthorized screen for someone trying to log into Federalist is unfriendly


wslack:
The screen for a user who can't log into Federalist is an ugly UNAUTHORIZED message in plain text. We should make it friendlier!

PR #856, opened 2017-04-11
Add home component

jmhooper:
This is a WIP commit to change the content displayed in `index.html` into a component. This will pave the way for us to address issues like #801.

At moment, my thinking is to move the static HTML into a string in the component, and render that when the component mounts. When the component receives props, we check for a user. If a user is present, we redirect to `/sites`.


jmhooper:
I've updated the tests and the commit message so this should be ready to go. The commit message:

```
This commit adds a new component which is meant to be displayed at the root before the user logs in. As soon as a user is present in the state, the component redirects to `/site`.

The component has a bunch of static HTML, much of which does not elegantly translate to JSX, so the HTML is stored alongside the component in `home.html`. The HTML is loaded as a string using `stringify` and then set when the component is rendered.
```

frontend/reducers/user.js
```diff
@@ -5,6 +5,10 @@ import {
 export default function user(state = false, action) {
   switch (action.type) {
   case USER_RECEIVED:
+    if (!action.user) {
+      return false
```
brandocalrissian: TIL `false.foo` is something that can actually have a useful meaning. We're getting a little lucky here, but it works.

frontend/components/home.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  redirectForLoggedInUser() {
+    if (this.props.storeState.user) {
+      pushHistory("/sites")
+    }
+  }
+
+  render() {
+    return <div dangerouslySetInnerHTML={{ __html: homeHTML }}/>
```
jmhooper: Yeah, there's a few things that trip up react. Also note the SVG markup in there.

brandocalrissian:
Will merge when the CI comes back green.

PR #860, opened 2017-04-13
Add user loading state

jmhooper:
This commit adds a loading state for the user data. The following modifications are present in this commit

- Add a `USER_FETCH_STARTED` action to change the user state to a loading state
- Modify the `fetchUser` action to dispatch the `USER_FETCH_STARTED` action before sending the request
- Modify the user reducer to store the user data alongside a `isLoading` property that marks if the user is loading
- Change the home component to not render until the user is done loading
- Change other components to work with the new user loading state


jmhooper:
Note that #856 is the base of this PR. When #856 is completed, we'll want to rebase this and change the base for the PR.

frontend/components/loadingIndicator.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const LoadingIndicator = () => (
+  <div className="main-loader" id="main-loader">
+    <div className='uil-ring-css' style={{transform: 'scale(0.6)'}}>
+      <div></div>
```
brandocalrissian: I'm presuming this empty div is here for a reason...

PR #861, opened 2017-04-13
Remove builds reducer

jmhooper:
The authoritative list of builds lives alongside the site data in the site redux state, making all of the code to build the build redux state dead code. This commit cleans all of that code out since it isn't used for anything.


brandocalrissian:
Can merge when the CI sanity check goes green.

PR #865, opened 2017-04-17
Add loading state for published files list table

jmhooper:
This commit changes the redux state for published files so that it can be used to determine if published files are being loaded. It also changes the published files list table so that it displays a loading message when files are loading.

### DIFF
--- a/frontend/components/siteContainer.js
+++ b/frontend/components/siteContainer.js
@@ -64,9 +64,7 @@ class SiteContainer extends React.Component {
     const publishedBranches = storeState.publishedBranches.filter(branch => {
       return branch.site.id === site.id
     })
-    const publishedFiles = storeState.publishedFiles.filter(file => {
-      return file.publishedBranch.site.id === site.id
-    })
+    const publishedFiles = storeState.publishedFiles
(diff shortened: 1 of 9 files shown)


frontend/components/siteContainer.js
```diff
@@ -64,9 +64,7 @@ class SiteContainer extends React.Component {
     const publishedBranches = storeState.publishedBranches.filter(branch => {
       return branch.site.id === site.id
     })
-    const publishedFiles = storeState.publishedFiles.filter(file => {
-      return file.publishedBranch.site.id === site.id
-    })
```
brandocalrissian: We're relying on the `sitePublishedFilesTable` component to trigger `publishedFileActions.fetchPublishedFiles(site, branch)` during the `componentDidMount` hook in order to reset and refetch the published files when displaying them. I wasn't expecting this coupling, but I don't have a better solution since we've committed to using the mount hooks to key our data fetches. Given that, yes, the filter removed here was extraneous.

jmhooper:
Looks like this got merged into `jmhooper-published-files-api`, but never into staging. I'll take of that real quick.

brandocalrissian:
👍

PR #867, opened 2017-04-18
Add loading state for build logs

jmhooper:
This commit changes the redux state for build logs so that it marks when a build log data is being loaded from the API. Additionally, it displays a loading message in place of the build logs table when the data is loading.

### DIFF
--- a/frontend/components/site/siteBuildLogs.js
+++ b/frontend/components/site/siteBuildLogs.js
@@ -7,26 +7,26 @@ class SiteBuildLogs extends React.Component {
   }
 
   buildLogs() {
-    if (!this.props.buildLogs) {
-      return [];
-    }
 
-    return this.props.buildLogs.filter(log => {
-      return log.build.id === parseInt(this.props.params.buildId);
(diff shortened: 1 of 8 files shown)


frontend/components/site/siteBuildLogs.js
```diff
@@ -7,20 +7,18 @@ class SiteBuildLogs extends React.Component {
...
-    });
   }
 
   render() {
     if (this.buildLogs().length > 0) {
       return this.renderBuildLogsTable()
+    } else if (this.props.buildLogs.isLoading) {
+      return this.renderLoadingState()
     } else {
       return this.renderEmptyState()
```
brandocalrissian: I think it's a little interesting we did
```
if (this.props.publishedFiles.isLoading) {
       return this.renderLoadingState()
} else if (!files.length) {
      return this.renderEmptyState()
} else {
      return this.renderPublishedFilesTable(files)
}
```
for site published files and broke down the three-part condition differently here. I'd pick whichever one makes sense and go with it.

Issue #832, opened 2017-03-29
Add loading states


jmhooper:
Currently our app has no loading states. When you go to a page, you'll often see a message saying something like "No Sites added", which will then disappear and be replaced with a list of sites. We should add a message that says something like "Loading sites" that is then replaced with a "No Sited added" or a list of sites depending on whether the user has added any sites.

jmhooper:
The following parts of the UI should display a loading state while we wait for data:

- [x] Home page while we determine if the user is logged in
- [ ] Site list while site data is loading
- [ ] GH branch list while GH branch data is loading
- [ ] Build tables while build data is loading
- [ ] Published branch table while published branches are loading
- [ ] Published branch files table while files are loading

wslack:
So many changes made, yay!

PR #875, opened 2017-04-19
Update message for people loading site list

wslack:
Add note that the site list flashes a "no sites" message

### DIFF
--- a/frontend/components/siteList/siteList.js
+++ b/frontend/components/siteList/siteList.js
@@ -12,7 +12,7 @@ const getSites = (sites) => {
     return (
       <div className="usa-grid">
         <h1>No sites yet.</h1>
-        <p>Add one now.</p>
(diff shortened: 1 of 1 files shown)


frontend/components/siteList/siteList.js
```diff
@@ -12,7 +12,7 @@ const getSites = (sites) => {
     return (
       <div className="usa-grid">
         <h1>No sites yet.</h1>
-        <p>Add one now.</p>
+        <p>Add one now. (This message may also appear briefly before your site list loads).</p>
```
jmhooper: This works for now. Eventually we should replace with a loading state (#832).

wslack: yep; it was loading slowly for me when tethered so I did a quick fix for the moment

brandocalrissian:
Should we merge this puppy?

jmhooper:
Yep!

PR #876, opened 2017-04-19
Fix issues with Bluebird promises

jmhooper:
We worked on removing bluebird in #845. The work turned out to be incomplete, because we were still (incorrectly) using `Promise.props` and were still exposing the global `Promise` object.

The `Promise.props` call still worked because in JS you can treat an array like an object where the indexes are keys and we happened to be using one as such. Since we didn't remove the global Promise object, this issue was hidden until we pushed to production where the problem reveals itself.

The commit replaces `Promise.props` with the native `Promise.all`.


wslack:
Pending Travis

PR #878, opened 2017-04-20
Don't let users sign in twice

jmhooper:
Prior to this commit, if a user navigated to `/auth/github`, the authentication process would begin, regardless of whether that users was logged in or not. The result was that users who were logged in, could effectively login again, overwriting their session with a new logged in session.

This was harmless behavior until we added redirects after sign in. We added this for preview URLs. If an unauthenticated user attempted to view a preview, they would be redirected to the login url. After being logged in, they would be redirected back to the preview.

The problem arose when people attempted to view previews for sites they had not added, and therefore were not allowed to see. They would be redirected to the login, then redirected to the preview, then redirected back to login in a loop.

This commit fixes the issue, by redirecting logged in users to the root url if they attempt to re-authenticate.

### DIFF
--- a/api/controllers/auth.js
+++ b/api/controllers/auth.js
@@ -6,7 +6,11 @@ const AuthController = {
   },
 
   github: function (req, res) {
-    passport.authenticate("github")(req, res, req.next)
+    if (req.session.authenticated) {
+      res.redirect("/")
+    } else {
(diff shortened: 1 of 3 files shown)


wslack:
ahhh that makes sense

PR #879, opened 2017-04-20
Display loading indicator when waiting for user data to load

jmhooper:
This commit fixes a number of bugs that occured around the user loading states.

When the user data was loading, the site displayed a white screen instead of an indicator. This commit adds the indicator to the `<App/>` component while the user fetch is running.

When the user fetch was triggered, the `USER_FETCH_STARTED` action was not dispatched because the function was not actually invoked. This commit fixes that by adding the parens to invoke the function.

This commit removes the loading indicator from the Home component, since displaying it will be handled by the App component.

Finally, this commit adds the loading indicator to some of the components that were previously displaying loading messages.

### DIFF
--- a/frontend/actions/userActions.js
+++ b/frontend/actions/userActions.js
@@ -15,7 +15,7 @@ const dispatchUserReceivedAction = user => {
 
 export default {
   fetchUser() {
-    dispatchUserFetchStartedAction
+    dispatchUserFetchStartedAction()
     return federalist.fetchUser()
       .then(dispatchUserReceivedAction)
   },
--- a/frontend/components/app.js
+++ b/frontend/components/app.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import store from '../store';
 import alertActions from '../actions/alertActions';
+import LoadingIndicator from "./loadingIndicator"
 import Header from './header';
 
 class App extends React.Component {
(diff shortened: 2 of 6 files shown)


frontend/actions/userActions.js
```diff
@@ -15,7 +15,7 @@ const dispatchUserReceivedAction = user => {
 
 export default {
   fetchUser() {
-    dispatchUserFetchStartedAction
+    dispatchUserFetchStartedAction()
```
brandocalrissian: good catch

frontend/components/site/sitePublishedBranchesTable.js
```diff
@@ -62,7 +64,7 @@ class SitePublishedBranchesTable extends React.Component {
   }
 
   renderLoadingState() {
-    return <p>No published branch data available</p>
+    return <LoadingIndicator/>
```
brandocalrissian: Is it ever possible to have no published branch data here (we loaded it and it's empty)?

jmhooper: Potentially, if we build an empty site.

brandocalrissian: Are we OK with showing those users the loading indicator? That's a much smaller edge case than the expected flash-of-missing-content-notice case, so this is still better than what we have.

brandocalrissian: But that's just, like, my opinion, man.

jmhooper: Okay wait, I put this in the wrong spot. I thought this table had an empty state 😳

jmhooper:
@brandocalrissian: Found a mistake I made. Let me flip to WIP while I sort that out.

frontend/components/site/sitePublishedBranchesTable.js
brandocalrissian: gotcha

jmhooper:
Looks like I got ahead of myself. For my initial spike for the loading states I worked on the published files and then forgot that was never merged. The progress indicator I added in the published files table component has been removed.

This is no longer WIP and is ready for review.

PR #891, opened 2017-05-02
Add builds loading state

jmhooper:
This commit adds the loading state for builds. There's a good bit of legwork here because the builds are tightly coupled to sites in both the frontend and the API. This commit has changes to decouple them, and then add a loading state:

- Remove `builds` prop from the site API response
- Change the builds `find` action to be scoped to a given site
- Add a `publishedAt` attribute to the site model (since the frontend can no longer use the builds array in the site API response to calculate published times)
- Add a build reducer and associated actions
- Add a loading state to the `SiteBuilds` component.

### DIFF
--- a/frontend/components/site/sitePublishedFilesTable.js
+++ b/frontend/components/site/sitePublishedFilesTable.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import publishedFileActions from "../../actions/publishedFileActions";
+import LoadingIndicator from '../loadingIndicator'
 
 class SitePublishedBranch extends React.Component {
   componentDidMount() {
@@ -56,8 +57,7 @@ class SitePublishedBranch extends React.Component {
   }
 
   renderLoadingState() {
-    // TODO: Replace with a loading component
-    return <p>Loading published files</p>
+    return <LoadingIndicator/>
   }
 
   renderEmptyState() {
--- a/frontend/reducers/builds.js
+++ b/frontend/reducers/builds.js
@@ -1,0 +1,30 @@
+import {
+  buildsFetchStartedType as BUILDS_FETCH_STARTED,
+  buildsReceivedType as BUILDS_RECEIVED,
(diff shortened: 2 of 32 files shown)


jmhooper:
Just pushed the last set of change to the frontend. This is no longer WIP

public/swagger/Site.json
```diff
@@ -35,6 +35,9 @@
     "publicPreview": {
       "type": "boolean"
     },
+    "publishedAt": {
+      "type": "string"
```
brandocalrissian: do we want to indicate this field uses a "date-time" format? not critical, but would make it easier to parse the results if you care.

Of course, if we're not following that format, then, no.

frontend/components/site/sitePublishedFilesTable.js
```diff
@@ -57,7 +58,7 @@ class SitePublishedBranch extends React.Component {
 
   renderLoadingState() {
     // TODO: Replace with a loading component
```
brandocalrissian: drop TODO as it's TODONE.

frontend/reducers/builds.js
```diff
@@ -0,0 +1,30 @@
...
+  buildsFetchStartedType as BUILDS_FETCH_STARTED,
+  buildsReceivedType as BUILDS_RECEIVED,
+  buildRestartedType as BUILD_RESTARTED,
+} from "../actions/actionCreators/buildActions";
+
+const initialState = {
+  isLoading: false,
+}
+
+export default function buildLogs(state = initialState, action) {
```
brandocalrissian: This works because we're now only storing the builds for a particular site in our state and reloading them when visiting a different site.

brandocalrissian: I confused myself for a second and then felt compelled to write it down once I came back to reality.

jmhooper: This function should be named `builds` instead of `buildLogs` though

PR #895, opened 2017-05-03
Add demo branch and demo domain to site

jmhooper:
This commit adds a feature that allows users to publish content to a new custom url named a "beta url". There are 2 configs for this: the beta domain and the beta branch.

A few changes were made to make this happen

- Add the attributes to the site model
- Add the inputs for the attributes to the site settings screen
- Change the SQS service to set the base url for the beta branch
- Change the site model to validate that the default and beta branch aren't the same
- Change the site model to validate that the domain and beta domain are not the same
- Trigger a rebuild of the default and beta branch when a site's configs are changed

### DIFF
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -102,6 +102,8 @@ module.exports = {
     }).then(() => {
       const params = Object.assign(site, req.body)
       return site.update({
+        demoBranch: params.demoBranch,
+        demoDomain: params.demoDomain,
         config: params.config,
         previewConfig: params.previewConfig,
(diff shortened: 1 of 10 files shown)


jmhooper:
Moving this to WIP because we're going to change the name of the feature from "Beta domain" to "Demo domains".

jmhooper:
I've amended the commit so `beta` is now `demo`.

brandocalrissian:
Looks like the new site properties were put in the same place as where the builds were taken out.

jmhooper:
Resolved the conflicts on this one.

PR #896, opened 2017-05-05
Site loading state

jmhooper:
This branch has 2 commits. The first removes some dead code that was previously used to manage sites in the frontend. This was mostly unused action and reducer code.

The other commit adds a loading state to the sites list. When a site fetch starts, the sites state is marked as loading and the sites list displays a loading indicator. When the fetch is complete, the sites list is displayed. This second commit is WIP at this time.

### DIFF
--- a/frontend/reducers/sites.js
+++ b/frontend/reducers/sites.js
@@ -1,4 +1,5 @@
 import {
+  sitesFetchStartedType as SITES_FETCH_STARTED,
   sitesReceivedType as SITES_RECEIVED,
   siteAddedType as SITE_ADDED,
   siteUpdatedType as SITE_UPDATED,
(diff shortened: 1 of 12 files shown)


jmhooper:
I've amended the second commit so it is no longer WIP. This PR is ready for review.

frontend/reducers/sites.js
```diff
@@ -1,39 +1,59 @@
...
+        data: state.data.concat(action.site),
+      }
+    } else {
+      return state
+    }
 
   case SITE_UPDATED:
-    return mapPropertyToMatchingSite(state, action.siteId, action.site);
+    return {
+      isLoading: false,
```
brandocalrissian: Are there places where the site is updated, deleted, or the branches are received where the sites_fetch_started signal is called? I may be missing it. Otherwise, I'm not sure I understand why we're setting `isLoading` again here and the other places below.

brandocalrissian: Oh, we can get to a site without fetching all the sites first.

PR #897, opened 2017-05-05
Add published branches loading state to the frontend

jmhooper:
This commit changes the published branches redux state so it has an `isLoading` prop and `data` prop. This allows us to know if published branches are being loaded. The changes to the state are used to add a loading indicator to the published branch list if published branches are being downloaded.

### DIFF
--- a/frontend/components/site/sitePublishedBranchesTable.js
+++ b/frontend/components/site/sitePublishedBranchesTable.js
@@ -1,6 +1,7 @@
 import React from "react";
 import { Link } from 'react-router';
 import publishedBranchActions from "../../actions/publishedBranchActions";
+import LoadingIndicator from "../loadingIndicator"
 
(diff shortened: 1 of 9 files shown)


frontend/components/site/sitePublishedBranchesTable.js
```diff
@@ -1,26 +1,29 @@
...
+      return this.props.publishedBranches.data
     }
-    return this.props.publishedBranches.filter(branch => {
-      return branch.site.id === parseInt(this.props.params.id)
-    })
   }
 
   render() {
-    if (this.publishedBranches().length > 0) {
-      return this.renderBuildLogsTable()
```
jmhooper: This should not be named `renderBuildLogsTable`. Looks like an old copy/paste issue

jmhooper: Fixed

PR #904, opened 2017-05-10
Add loading state for GitHub branches

jmhooper:
This commit adds a loading state for the table of GitHub branches on the site settings screen.

This commit also makes a number of changes to clean up the code that renders GitHub branches:

- Rename `SitePreviewLinksTable` component to `SiteGithubBranchesTable`
- Move logic for fetching GitHub branches out of site actions / reducers into specialized actions / reducers
- Add an error state to the logic that fetches GitHub branches (GH branch fetches encounter a number of recoverable error states, e.g. private repos or deleted GitHub repos)


jmhooper:
This is currently WIP

jmhooper:
This branch has been cleaned up and is no longer WIP

frontend/components/site/siteGithubBranchesTable.js
```diff
@@ -0,0 +1,71 @@
...
+const renderErrorState = () => (
+  <p>
+    An error occured while downloading branch data from Github.
+    Often this is because the repo is private or has been deleted.
+  </p>
+)
+
+SiteGithubBranchesTable.propTypes = {
+  site: React.PropTypes.object.isRequired,
+  branches: React.PropTypes.object.isRequired,
```
brandocalrissian: This line is the only difference between `siteGithubBranchesTable.js` and `githubBranchesTable.js`. Can we extract this commonality?

brandocalrissian: Upon further review, we don't seem to be referencing this file anywhere. Can we drop it?

test/frontend/components/site/siteGithubBranchesTable.test.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import SiteGithubBranchesTable from '../../../../frontend/components/site/siteGithubBranchesTable';
```
brandocalrissian: If we drop the referenced file here in favor of the other (`githubBranchesTable`), we need to change this test to cover the other file (which isn't covered right now AFAIK).

frontend/components/site/siteGithubBranchesTable.js
jmhooper: Oops, looks like I copied the file without deleting the original 😳

frontend/components/site/githubBranchesTable.js
```diff
@@ -0,0 +1,71 @@
...
+    <td>
+      <a href={previewURL({ branch, site })} target="_blank">View</a>
+    </td>
+  </tr>
+)
+
+const previewURL = ({ branch, site }) => {
+  if (branch.name === site.defaultBranch) {
+    return site.viewLink
+  } else if (branch.name === site.demoBranch && site.demoDomain) {
```
jmhooper: For published branches, the view link actually appears on the branch object, so we've cleaned up the other places this logic appears. GitHub branches are a bit different since we don't get the view link in the API response from GitHub. 

I do think this can be moved into a function, not sure if we want to go outside of this module? If so, where does something like that live? In `frontend/utils`?

jmhooper:
Okay, cleaned this up a bit. I ended up going with `SiteGithubBranchesTable` for the component name because the naming is more consistent with the other components that are scoped to a site.

Still need to work on the preview link function, pending some more discussion.

frontend/components/site/githubBranchesTable.js
brandocalrissian: Hmm... well, we could derive the view link for the branch when pulling the redux state into our component tree and then ask the branch what its previewUrl is...

brandocalrissian: (if needed, we could do something like this: http://redux.js.org/docs/recipes/ComputingDerivedData.html)

brandocalrissian:
Well, the thread about the preview link is now hidden behind the outdated (deleted) file...

PR #906, opened 2017-05-12
Push build status back to GitHub

jmhooper:
This commit adds a new service, the `GithubBuildStatusReporter`. This takes a build and uses the [GitHub status API](https://developer.github.com/v3/repos/statuses/) to push the build status back to GitHub. If the build is processing or errored, it also pushes a link to the build logs. If the build has completed successfully, it pushes a link to the preview link for the build.

Adding this required 2 changes:

First, I needed to add a `signedInAt` attribute to users. This lets us fetch users in the order they last signed in. This way, if the committing user is not a Federalist user, and we don't have an access token for them, we can instead use the access token of the last user to sign in, since that access token is likely to be the freshest.

Second, I added a `commitSha` attribute to the Build model. This was because the GitHub status API requires a commit sha in the request to tie builds to commits. I'm questioning this decision now, since it binds builds to commits, when the reality is that builds refer to branches, building the most recent commit on the branch. I may change this so it uses the most recent commit sha on the given branch.

### DIFF
--- a/test/api/unit/services/GithubBuildStatusReporter.test.js
+++ b/test/api/unit/services/GithubBuildStatusReporter.test.js
@@ -1,0 +1,218 @@
+const crypto = require("crypto")
+const expect = require("chai").expect
+const nock = require("nock")
+const factory = require("../../support/factory")
+const githubAPINocks = require("../../support/githubAPINocks")
+const { Build, User } = require("../../../../api/models")
+
+const GithubBuildStatusReporter = require("../../../../api/services/GithubBuildStatusReporter")
+
+describe("GithubBuildStatusReporter", () => {
+  afterEach(() => {
+    nock.cleanAll()
+  })
+
+  describe("reportBuildStatus(build)", () => {
+    context("with a build in the processing state", () => {
+      it("should report that the status is 'pending'", done => {
+        let statusNock
+
+        factory.build({
+          state: "processing",
+          site: factory.site({ owner: "test-owner", repository: "test-repo" }),
+          commitSha: "456def",
+        }).then(build => {
+          statusNock = githubAPINocks.status({
+            owner: "test-owner",
+            repo: "test-repo",
+            repository: "test-repo",
+            sha: "456def",
+            state: "pending",
+          })
+
+          return GithubBuildStatusReporter.reportBuildStatus(build)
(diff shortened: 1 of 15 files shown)


jmhooper:
This PR _should_ be ready to go. I want to manually push it to staging to and test it out with actual webhooks / against the actual GH API. I'll do that, make any changes, and then it should no longer be WIP.

jmhooper:
I tested this out on staging it worked besides the need for some minor tweaks. This is no longer WIP.

wslack:
@shawnbot want to try it out?

jmhooper:
@shawnbot: As soon as this PR is merged you can

test/api/unit/services/GithubBuildStatusReporter.test.js
```diff
@@ -0,0 +1,218 @@
...
+        factory.build({
+          state: "processing",
+          site: factory.site({ owner: "test-owner", repository: "test-repo" }),
+          commitSha: "456def",
+        }).then(build => {
+          statusNock = githubAPINocks.status({
+            owner: "test-owner",
+            repo: "test-repo",
+            sha: "456def",
+            targetURL: `http://localhost:1337/sites/${build.site}/builds/${build.id}/logs`,
```
brandocalrissian: We're coupling our expected host/port/scheme combination with strings here. Not critical, but a string change somewhere else can break these tests without actually breaking this functionality.

brandocalrissian: Just something to think about for now.

wslack:
@brandocalrissian @jmhooper 

It looks like repos on `staging` and `prod` can only get statuses from one of them when doing a PR. How do we make sure that prod is the one sending statuses? (18f.gsa.gov ran into this)

wslack:
cc @jseppi

jmhooper:
Hmm, perhaps the problem is the collision on the name of the app performing the status update? Maybe we can try changing the names to `federalist-staging\build` and `federalist\build`?

jseppi:
Sounds reasonable. Key off of `APP_ENV`/`environment.env`?

jmhooper:
Yeah, that makes sense to me

jseppi:
on it

PR #924, opened 2017-05-22
Show an alert on the home page after an unauthorized login attempt

### LINKED ISSUE #801: Unauthorized screen for someone trying to log into Federalist is unfriendly

jseppi:
Ref #801

Also add simple test case for the `Home` component with and without the unauthorized alert showing.

It looks like this:

screen shot 2017-05-30 at 5 53 40 pm

To Do:
- [x] Adjust authentication tests to reflect the new behavior
- [x] Review/adjust alert message (@wslack)


jmhooper:
It looks like the auth request specs are failing. It is expecting a 401, but we changed it so instead it is redirecting.

brandocalrissian:
will check it out!

wslack:
Hold on this

wslack:
edited! Thoughts?

frontend/components/home.js
```diff
@@ -32,7 +32,7 @@ class Home extends React.Component {
           <div className="usa-alert usa-alert-error usa-alert-home" role="alert">
             <div className="usa-alert-body">
               <h3 className="usa-alert-heading">Unauthorized</h3>
-              <p className="usa-alert-text">Sorry, you are not authorized to log in to Federalist.</p>
+              <p className="usa-alert-text">Apologies; you don't have access to Federalist! Please contact the Federalist team if this is in error.</p>
```
jseppi: I think you'll have to escape the `'` in `don't` --> `don\'t`

Or, even better, use `&apos;` --> `don&apos;t`

wslack: done!

jseppi:
Looks good to me, @wslack. I updated the screenshot in the root comment.

wslack:
I'm happy but travis is not.....

jseppi:
Linting errors since the PR that added linting on modified source files is now in staging branch. I'll fix those.

jseppi:
bleh, I think the associated Travis build got stuck during GitHub's API outage this morning...

wslack:
we can reboot with another small commit right?

jseppi:
Or close and reopen, it turns out.

.eslintrc
```diff
@@ -1,4 +1,15 @@
 {
   "extends": "airbnb",
-  "parser": "babel-eslint",
+  "rules": {
+    "react/jsx-filename-extension": [0],
```
jseppi: This would require touching a bunch of stuff, so I have disabled it for now

.eslintrc
```diff
@@ -1,4 +1,15 @@
...
+    "comma-dangle": ["error",
+      {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "always-multiline",
+        "exports": "always-multiline",
+        "functions": "never"
```
jseppi: Same for this one, which is having dangling commas in function argument lists.

test/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "env": {
+    "mocha": true
+  },
+  "rules": {
+      "no-unused-expressions": [0] // because chai's expect is weird
```
jseppi: This is because a lot of chai expectations, like `expect(...).to.be.equal;` use getter methods instead of functions :\

wslack:
Ok, so this is now ready to go into `staging`?

Also James, why use redirects instead of a 401 response, as Brandon mentioned?

jseppi:
I did this implementation initially as my first pairing session with @jmhooper, mostly as a learning exercise.

I actually agree now that a `401` response and a custom error page would be better, but it will be more of an implementation lift than what is here.

I think how it could better be done is:
1. extract common site components (header, footer, etc) into a template
1. refactor the home page to use the template, and stop loading its content with React
1. refactor the single-page app to use the template
1. create a custom 401 page that uses the template and return that after an unauthorized user authenticates with GitHub

brandocalrissian:
I looked at the changes since the last review and still approve. Merging.

PR #926, opened 2017-05-23
Run linter on CI

jmhooper:
This commit tells Travis to run ESLint on all changed files when a build runs for a PR. It will fail the build if the linter finds any errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,10 +1,4 @@
 {
-  "extends": "airbnb/legacy",
-  "rules": {
-    "comma-dangle": ["error", "always-multiline"],
-    "no-inner-declarations": ["error", "functions"],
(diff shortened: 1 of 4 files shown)


jmhooper:
This PR is no longer WIP

jmhooper:
Another thing to note about this commit is that it remove some of the rules to favor the standard airbnb eslint config.

jmhooper:
I would like to get @jseppi's 👀 on this one before merging since he will have to deal with the consequences longer than me

jmhooper:
> However, I think we might want to use the regular airbnb config for the front end scripts, since it includes all kinds of goodies for linting React.

Good catch. I updated it to use the latest config.

PR #929, opened 2017-05-23
Coalesce env vars that depend on the apps hostname

jmhooper:
There are a number of env vars whose values are function of the app's hostname. Traditionally, these vars have been added to the env as such:

```
FEDERALIST_BUILD_STATUS_CALLBACK='https://federalist-staging.fr.cloud.gov/v0/build/:build_id/status/:token'
```

Since the only variable in these env vars is Federalist's hostname, it makes sense to remove them and use the hostname env var to build them at runtime. This commit does the work to that end.

- Remove unused `APP_NAME` and `APP_DOMAIN` vars
- Remove build container callback vars and build URLs inline when creating message body

### DIFF
--- a/api/services/SQS.js
+++ b/api/services/SQS.js
@@ -1,11 +1,64 @@
-const AWS = require('aws-sdk')
-const logger = require("winston")
-const url = require("url")
-const config = require("../../config")
-const buildConfig = config.build
-const s3Config = config.s3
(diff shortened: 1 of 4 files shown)


jmhooper:
This PR has an inflated diff. I've started cleaning up linter errors in light of #926. To see functional changes, looks at the first commit. To see style changes, look at the second commit.

PR #930, opened 2017-05-23
Add small margin-top to usa-disclaimer-official

jseppi:
This PR adds a `margin-top: 3px` to the `.usa-disclaimer-official` class so that the top of that little banner doesn't touch the top of the browser viewport.

Before:
screen shot 2017-05-23 at 2 59 57 pm

After:
screen shot 2017-05-23 at 2 59 49 pm

### DIFF
--- a/public/styles/sass/_header.scss
+++ b/public/styles/sass/_header.scss
@@ -107,3 +107,7 @@ $logo-height: 29px;
 .header .usa-button {
   margin-top: 0;
 }
+
+.usa-disclaimer-official {
+  margin-top: 3px;
+}
--- a/public/styles/styles.css
+++ b/public/styles/styles.css
@@ -442,6 +442,9 @@ nav li a:visited {
 .header .usa-button {
   margin-top: 0; }
 
+.usa-disclaimer-official {
+  margin-top: 3px; }
+
 .hero {
   background-color: #618ec8;
   background-image: url("/images/essay-blue.png");


wslack:
Does it need to be in both places?

jseppi:
I think `styles.css` is the "built" version of the all the SASS files. It appears to be checked into the repo, so I think it does need to be there. @jmhooper would have a better handle on how that works.

That said, ideally `styles.css` would be built during CI or deployment, but that's a separate issue from this PR.

Also, the version of WDS in use here is fairly old (0.9.0). More recent versions have different "official government site" banner markup and have the text aligned properly, so upgrading WDS is another potential fix for this. Though I'm sure changing WDS versions would open its own can of worms.

jmhooper:
Yes, it looks like we should be able to move `styles.css` out of the repo. Additionally, we should probably remove the SASS files from `public` so they aren't served on federalist.fr.cloud.gov.

wslack:
If @hbillings has time this week, this might a good thing to check. In the meantime, I think we should approve/merge this.

wslack:
Let's leave the branch alone in case we should revert in favor of a more significant change.

hbillings:
Just noting this here for future reference: @jseppi is correct that the old version of the WDS uses different class names -- the disclaimer bar is now `.usa-banner-header`. It looks like Federalist is currently on the same version of the WDS that the (team template)[https://github.com/18F/federalist-modern-team-template] is on. In upgrading that, I've had to change a lot of markup, notably things like:
- `.usa-content` is mostly unused
- the navbar has a different configuration
- the disclaimer banner is different
- the footer markup is different

It might be worth trying to update the WDS in a separate branch just to see how broken the results are. If Federalist uses mostly custom markup (which I'm guessing it might, given how different parts of it look from the default), it may not be too big a pain.

PR #937, opened 2017-05-24
Move scss files out of public/ to frontend/

### LINKED ISSUE #932: Remove SASS source files from public directory
Currently the SASS source files located in `/public/styles/sass` are served from the production site. There is no need for this, so they should be moved to a different directory.

Ref: https://github.com/18F/federalist/pull/930#issuecomment-303528979

jseppi:
This will prevent our .scss source files from being served on the live site

Closes #932


jseppi:
Yeah, I kinda thought the same. However, I also plan on looking into changing over to webpack, in which case that location will be convenient since webpack will also handle the sass compiling.

jmhooper:
Gotcha. I thinks that's a good idea.

Issue #941, opened 2017-05-25
Get CodeClimate configured properly


jseppi:
Right now the repo's Code Climate score is 2.3 (out of 4). However, in quickly looking at the Code Climate [account for this project](https://codeclimate.com/github/18F/federalist/), it is reporting that the last build was 12 months ago o_O so obviously something is misconfigured.

Issue #938, opened 2017-05-24
Cleanup contents of `scripts` folder


jmhooper:
The scripts folder has some old executable files that look like they may not work or may not be used. We should go through those and clean up the ones that we use and remove scripts that are no longer needed

Issue #935, opened 2017-05-24
Excise remaining sails cruft


jseppi:
There still appears to be a lot of sails.js junk left over in the repo, mostly in comments, but also in some (I think unused) source files. I think all mentions, and certainly unused code files, should be removed to increase clarity.

<details>
<summary>Expand to see results of search for "sails" in the repo</summary>

```
$ ag sails
config/env/test.js
31:    adapter: 'sails-postgresql',

config/local.sample.js
16:      adapter: 'sails-postgresql',

Dockerfile
14:  SAILS_LOG_LEVEL= \

public/js/dependencies/sails.io.js
9: * sails.io.js
11: * JavaScript Client (SDK) for communicating with Sails.
14: * using WebSockets from the browser to talk to your Sails server.
17: * http://sailsjs.org/#!documentation/reference/BrowserSDK/BrowserSDK.html
20: * This file allows you to send and receive socket.io messages to & from Sails
54:    version: '__sails_io_sdk_version',
55:    platform: '__sails_io_sdk_platform',
56:    language: '__sails_io_sdk_language'
59:  // Current version of this SDK (sailsDK?!?!) and other metadata
79:   * to one or more Sails backend(s).  Automatically connects a socket and
81:   * is connected, the sails.io.js client will queue it up.
86:  function SailsIOClient(io) {
95:    if (!io) throw new Error('`sails.io.js` requires a socket.io client, but `io` was not passed in.');
131:        // All logs are disabled when `io.sails.environment = 'production'`.
132:        if (io.sails.environment === 'production') return;
163:     * @param  {SailsSocket}  socket
175:        // (e.g. Ember does this. See https://github.com/balderdashy/sails.io.js/pull/5)
206:      window._sailsIoJSConnect = function(response) {
219:     * The JWR (JSON WebSocket Response) received from a Sails server.
239:      return '[ResponseFromSails]' + '  -- ' +
259:     * @param  {SailsSocket} socket  [description]
276:      var sailsEndpoint = requestCtx.method;
278:      socket._raw.emit(sailsEndpoint, requestCtx, function serverResponded(responseCtx) {
298:    // when the browser connects with `io.connect()`, we create our own constructor, `SailsSocket`.
300:    // when changes are rolled out in the future.  To get a `SailsSocket`, you can run:
302:    // io.sails.connect();
308:     * SailsSocket
311:     * to the Socket.io server running inside of Sails.
[...]

PR #942, opened 2017-05-25
Excise remaining sails cruft

jseppi:
This PR removes all remaining mentions of "sails" from the repository. Mostly these deletions were lines in dotfiles.

Also:
* Cleared out `.codeclimate.yml`, which was excluding paths to things that didn't even exist. This led me to take a quick look at the project on Code Climate, where I noticed that it doesn't look to be configured properly, so I filed #941.
* ~Deleted `exportSitesAsCsv.js` and associated bash script and npm task. This file looks like it was never converted after the initial removal of sails, so I am guessing it isn't used anywhere and is thus just vestigial code. Please confirm, @jmhooper. This is related to #938 as well.~
* Added new `scripts/lint-diff.sh` and `lint:diff` task in `package.json` to locally lint changed files. `scripts/lint-travis.sh` has now been modified to use `lint-diff.sh` internally.

To Do:
- [x] figure out why `npm run test:prepare` is failing on Travis. It is not failing locally.
- [x] un-`sails`-ify `exportSitesAsCsv.js`

Ref #935


jmhooper:
The `exportSitesAsCsv.js` is actually used, though apparently it hasn't been used recently 😳 . It's used by the `download-sites-export` script: https://github.com/18F/federalist/blob/staging/scripts/download-sites-export.sh

Moreover, being able to download the list of sites to see what is on Federalist will probably be a requisite for the coming ATO.

jmhooper:
Looks good. Have we actually done this on the backend in api.data.gov?

jseppi:
Not quite sure what you mean about api.data.gov

jmhooper:
I... am in the wrong issue. Was reviewing a totally different project and apparently changed to the wrong tab 😳

jseppi:
@jmhooper I have no idea why the Travis build is failing. It says `bundle.js` is missing during `npm test:prepare` (ref https://travis-ci.org/18F/federalist/builds/236138872#L1718). Any ideas?

jmhooper:
@jseppi: It looks like the deletion of `public/js/dependencies/sails.io.js` took `public/js` with it. We'll need to add `public/js/.keep` to the repo to get it back.

jseppi:
oh geez, thanks for figuring that out!

jseppi:
Ready for review, @jmhooper!

jseppi:
@jmhooper donezo!

PR #943, opened 2017-05-25
update template URLs

wslack:


config/templates.js
```diff
@@ -3,16 +3,16 @@ module.exports = {
...
-    example: 'https://federalist-proxy.app.cloud.gov/site/18f/federalist-uswds-template/',
+    example: 'https://federalist-uswds-template.18f.gov',
     title: 'U.S. Web Design Standards',
     description: 'A flexible Jekyll implementation of the U.S. Web Design Standards',
     thumb: "/images/uswds.thumb.png",
   },
   'team': {
     owner: '18f',
-    repo: 'federalist-modern-team-template',
+    repo: 'https://federalist-modern-team-template.18f.gov',
```
jseppi: I think you changed the wrong property here. The new link should be down in `example:`, not in `repo:`.

wslack: what I get for multitasking fixing now

wslack:
@jseppi fixed stuff, check again?

config/templates.js
```diff
@@ -3,26 +3,26 @@ module.exports = {
...
     thumb: "/images/team.thumb.png",
   },
   'landing': {
     owner: '18f',
     repo: 'federalist-landing-page-template',
     branch: 'master',
-    example: 'http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/federalist-landing-page-template/',
-    title: 'Single-page template',
+    example: 'https://federalist-landing-template.18f.gov',
+    title: 'Research page template',
```
jseppi: Nit: the other `title`s are in all uppercase instead of sentence case like this one. Not sure which you/we prefer, but consistency would be good.

jseppi:
Travis is failing because you have some linting problems in `config/templates.js` (see https://travis-ci.org/18F/federalist/builds/236133443#L2396) and we're now linting any modified JS files contained in a PR.

I can quickly fix the linting bugs in this branch if you'd like.

wslack:
If my fixes didn't do it, please jump in and help out!

Issue #947, opened 2017-05-30
google-fonts.css is missing


jseppi:
The `google-fonts.css` stylesheet, referenced in `index.html`, is missing both in dev and on the current production site.

I'm not sure what it was/is or if it should be there. If it is no longer needed we should just remove the  `<link>` to it in `index.html`

<img width="484" alt="screen shot 2017-05-30 at 9 07 52 am" src="https://cloud.githubusercontent.com/assets/697848/26587281/7b976164-4517-11e7-9baf-6a652df2ea4c.png">

jmhooper:
I also have no idea what it was there for either so, 👍 to removing it

jseppi:
My guess is that it was an old uswds thing.

wslack:
SGTM

Issue #946, opened 2017-05-26
Add cache busting when bundled assets change


jmhooper:
Currently, when the bundled JS changes, a lot of browsers will cache the old JS leading to problems where they need to hard refresh. We should do something like add a digest to the request path for the assets or otherwise to bust the browser cache when the assets change.

jseppi:
This can be rolled in with the move to webpack

Issue #939, opened 2017-05-24
Switch to Webpack


jmhooper:
It would be nice to move away from Browserify towards Webpack which has more features, tooling and configuration options.

Ref #937

PR #949, opened 2017-05-30
Use webpack instead of browserify

jseppi:
This PR replaces `browserify` and all its `ify`-friends with `webpack`, which is faster, more modern, easier to understand (for me at least), and well-supported for React projects. It also can handle transforming SASS and adding hashes to built files (#946).

I was also able to clean up a number of our npm tasks, produce a better production-ready JS bundle (ref #968), and add sourcemaps to our built JS and CSS (so that we can debug more easily!).

ref #939 

To Do:
  - [x] dependencies vs devDependencies?
  - [x] remove browserify packages
  - [x] remove browserify tasks
  - [x] update build scripts
  - [x] use webpack for SASS
  - [x] add hash/digest to included scripts and css (related #946) 
  - [x] minify/etc for production release
  - [x] add new tests for main controller
  - [x] get rid of references to google-fonts.css (ref #947)
  - [x] fix sass loading to use correct path for images


package.json
```diff
@@ -24,28 +29,26 @@
...
     "urlsafe-base64": "^1.0.0",
+    "webpack": "^2.6.1",
     "whatwg-fetch": "^1.0.0",
     "winston": "^2.3.1",
     "yamljs": "^0.2.8"
   },
   "scripts": {
     "postinstall": "npm run uswds && npm run migrate:up",
-    "build": "npm run browserify && npm run sass && npm run autoprefixer",
+    "build": "webpack -p",
```
jseppi: runs webpack in production mode

package.json
```diff
@@ -24,28 +29,26 @@
...
-    "browserify": "browserify frontend/main.js -o public/js/bundle.js -t [ stringify --extensions [ .html ] ] -t [ babelify --presets [ es2015 react stage-2 ] ]",
-    "sass": "node-sass frontend/sass/styles.scss -o public/styles",
-    "autoprefixer": "postcss --use autoprefixer -r public/styles/styles.css",
     "watch": "npm run uswds && node index.js & npm run watch:client",
     "watch:server": "nodemon index.js --ignore frontend/ --ignore .tmp/ --ignore node_modules/",
-    "watch:client": "npm run watch:client:sass & npm run watch:client:autoprefix & npm run watch:client:js",
-    "watch:client:sass": "node-sass -w -r frontend/sass/ -o public/styles",
-    "watch:client:autoprefix": "npm run sass && postcss --use autoprefixer -r -w public/styles/styles.css public/styles/styles.css",
-    "watch:client:js": "watchify frontend/main.js -o public/js/bundle.js -t [ stringify --extensions [.html] ] -t [ babelify --presets [ es2015 react stage-2 ] ]",
+    "watch:client": "webpack --progress --colors --watch",
```
jseppi: this task runs webpack in non-production mode, and rebuilds assets on file changes

brandocalrissian:
This is still WIP, yes? Will hold off on approval/notes then (sifting through github notifications).

jseppi:
Sorry, yeah -- I noticed that the Main controller didn't have any tests to begin with, and I modified it here, so want to add some while I'm mucking around with it :)

jseppi:
Ready for final review, @brandocalrissian

jseppi:
Looks like the merge from `staging` went sideways somehow. I'll look into it.

jseppi:
It wasn't really the merge, but a slight change in image paths in `_alerts.scss` that was the problem. That is fixed now.

However, it is kind of a band-aid, because it made me realize that _all_ uswds image paths are going to be broken due to how the uswds assets are copied over in the `uswds` task of `package.json`. I'd like to fix that in a separate PR though so we can get this closed out (it is a problem that predates the webpack stuff here). I think a good approach would be to remove all those custom `uswds` tasks and instead use webpack to manage uswds styles and assets.

jseppi:
Build is fixed, @brandocalrissian

package.json
jseppi: ref https://webpack.js.org/guides/production-build/#the-automatic-way

wslack:
Also fixes #947

package.json
```diff
@@ -5,17 +5,22 @@
   "description": "",
   "keywords": [],
   "dependencies": {
+    "assets-webpack-plugin": "^3.5.1",
```
brandocalrissian: Not sure why this is in dependencies instead of in devDependencies.

package.json
```diff
@@ -5,17 +5,22 @@
...
     "aws-sdk": "^2.1.36",
+    "babel-loader": "^7.0.0",
+    "babel-polyfill": "^6.16.0",
     "body-parser": "^1.16.1",
     "cfenv": "^1.0.0",
     "connect-session-sequelize": "^4.1.0",
     "core-js": "^2.4.1",
     "db-migrate": "^0.9.17",
     "deep-extend": "^0.4.1",
+    "ejs": "^2.5.6",
```
brandocalrissian: Also not sure about having this in the dependencies since we only seem to be using it to js and css bundles in `index.html`.

jseppi:
@brandocalrissian: yeah, I need to get a better handle on the deps. Basically the airbnb eslint config says that anything that gets `require`d or `import`ed should be in `dependencies` instead of `devDependencies`, so I was just following along with that.

package.json
jseppi: `ejs` definitely should be in `dependencies` because it is used by the express server.

Issue #948, opened 2017-05-30
Fix React deprecation notices


jseppi:
React has deprecated a couple things we are using. Fixes should be fairly easy. The deprecation warnings are showing up both in local dev and on prod.

- PropTypes
- React.createClass

<img width="482" alt="screen shot 2017-05-30 at 9 09 57 am" src="https://cloud.githubusercontent.com/assets/697848/26587447/1055ec26-4518-11e7-8bf1-5623492fcc35.png">

jmhooper:
This is fixed in our code, but we still need to upgrade some deps to get them off deprecated React functionality.

jseppi:
Oh, sorry, it's not actually fixed in our code. I had created a branch with a fix to our code, but because that touched tons of files, which then required lots of linting fixes, I punted on it.

I think the approach should be to fix it in files as they are touched, but maybe not whole-hog.

jseppi:
Or, as discussed, go ahead and make the fix, but ignore the linting errors as long as tests pass.

wslack:
This was addressed by James!

PR #950, opened 2017-05-30
[WIP] Fix react deprecations

jseppi:
work-in-progress

Ref #948


jseppi:
Hrm, this is going to be a pain with all the linting problems. Additionally taking care of our own code will not stop the deprecation warnings since some of the packages (like `react-router`) we are using are also non-compliant.

jseppi:
Going to close this out for now -- too many linting problems, and the notices won't go away since some of our dependencies are also using the deprecated API. However, after #949 is merged, the console warnings will go away in production since we'll be using a proper production-ready js bundle.

Issue #955, opened 2017-05-31
Sites page - Add margin to left of Delete button


jseppi:
The button and the text run into each other:

<img width="365" alt="screen shot 2017-05-31 at 8 18 36 am" src="https://cloud.githubusercontent.com/assets/697848/26640441/5769d26e-45ed-11e7-9cf4-a4f244d10ef1.png">

jseppi:
On prod currently  this actually looks even worse:

<img width="531" alt="screen shot 2017-06-02 at 3 09 23 pm" src="https://cloud.githubusercontent.com/assets/697848/26742963/8aeec428-47a5-11e7-86c6-762f575ae4fd.png">

The icon is visible because the icon path for the alert class used for that div was corrected in https://github.com/18F/federalist/pull/924

wslack:
Let's fix it all!

PR #966, opened 2017-06-01
Load uswds with webpack

jseppi:
As mentioned in [this comment](https://github.com/18F/federalist/pull/949#issuecomment-305523075), USWDS styles/assets are currently managed by a series of npm tasks (listed in `package.json`). One problem that currently exists with that method is that the paths to assets (like images referenced in the uswds.css) are not correct.

This PR brings uswds into our webpack build, which:
- solves the asset paths problem (including getting rid of custom style rules to assets that were overridden)
- adds all the uswds style rules into our main styles bundle, which reduces the number of client requests
- removes the 4 uswds npm tasks, thus simplifying further our `package.json` scripts
- fixes a 404 on an asset on the `site/X/settings` view. This is related to #955.

To Do:
- [x] figure out and fix the failing `mocha` run. Potential fix in this commit: https://github.com/coryhouse/react-slingshot/commit/16ec28c9029bf7e2b65b26c22a1c2daadab427a2
- [x] fix the styling of the delete site button on the `site/X/settings` view (see screenshot below)
- [x] get the npm watch tasks to work. Some options for this:
    - *I went with this*: reload the server on asset change, because it needs to reload `webpack-manifest.json` to get the newest asset names. This will maybe slow down development since the server has to reload.
    - ~only use hashed names in production (by checking `NODE_ENV`). This might be the most "correct" thing to do, but will require changing builds a little bit.~
    - ~read `webpack-manifest.json` on each client request to the main controller. This is the easiest method.~

Now that the asset paths are fixed, I actually found an ugly design bug on `site/X/settings` view:

screen shot 2017-06-01 at 1 30 33 pm

I fixed it by removing the background-image and adding a small margin to the delete button:
screen shot 2017-06-01 at 2 02 44 pm


jseppi:
@wslack fixing the uswds image paths causes some slight visual differences on the site settings view: basically the alert icons work now. Is this ok?

In the before case, as in production currently, there are `404` errors logged for the missing icons. Whatever you decide, we should at least fix the 404 errors.

Before:

screen shot 2017-06-01 at 2 06 04 pm

After:

screen shot 2017-06-01 at 2 05 37 pm

jseppi:
@brandocalrissian ready to review! Or, if you prefer to just do one review, I can merge this into my other PR #949

wslack:
@jseppi no issues with the missing icons

jseppi:
To be clear, you mean you want the icons to not be there, correct? If so the left padding on those boxes should also be reduced in my opinion.

wslack:
We should fix the bugs but having the icons isn't a big problem to me; they don't take away much. However, if easier to remove the icons, let's just do that!

jseppi:
It's easier to leave them there now that the paths are fixed :) Happy to modify the styling to remove them though.

brandocalrissian:
I changed the base branch-- already merged the `939-webpack` branch to staging.

frontend/sass/styles.scss
```diff
@@ -1,3 +1,5 @@
+@import '../../node_modules/uswds/dist/css/uswds.css';
```
brandocalrissian: having the hardcoded path to our project root `node_modules` seems clunky, but I don't see an easy way to avoid it if we want to bring the uswds into sass-land.

frontend/components/disclaimer.js
```diff
@@ -1,13 +1,15 @@
...
+
 const Disclaimer = () =>
   <div className="usa-disclaimer">
     <div className="usa-grid">
       <span className="usa-disclaimer-official">
         <img
           className="usa-flag_icon"
           alt="US flag signifying that this is a United States Federal Government website"
-          src="/images/uswds/us_flag_small.png"
+          src={flagUrl}
```
brandocalrissian: Is this related to the travis-ci fail? I was expecting it to be from the change of base branch, but then I saw this:

```
/home/travis/build/18F/federalist/node_modules/uswds/dist/img/us_flag_small.png:1
(function (exports, require, module, __filename, __dirname) { �PNG
```

frontend/sass/styles.scss
jseppi: Yeah, I don't know of a better way. If for some reason the path changes in a future release, the webpack build will fail, so it should be "safe" to use this.

frontend/components/disclaimer.js
jseppi: I believe the failure was due to the lint task failing

jseppi: oh and also that invalid character...

jseppi: Hrm, it's apparently something with mocha not an actual problem. Trying to figure out how to make mocha not fail on it...

jseppi:
Fixed the mocha problem, @brandocalrissian. The issue was that I am using some webpack features for loading non-js assets via `require(...)`, but mocha of course doesn't know how to handle that in the same way. I've added a new file, `_mochaSetup.js`, that will be run with tests and will override `require(...)` for specific file extensions with a noop function.

jseppi:
Since @jmhooper is around this week, I think he could review instead/also.

Issue #956, opened 2017-05-31
Federalist prod PR status are not reporting, but staging is


jseppi:
We saw this on https://github.com/18F/18f.gsa.gov/pull/2442. The **production** federalist [build succeeded](https://federalist.18f.gov/sites/3/builds/6531/logs), but its PR status was not reported. The federalist **staging** build failed, due to a [build timeout](https://federalist-staging.fr.cloud.gov/sites/32/builds/1826/logs) of currently unknown cause, and its PR status did report (as failed).

cc @wslack 

To Do:
- [x] Confirm that a site with builds happening both in federalist prod and federalist staging shows two independent GitHub statuses with the correct names ("federalist/build" and "federalist-staging/build")

wslack:
Expected behavior: the build status reports in the PR, or both build statuses report.

jmhooper:
Ref #906

jseppi:
Confirmed that both status reporters are reporting!

<img width="791" alt="screen shot 2017-06-07 at 3 18 14 pm" src="https://user-images.githubusercontent.com/697848/26899408-a2e2a246-4b94-11e7-9b8d-747e4dfc553b.png">

(from one of my test repos: https://github.com/jseppi/test-federalist-site/pull/1)

wslack:
Woo hoo cc @gboone

PR #971, opened 2017-06-05
Set context name in status report based on app env

jseppi:
This PR adds a new app config value called `app_env` that is derived from the `APP_ENV` environment variable. It defaults to `'development'`. This setting is used to set the GitHub status notification "context" to a name based on that setting, such as `federalist-staging/build`. For production, we still use the nicer `federalist/built` value.

To Do:

- [x] add test for status context based on app env

Ref #956


api/services/GithubBuildStatusReporter.js
```diff
@@ -1,104 +1,106 @@
...
-    site = model
-    return loadBuildUserAccessToken(build)
-  }).then(accessToken => {
-    githubClient = authenticateGithubClient(accessToken)
+    site = model;
+    return loadBuildUserAccessToken(build);
+  }).then((accessToken) => {
+    const githubClient = authenticateGithubClient(accessToken);
+
+    const context = environment.env === 'production'
```
jseppi: There's a lot of noise in the diff due to linting, but this is the relevant change.

jmhooper: I am not sure that the `environment` service exposes that `env` var. That is used primarily to get values from the `federalist-${environment}-env` user provided service.

Instead, we may want to consider adding either an `app_env` config to `config/app.js`, or a `github_status_context` config to `config/build.js`.

jseppi:
I went with checking `process.env.APP_ENV` in GithubBuildStatusReporter.js. I had a hard time working with the tests to instead make it part of `config.app`, but I think that would be a better place to set it (and provide a default value).

jseppi:
I don't know why I was thinking I couldn't do the same thing I was doing with `process.env.APP_ENV` for `config.app.app_env`. Disregard previous comment :)

Issue #957, opened 2017-05-31
Single user has issue cloning new repos


wslack:
A specific user has been unable to add a new template on federalist.18f.gov. I have confirmed that myself and other users (inside and outside GSA) can use the templates just fine.

This user gets this error in Chrome: `{"message":"Not Found","documentation_url":"https://developer.github.com/v3"}` and this in Firefox: `You are not authorized to perform that action`
<img width="764" alt="screen shot 2017-05-31 at 6 33 39 pm" src="https://cloud.githubusercontent.com/assets/10016577/26657030/b3e700b0-462f-11e7-8871-9a0a948b5a96.png">

At some point in the past, the user was able to add a site, but the error logs show this:
```[clone.sh] Cloning site
Cloning into '.'...
fatal: Remote branch master not found in upstream origin
Unexpected end of command stream
```

What we've tried:

- Different computers at the user's home (not on a VPN)
- Different GitHub accounts
- Different browsers
- Incognito browser windows
- Hard refreshes
- Different names of repos
- Looking at console logs in the user's browser and on the Federalist site (which all show a 404 error)

We have not tried:

- Connecting from a different location
- Testing with the staging environment
- Trying to add a forked template repo manually

wslack:
Sent to user:

Two things to try:
1. Please try the workflow at our staging site (where we test new code): federalist-staging.fr.cloud.gov

2. If you don't mind, could you try manually forking one of the template repos (top right button on this page, https://github.com/18F/federalist-modern-team-template) and then trying to add it to Federalist manually? That will help us understand more about the issue.

afeld:
That sounds like they are trying to use the wrong URL or something.

wslack:
@afeld they are clicking this button in Federalist; just as others do for whom the button works:
<img width="337" alt="screen shot 2017-05-31 at 6 47 16 pm" src="https://cloud.githubusercontent.com/assets/10016577/26657401/9d183f3c-4631-11e7-9de3-51a963e59c16.png">

toolness:
Hmm, I had a sort of similar problem when trying to access my super simple project gallery page from pycon.  It just uses the github API to get a list of all the repos I've made, but it was returning a similar `You are not authorized to perform that action` response.

I had thought that perhaps it was due someone at pycon somehow abusing github, and maybe github had temporarily banned our IP or IP range from making API requests, but I'm not sure.

[...]

PR #973, opened 2017-06-05
Fix issue where Federalist was creating org sites instead of user sites

jmhooper:
When creating a repo, the API request is different for creating the repo for a user versus for an org. Because of this reason, we need to compare the repo owner string to the username to determine what kind of request to send.

Federalist downcases the repo name for a new site when creating a site from a template. Because of this, if the user has capital letters in their username, the comparison will fail where it should succeed.

For example, if a user named `FederalistUser` tries to create a template site, the owner name will be `federalistuser`. When the owner string is compared to the user's username, the comparison will fail and a request will be sent to the new org repo API endpoint, resulting in a 404 since the `federalistuser` org does not exist.

This commit fixes the issue by downcasing the username and owner before the comparison. Additionally, a regression test is added for this case.

Ref #957

### DIFF
--- a/api/services/GitHub.js
+++ b/api/services/GitHub.js
@@ -1,184 +1,162 @@
-const Github = require("github")
-const config = require("../../config")
-const { User } = require("../models")
+const Github = require('github');
+const config = require('../../config');
+const { User } = require('../models');
 
-const createRepoForOrg = (github, options) => {
-  return new Promise((resolve, reject) => {
-    github.repos.createFromOrg(options, (err, res) => {
-      if (err) {
-        err.status = err.code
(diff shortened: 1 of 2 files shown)


jmhooper:
A note to reviewers: This PR has 2 commits. The first contains the meaningful changes. The second is changes to satisfy the linter.

wslack:
Should we move this into master immediately or have Chris test on staging?

jmhooper:
@wslack: If it isn't too much to ask of Chris to test on staging, I say we ask for him to test it out there and then we merge the fix if it works

wslack:
100% ok

Issue #905, opened 2017-05-11
Consider adding a limit to the number of builds in the build log


jmhooper:
We should consider adding a limit to the number of builds displayed in the build log. Currently, it displays all of a site's builds, which can be quite a lengthy list after time. We may want to limit that to something like the most recent 100 builds.

wslack:
Concur.

wslack:
asking users for feedback in slack

wslack:
See https://gsa-tts.slack.com/archives/C0801BNJE/p1494529110052320

PR #974, opened 2017-06-06
Limit builds find action to 100 builds

jmhooper:
The build log table has started to load slowly, especially for heavier sites like 18f.gsa.gov. This commit modifies the build controller so it limits the builds to the 100 most recent builds to make that API call fire more quickly for heavy sites.

Ref #905

### DIFF
--- a/frontend/components/site/siteBuilds.js
+++ b/frontend/components/site/siteBuilds.js
@@ -1,31 +1,53 @@
 import React from 'react';
 import { Link } from 'react-router';
-import LoadingIndicator from '../loadingIndicator'
+import LoadingIndicator from '../loadingIndicator';
 import { duration, timeFrom } from '../../util/datetime';
 import buildActions from '../../actions/buildActions';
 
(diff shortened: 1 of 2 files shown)


jmhooper:
Meaningful changes are in the first commit. The second commit fixes linter errors.

jseppi:
Should we include something in the web ui to tell users that only the 100 most recent builds are shown?

jmhooper:
Perhaps if there's 100 builds, we put a paragraph at the bottom of the table saying the results may have been truncated?

jseppi:
That sounds good to me!

jmhooper:
I've pushed a commit that adds the message. As always, there's commit with meaningful changes followed by commits that fix linter issues.

jseppi:
Did you mean to close this, @jmhooper?

jmhooper:
Yeah, I rebase and merged, but didn't push the changes to this branch so it was marked closed instead of merged when I did.

PR #976, opened 2017-06-06
Add cache-control header

jmhooper:
Since the app is served in front of CloudFront, the API needs a Cache-Control header to tell CloudFront not to cache API response. This commit adds a middleware in `app.js` to make that happen.


jmhooper:
@jseppi: 👍 will update

jseppi:
It would also be great if there was a test that covered this :)

jmhooper:
@jseppi: I've added tests to check for the Cache-Control header

wslack:
@jseppi is the test 👍 ?

PR #980, opened 2017-06-06
Merging master back into staging after #979

jseppi:


jseppi:
Hrm, closing for now because we don't want to merge `staging` into `master`, which is required by the "update branch" protection we have enabled.

PR #981, opened 2017-06-07
Move front end-only deps into devDependencies

jseppi:
This PR moves our front end dependencies into the `devDependencies` section of `package.json` and disables the `import/no-extraneous-dependencies` rule in our .eslintrc. This will allow us to install only server-side packages for production, assuming that we've pre-built all the front end assets prior to deploys.

It also fixes the `license` property of package.json to be the proper license identifier (`CC0-1.0`) that we've asserted in LICENSE.md.


jseppi:
cc @wslack

jmhooper:
Cool, so we're calling the deps necessary to build / test the site dev dependencies and the deps necessary to actually run the site regular dependencies?

jseppi:
Yes, that's my plan. It should make the app images smaller for deployment, and I suppose also mean less surface area for security.

PR #982, opened 2017-06-07
Update staging manifest APP_HOSTNAME to use federalist-staging.18f.gov domain

jseppi:
Changing `APP_HOSTNAME` from `https://federalist-staging.fr.cloud.gov` to `https://federalist-staging.18f.gov` so that we can use the new CloudFront-fronted custom domain.

### DIFF
--- a/staging_manifest.yml
+++ b/staging_manifest.yml
@@ -15,7 +15,7 @@ services:
 env:
   NODE_ENV: production
   APP_ENV: staging
-  APP_HOSTNAME: https://federalist-staging.fr.cloud.gov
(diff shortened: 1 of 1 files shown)


jseppi:
I'd like some assistance from @jmhooper on the rest of this. I _think_ that the only thing that needs to be done after this is updating a few settings in our user-provided service in the `staging` space, but want to confirm.

I also noticed that our `swagger/index.yml` references `federalist-staging.fr.cloud.gov`. This seems like it should instead reference our production URL, or maybe whatever the current `APP_HOSTNAME` is, but I'm unsure of the intent of that file.

Issue #900, opened 2017-05-09
Add a mechanism for Federalist to be able to easily load and display an error or maintenance message


wslack:
Would display before the menu bar

jmhooper:
Discussed using a user provided service for this one. That may not work though because we'd want the value present at asset compilation which happens in CI.

jseppi:
Good point, @jmhooper. That might be more reason to implement https://github.com/18F/federalist/issues/958, which would allow us to set server-side context variables and have them displayed in the rendered site template.

jmhooper:
For the time being, I was planning on using EJS to render it above the content based on a UUPS

jseppi:
That sounds like it'll work out -- it'll just give us less control over the placement, but I think the very top of the page probably is where we'd want it anyway.

jmhooper:
Yeah, I'm thinking it should not incur too much debt since it will fit in well with the plan outlined in #958

PR #987, opened 2017-06-07
Add site wide error message

jmhooper:
This commit adds the ability to display a site wide error message. This can be done via the `federalist-site-wide-error` user provided service which takes the following shape:

```json
{
  "DISPLAY": true,
  "HEADING": "There has been a site wide error",
  "BODY": "This error is breaking things. <a href='www.example.gov'>Click here</a> for more info."
}
```

The error is displayed at the top of the site. The body is displayed as unescaped HTML to allow for the inclusion of links, bulleted lists, etc

Ref #900

### DIFF
--- a/api/controllers/main.js
+++ b/api/controllers/main.js
@@ -7,10 +7,16 @@ const webpackAssets = JSON.parse(fs.readFileSync(manifestPath, 'utf-8'));
 
 module.exports = {
   index(req, res) {
-    res.render('index.html', {
-      siteWideError: SiteWideErrorLoader.loadSiteWideError(),
+    const context = {
+      siteWideError: null,
       jsBundleName: webpackAssets['main.js'],
       cssBundleName: webpackAssets['main.css'],
-    });
+    };
+
+    if (req.session.authenticated) {
+      context.siteWideError = SiteWideErrorLoader.loadSiteWideError();
+    }
+
+    res.render('index.html', context);
   },
 };
--- a/api/controllers/main.js
+++ b/api/controllers/main.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const path = require('path');
+const SiteWideErrorLoader = require('../services/SiteWideErrorLoader');
 
 const manifestPath = path.join(__dirname, '..', '..', 'webpack-manifest.json');
 const webpackAssets = JSON.parse(fs.readFileSync(manifestPath, 'utf-8'));
@@ -7,6 +8,7 @@ const webpackAssets = JSON.parse(fs.readFileSync(manifestPath, 'utf-8'));
 module.exports = {
   index(req, res) {
     res.render('index.html', {
+      siteWideError: SiteWideErrorLoader.loadSiteWideError(),
       jsBundleName: webpackAssets['main.js'],
       cssBundleName: webpackAssets['main.css'],
     });
--- a/api/services/SiteWideErrorLoader.js
+++ b/api/services/SiteWideErrorLoader.js
@@ -4,14 +4,13 @@ const cfenv = require('cfenv');
 const loadSiteWideError = () => {
   const appEnv = cfenv.getAppEnv();
   const siteWideErrorEnv = appEnv.getServiceCreds('federalist-site-wide-error');
-  if (siteWideErrorEnv) {
+  if (siteWideErrorEnv && siteWideErrorEnv.HEADING && siteWideErrorEnv.BODY) {
     return {
-      display: siteWideErrorEnv.DISPLAY,
       heading: siteWideErrorEnv.HEADING,
       body: siteWideErrorEnv.BODY,
     };
   }
-  return { display: false, heading: '', body: '' };
+  return null;
 };
 
 module.exports = { loadSiteWideError };
--- a/api/services/SiteWideErrorLoader.js
+++ b/api/services/SiteWideErrorLoader.js
@@ -1,0 +1,17 @@
+const cfenv = require('cfenv');
+
+
+const loadSiteWideError = () => {
+  const appEnv = cfenv.getAppEnv();
+  const siteWideErrorEnv = appEnv.getServiceCreds('federalist-site-wide-error');
+  if (siteWideErrorEnv) {
+    return {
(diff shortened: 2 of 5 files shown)


wslack:
Can we add documentation on how to update the error message in federalist-docs as part of this message?

jseppi:
Here is how it looks, for reference:

screen shot 2017-06-08 at 10 09 30 am

jseppi:
Filed https://github.com/18F/federalist-docs/issues/87 for adding how-to

wslack:
Would it be possible to have the error only appear for logged in users? Sorry, just thinking about that.

jmhooper:
Yes. Should that be something that is configurable. For example, what happens in auth or something like that breaks?

wslack:
Oh fair: yes, it should be configurable. Also, it would be great if we could have blue alerts for some messages and yellows for others (if that's not too much work).

jmhooper:
Hmm, how about we make it only visible to authed users for now (easy to change in pinch) and open issues for both configurations?

wslack:
@jmhooper SGTM

jmhooper:
Hmm, making it appear for only logged in users is actually turning out to be more difficulty than expected.

jmhooper:
If I check whether the user is logged in to determine if I should display the banner, it does not appear after logging in until the user refreshes.

wslack:
@jmhooper that's fine; let's go with universal - no worries!

jseppi:
Hrm, that got me thinking: With this implementation, if a user visits the site when there is a site-wide warning, it will never go away until they do a page refresh.

jseppi:
I suppose a solution that will satisfy both only showing it for logged-in users, and be able to have the warning shown/dismissed/updated upon change of the user-provided service values would involve setting up an api endpoint, and then having the React app read from it via long-polling, or a socket connection, or maybe just upon route change (I think that could be done as a Redux middleware).

jmhooper:
I've figured out a way to display it for only logged in users. Commit is on the way up now.

jmhooper:
@wslack, the warning message will display until the use either refreshes or re-visits the site

Adding an API route to serve site wide errors and picking them up in the frontend sounds like a reach to cover the edge case where an alert is displayed for the duration of a user's session after we turn off the alert.

wslack:
but you don't have to clear a cache? That sounds ok then.

wslack:
I guess - if they click around in the site, the error will stay up, correct?

jmhooper:
Yeah, since they're technically still on the same page. CloudFront _shouldn't_ cache the page after the cache-control change on staging is merged.

wslack:
that's ok. As long as it doesn't stick through loading a new session

api/controllers/main.js
```diff
@@ -1,12 +1,19 @@
...
 
 module.exports = {
   index(req, res) {
+    let siteWideError = { display: false };
+    if (req.session.authenticated) {
+      siteWideError = SiteWideErrorLoader.loadSiteWideError();
+    }
+
     res.render('index.html', {
+      siteWideError,
```
jseppi: What about doing the logic for including `siteWideError` here in the controller instead of in the template. That is, something like:

```js
const siteWideError = SiteWideErrorLoader.loadSiteWideError();
const context = { jsBundleName: webpackAssets['main.js'], cssBundleName: webpackAssets['main.css'] };
if (req.isAuthenticated() && siteWideError.display) {
  context.siteWideError = siteWideError;
}
res.render('index.html', context);
```

Then, in `index.html`,  you can just check for presence of `siteWideError`.

This is somewhat stylistic, but I think I prefer it instead of having the mutable `siteWideError` variable created "partially" (ie, just with the `display` property).

api/routers/main.js
```diff
@@ -1,7 +1,8 @@
 const router = require('express').Router();
+const passport = require('../policies/passport');
 const MainController = require('../controllers/main');
 
-router.get('/', MainController.index);
-router.get('/sites(/*)?', MainController.index);
+router.get('/', passport, MainController.index);
+router.get('/sites(/*)?', passport, MainController.index);
```
jseppi: I don't think you need to include the `passport` middleware here if you use `req.isAuthenticated()` as the method to check for authentication in the controller.

api/controllers/main.js
jseppi: heck, maybe you could just get rid of the `display` property, and return `null` from `loadSiteWideError()` if the `body` (and optionally, `header`) is not present.

api/routers/main.js
jseppi: I am totally wrong about this

jmhooper: No worries. I actually think `req.session.authenticated` is available through the session middleware and not passport so we may still be able to pull it out.

jseppi: I'm actually wrong again. `req.isAuthenticated()` does exist and seems to do what we want. I was switching between projects and confused myself. Gack, many sorries!

api/controllers/main.js
jmhooper: Yeah, the display prop make sense in the binary work of error / no error, but doesn't carry over into the world of looking for errors and authenticated sessions.

jmhooper:
@jseppi: See the approval. I'll make a few of those changes you suggested and then merge.

api/routers/main.js
jmhooper: :rofl:

jmhooper:
@jseppi: I've amended the commit. Mind giving it another 👀

api/controllers/main.js
```diff
@@ -1,14 +1,22 @@
...
+const SiteWideErrorLoader = require('../services/SiteWideErrorLoader');
 
 const manifestPath = path.join(__dirname, '..', '..', 'webpack-manifest.json');
 const webpackAssets = JSON.parse(fs.readFileSync(manifestPath, 'utf-8'));
 
 module.exports = {
   index(req, res) {
-    res.render('index.html', {
+    const context = {
+      siteWideError: null,
```
jmhooper: I ended up needing this here or I'd get a `ReferenceError` from EJS when there wasn't an error to display.

Issue #990, opened 2017-06-08
Internal server error when deleting a site


jmhooper:
It looks like the server will sometimes respond with a 500 error for a request to delete a site. The log output is as follows:

```
error: Sending 500:  MalformedXML: The XML you provided was not well-formed or did not validate against our published schema
```

The stack trace reaches into the AWS SDK node modules, suggesting there is a syntax error in the request we are sending to S3 to either list or delete the contents of the S3 bucket.

jseppi:
Could it be a bug in a dependency? Versions could be misaligned from production to staging since we don't have a dependency lock file.

jmhooper:
Potentially. I think it may also be related to the contents of the S3 bucket. I'm seeing it for one of my sites and not the other.

PR #992, opened 2017-06-08
Fix issue where SiteRemover caused an internal server error

jmhooper:
This commit fixes 2 issues with the S3SiteRemover service, and adds regression test to prevent the issues in the future.

Both issues relate to how the `delete-objects` API request was sent to the S3 API.

The first issue was that an empty request (i.e. a request to delete 0 objects) would cause a the S3 API to respond with a 500 which would propagate to the S3SiteRemover. This commit adds a guard clause to the top of the `deleteObjects` function to prevent this from happening.

The second issue happened because the `delete-objects` API requests does not allow for the deletion of more than 1000 objects at a time. This commit changes `deleteObjects` function so that it returns a Promise that recursively deletes objects 1000 at a time. This means at most 1000 objects are deleted at a time and no deletion requests are happening simultaneously threatening to break AWS's rate limit.

Ref #990

### DIFF
--- a/api/services/S3SiteRemover.js
+++ b/api/services/S3SiteRemover.js
@@ -1,55 +1,70 @@
-const AWS = require("aws-sdk")
-const config = require("../../config")
+const AWS = require('aws-sdk');
+const config = require('../../config');
 
-const s3Config = config.s3
+const s3Config = config.s3;
 const s3Client = new AWS.S3({
   accessKeyId: s3Config.accessKeyId,
   secretAccessKey: s3Config.secretAccessKey,
   region: s3Config.region,
-})
+});
 
-const removeSite = site => {
-  return Promise.all([
-    getObjectsWithPrefix(`site/${site.owner}/${site.repository}`),
-    getObjectsWithPrefix(`demo/${site.owner}/${site.repository}`),
-    getObjectsWithPrefix(`preview/${site.owner}/${site.repository}`),
-  ]).then(objects => {
-    objects = [].concat.apply([], objects)
-    return deleteObjects(objects)
-  })
-}
+/**
+  Deletes the array of S3 objects passed to it.
 
-const deleteObjects = objects => {
+  Since AWS limits the number of objects that can be deleted at a time, this
+  method deletes the objects 1000 at a time. It does so recursively so each
+  group of 1000 is deleted one after the other instead of simultaneously. This
+  prevents the delete requests from breaking AWS's rate limit.
+*/
+const deleteObjects = (objects) => {
+  if (objects.length === 0) {
+    return Promise.resolve();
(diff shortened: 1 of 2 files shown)


jmhooper:
Unfortunately I did not break this one up into commits with changes to the code and then linter changes. The only method that is changed is the `deleteObjects` method in the S3SiteRemover and there are 2 new tests in the tests for that code.

test/api/unit/services/S3SiteRemover.test.js
```diff
@@ -1,82 +1,124 @@
...
+      .catch(done);
+    });
+
+    it('should not delete anything if there is nothing to delete', (done) => {
+      AWSMocks.mocks.S3.listObjects = (params, cb) => cb(null, {
+        Contents: [],
+      });
+
+      AWSMocks.mocks.S3.deleteObjects = () => {
+        throw new Error('Attempted to delete objects when there should be none to delete');
```
jseppi: explanatory comment for why this is not hit would be good

test/api/unit/services/S3SiteRemover.test.js
```diff
@@ -1,82 +1,124 @@
...
+      AWSMocks.mocks.S3.deleteObjects = (params, cb) => {
+        expect(params.Delete.Objects).to.have.length.at.most(1000);
+        deleteObjectsCallCount += 1;
+        cb();
+      };
+
+      factory.site()
+      .then(site => S3SiteRemover.removeSite(site))
+      .then(() => {
+        expect(deleteObjectsCallCount).to.equal(3);
```
jseppi: Seeing the `750` above and the the `3` here kind of threw me off. An explanatory comment about why (from reading the code it is because of real, demo, and preview sites each having the 750 objects) mocking `listObjects` to return 750 elements will result in more than a single call to `deleteObjects` would be nice.

api/services/S3SiteRemover.js
```diff
@@ -1,55 +1,62 @@
...
-  return Promise.all([
-    getObjectsWithPrefix(`site/${site.owner}/${site.repository}`),
-    getObjectsWithPrefix(`demo/${site.owner}/${site.repository}`),
-    getObjectsWithPrefix(`preview/${site.owner}/${site.repository}`),
-  ]).then(objects => {
-    objects = [].concat.apply([], objects)
-    return deleteObjects(objects)
-  })
-}
+const deleteObjects = (objects) => {
```
jseppi: explanatory comment about the 1000-object limit and how this works would be nice

jmhooper:
@jseppi: I have updated the code to contain some explanatory comments.

jseppi:
I need to get good at rebasing like you :)

PR #997, opened 2017-06-09
Fix site builds text and styles

jmhooper:
This commit fixes some style and content issue in the site builds
component.

- Re-word text about builds list being truncated
- Style restart button to look like a link

To Do:
- ~Make the Restart and Logs font look the same. @jseppi thinks it is a font-weight difference.~ Edit: it's not font-weight. All style properties look to be the same.

### DIFF
--- a/frontend/components/site/siteBuilds.js
+++ b/frontend/components/site/siteBuilds.js
@@ -30,13 +30,17 @@ class SiteBuilds extends React.Component {
   }
 
   static restartLink(build) {
+    /* eslint-disable jsx-a11y/href-no-hash */
     return (
-      <button
+      <a
+        href="#"
+        role="button"
         onClick={e => SiteBuilds.restartClicked(e, build)}
       >
         Restart
-      </button>
+      </a>
     );
+    /* eslint-enable jsx-a11y/href-no-hash */
   }
 
   componentDidMount() {
@@ -97,7 +101,7 @@ class SiteBuilds extends React.Component {
             })}
           </tbody>
         </table>
-        { this.builds().length >= 100 ? <p>Build list may have been shortened</p> : null }
+        { this.builds().length >= 100 ? <p>List only displays 100 most recent builds.</p> : null }
       </div>
     );
   }
--- a/test/frontend/components/site/siteBuilds.test.js
+++ b/test/frontend/components/site/siteBuilds.test.js
@@ -78,7 +78,7 @@ describe('<SiteBuilds/>', () => {
 
     const wrapper = shallow(<SiteBuilds {...props} />);
     expect(wrapper.find('p')).to.have.length(1);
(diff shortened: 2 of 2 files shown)


jmhooper:
The logs and restart links for the builds are just different enough to be annoying:

image

jseppi:
I'll try to figure out why different

frontend/components/site/siteBuilds.js
```diff
@@ -97,7 +98,7 @@ class SiteBuilds extends React.Component {
             })}
           </tbody>
         </table>
-        { this.builds().length >= 100 ? <p>Build list may have been shortened</p> : null }
+        { this.builds().length >= 100 ? <p>List only display 100 most recent builds.</p> : null }
```
jseppi: `display` -> `displays`

jseppi:
Oh dang, it doesn't actually appear to be any styling rules making that difference happen. And in Firefox the text looks equivalent:

screen shot 2017-06-09 at 4 33 43 pm

test/frontend/components/site/siteBuilds.test.js
```diff
@@ -78,7 +78,7 @@ describe('<SiteBuilds/>', () => {
 
     const wrapper = shallow(<SiteBuilds {...props} />);
     expect(wrapper.find('p')).to.have.length(1);
-    expect(wrapper.find('p').contains('Build list may have been shortened')).to.be.true;
+    expect(wrapper.find('p').contains('List only display 100 most recent builds.')).to.be.true;
```
jseppi: `display` -> `displays`

jseppi:
I also noticed that in `_logs.scss` we have

```css
.build-log-table td:last-child {
  word-break: break-word;
}
```

which sometimes causes the word "Restart" to have its letter broken up over lines when screen sizes are narrower. I can't tell the original intent of this rule, but I _think_ it is for the actual table of logs. It might be good to target that table specifically with this styling so that it doesn't mess up the Restart text.

jmhooper:
@jseppi: I amended the commit.

> I can't tell the original intent of this rule, but I think it is for the actual table of logs. It might be good to target that table specifically with this styling so that it doesn't mess up the Restart text.

We may want to address this in a separate PR.

jseppi:
We could punt on making the Restart link a button for now and file a separate issue for doing that, with an explanation of the font rendering problem.

jmhooper:
@jseppi: You mean reverting it to an anchor and ignoring the linter issue?

jseppi:
Yeah. Or wrapping that function with `/* eslint-disable whatever-that-rule-is */.../* eslint-enable whatever-that-rule-is */`

But you might also be able to satisfy the a11y linting by putting `role="button"` on the original anchor tag instead of using a `<button>`!

jmhooper:
Done and done. Merging....

Issue #1000, opened 2017-06-12
Investigate and possibly replace custom redux bindings with react-redux


jseppi:
I just noticed today that we aren't actually using `react-redux` but instead of some custom code to bind react and redux state. We should probably use the real bindings. 

Need to investigate why this was done and what the level-of-effort would be do use `react-redux`. My intuition is that it was done because of our use of `react-router`, which may have made using `react-redux` complicated/hard at the time. There are docs on how to do this now: http://redux.js.org/docs/advanced/UsageWithReactRouter.html

Likely benefits of using react-redux:
- Easier to use other third-party modules that assume `react-redux`
- Less of our own code to maintain and document
- Better understanding by devs due to consistency with other react+redux projects
- We'd get rid of all the `actionCreators`, which will reduce our custom code

jseppi:
It looks like we will be able to do the conversion piecemeal, which is nice. For instance, I'm able to wrap our custom `<Provider>` with the react-redux Provider (named below `<ReactReduxProvider>`):

```js
store.subscribe(() => {
  render((
    <ReactReduxProvider store={store}>
      <Provider state={{ get: store.getState }}>
        <Router history={browserHistory}>
          {routes}
        </Router>
      </Provider>
    </ReactReduxProvider>
  ), mainEl);
});
```

and I think I'll be able to use react-redux to connect components one-by-one.

jmhooper:
Discussed this with @jseppi this morning and we both agree that the initial work to remove the custom provider and use the `react-redux` provider is pretty trivial and that the custom provider can be removed piecemeal. 

Using the `react-redux` provider lets us pull in modules for things like notifications or logging from the community (saving us from writing custom code in the future). Also there's the advantage of removing code from our codebase that we'd otherwise have to maintain.

Issue #984, opened 2017-06-07
Users should be shown a message when they save their site settings


jseppi:
Show an alert, either Success or Failure, after a user saves their site settings.

PR #1004, opened 2017-06-13
Add react-notification-system notification for site updates

jseppi:
This sets up [react-notification-system](https://github.com/igorprado/react-notification-system) via [react-notification-system-redux](https://github.com/gor181/react-notification-system-redux)
and a custom redux middleware to show alerts when specific action types
occur. For now, only the `'SITE_UPDATED'` action is setup to trigger a
notification (ref #984), but the code here is extensible to easily add more notifications.

Also:
- Fix up `npm run watch` task to not restart the server on front end asset changes.
  This involved a slight modification to the main.js controller and some adjustments
  to nodemon.json. It makes the dev environment faster since the app server doesn't
  need to restart when the asset bundles change.
- Start replacing React.PropTypes with the standaline prop-types module (ref #948).
- Add react-redux and wrap our react app with its Provider. We should transition our
  custom Provider out in favor of react-redux's (ref #1000).

To Do:
- ~Try to get rid of our custom redux state provider~ Moved to PR #1005 so that this one can be separately reviewed.
- [x] Add tests for new code


jseppi:
Looks like this currently:
screen shot 2017-06-13 at 1 38 19 pm

Issue #958, opened 2017-06-01
Extract common site elements into template


jseppi:
Implementation sketch:

1. extract common site components (header, footer, etc) into a template
1. refactor the home page to use the template, and stop loading its content with React
1. refactor the single-page app to use the template
1. create a custom 401 page that uses the template and return that after an unauthorized user authenticates with GitHub

This has a few benefits that I can think of off the top of my head:
1. Easier to add additional static content or error pages (like a custom 401 page, ref #924)
1. Smaller main JavaScript bundle since it will no longer in-line the `home.html` content
1. Simplified code, which will be easier to maintain/modify, especially for the header (which right now is all done in React).
1. Improved routing. Right now I sometimes notice a slight flash of the original Home page content when signed in, before the Sites page content is loaded by the React router.

wslack:
Let's discuss in standup tomorrow!

PR #1005, opened 2017-06-13
Use react-redux instead of custom Provider

jseppi:
Here's a first round of refactoring our React app to use `react-redux` (ref #1000) in place of the custom `Provider` component we had.

Summary of changes:
- Remove our `CustomProvider` and replace it with `react-redux`'s `Provider`
- Use `react-redux`'s `connect` method to map our store's `state` to `props` of components.
- Add `redux-devtools-extension` so we can use the [Redux Devtools browser extension](http://extension.remotedev.io/)
- Adjust tests to operate on component `props` instead of store `state`
- Add additional tests for `App` component


jseppi:
@jmhooper: I think this is ready for review. Pending successful review, I'll `rebase` before a merge because this has some particularly ugly commit history as I was figuring things out.

jseppi:
Please put it through the ringer though since this is a somewhat fundamental change :)

frontend/components/app.js
```diff
@@ -1,35 +1,16 @@
...
-
-class App extends React.Component {
-  constructor(props, context) {
-    super(props, context);
-    this.state = this.getStateFromStore();
-  }
-
+export class App extends React.Component {
   componentWillReceiveProps(nextProps) {
-    const state = this.getStateFromStore();
```
jmhooper: I'll be happy when we can move everything over to our notification lib and all of this code can go away.

frontend/components/app.js
```diff
@@ -46,48 +27,75 @@ class App extends React.Component {
   }
 
   render() {
-    const { children } = this.props;
-    const storeState = this.state;
+    const { user, children, notifications } = this.props;
 
-    if (storeState.user.isLoading) {
+    if (!user || user.isLoading) {
```
jmhooper: The [user reducer sets the state to `false` if the user fetch fails](https://github.com/18F/federalist/blob/staging/frontend/reducers/user.js#L17-L19). This means that for an unauthenticated user, `this.props.user` will be false.

When there is an authed user, the `App` component renders the `Home` component at its index. The `Home` component [redirects to the site list when it finds a logged in user](https://github.com/18F/federalist/blob/staging/frontend/components/home.js#L18). 

But, without an authed user, the `App` component expects to render the `Home` component, and for the `Home` component to display our fancy marketing content. It looks like this change breaks that by having the App component perpetually render the loading indicator if the user is not logged in.

jseppi: 👀

jseppi: oof, good catch. I think this is another good case for https://github.com/18F/federalist/issues/958

jseppi: I'll fix it 'as-is' here though, to keep this PR from ballooning

jseppi:
@jmhooper - I fixed the bug you found and added tests to cover those conditions

frontend/components/app.js
```diff
@@ -46,48 +27,75 @@ class App extends React.Component {
   }
 
   render() {
-    const { children } = this.props;
-    const storeState = this.state;
+    const { user, children, notifications } = this.props;
 
-    if (storeState.user.isLoading) {
+    if (user.isLoading) {
```
jmhooper: Safe to assume that `user` is never undefined? I can't seem to make it break so I think so.

jseppi: the user reducer sets its initial state to so it will never be undefined

jseppi:
I felt the same thing about the apparent quickness with this change!

I'll rebase this PRs commit history to be cleaner, then :shipit:

wslack:
it does feel snappier.

Issue #969, opened 2017-06-02
Evaluate effort to add Hugo support


wslack:
Per https://github.com/18F/federalist/issues/756#issuecomment-305671183

> As we move DigitalGov over to Federalist, we would like to formally request that the Federalist Team consider supporting HUGO as a build option.

> We currently have around 1550 blog posts, 120 pages, 300 event pages, and a few hundred additional collection pages.

> This week, our team met with one of the members of the HUGO team here in New York. His name is Bud Parr. It became increasingly clear that in order to move DigitalGov to Federalist, we are going to need the key component Hugo offers — speed.

> The tradeoffs between Jekyll and Hugo are slim, and the community around HUGO appears to be growing. It is also nice to be able to reach out to people on the core team with questions, should we get stuck.

To Do for a minimal implementation of this feature:

- [x] Finish up and merge https://github.com/18F/federalist-docker-build/pull/51 to staging 
- [x] Modify web UI to add `hugo` to build engine dropdown(s): See PR #1010.
- ~Add database migration to include `hugo`~ Not needed: `hugo` is already an enumerated value in the [site model](https://github.com/18F/federalist/blob/staging/api/models/site.js#L97)

wslack:
@jseppi to investigate level of effort

jseppi:
Self note: will need to be added in https://github.com/18F/federalist-docker-build/blob/staging/build.sh

jseppi:
Which version of hugo? Latest is currently `0.21`. Is there some way to specify a hugo version (like how a Gemfile can be used to specify a jekyll version?)? cc @jeremyzilar

jseppi:
Started a PR over in https://github.com/18F/federalist-docker-build/pull/51

jseppi:
Will also need to make some (I think small) modifications to the federalist app to allow selection of Hugo as the site engine.

jseppi:
A somewhat large concern I have is that `hugo` releases seem to be often (it does not yet have a stable 1.0 release), and can break on sites that were working on previous versions. This could be a big maintenance problem for Federalist. 

Would it be up to Federalist to keep up with the most current version, or should we have some way to let customers specify the version used? If we use the most current version and a site build breaks, then we'll have to effectively communicate that problem to our end users.

jeremyzilar:
Yeah, that is what they do for Nelify CMS — https://hugodocs.info/hosting-and-deployment/hosting-on-netlify/#building-with-a-specific-hugo-version

[...]

Issue #779, opened 2017-02-25
Support yarn?


shawnbot:
[Yarn](https://yarnpkg.com/en/) is a new, npm-backed package manager for Node that aims to be faster, more secure, and more reliable for installing dependencies than the `npm` CLI.

Given that projects with lots of churn (frequent pushes) and/or lengthy build times can cause other builds to queue for long periods of time, and that npm has been known to exhaust memory on some build containers, it might be worth investigating detection of `yarn.lock` files and running `yarn install` instead of `npm install`.

wslack:
We want to watch Yarn and see if it has staying power before prioritizing; will go in backlog for now. thank you!

jseppi:
I'm 👍 on `yarn` support for what it's worth. I'm not sure if it will be possible to get all the speed improvements of yarn, a lot of which comes from how it caches packages, because of how Federalist's build system works.

I also think we should switch this project to using yarn for development/CI. I tested it out and it works as a drop-in replacement for `npm` on this project. But that's a separate issue :)

jseppi:
This is about using yarn in the build container. Ie, detecting if a client site has a yarn.lock file and if so, installing their deps with yarn.

wslack:
Should it get higher priority? Will leave alone for now.

wslack:
Low as people can already install the NPM package.

apburnes:
Stale

PR #1010, opened 2017-06-15
Add ability to select hugo as a site's build engine

jseppi:
This PR adds hugo as a build engine option 🎉 

Also:
- Extract SelectSiteEngine component since it is used in two places
- Add test for new SelectSiteEngine component

ref #969 

Functional code changes are in the first commit; linter fixes are in the second.

### DIFF
--- a/frontend/components/AddSite/index.js
+++ b/frontend/components/AddSite/index.js
@@ -13,17 +13,19 @@ const propTypes = {
   storeState: React.PropTypes.shape({
     user: React.PropTypes.shape({
       username: React.PropTypes.string,
-      id: React.PropTypes.number
+      id: React.PropTypes.number,
     }),
-    error: React.PropTypes.string
-  })
-}
+    error: React.PropTypes.string,
+  }),
+};
+
+const defaultProps = {
+  storeState: null,
+};
 
 class AddSite extends React.Component {
   constructor(props) {
     super(props);
-
-    const { user } = props.storeState;
 
     this.state = {
       owner: this.defaultOwner(),
@@ -34,15 +36,6 @@ class AddSite extends React.Component {
 
     this.onChange = this.onChange.bind(this);
     this.onSubmit = this.onSubmit.bind(this);
-  }
-
-  defaultOwner() {
-    const userState = this.props.storeState.user
-    if (userState.data) {
-      return userState.data.username
-    } else {
-      return ""
-    }
   }
 
   onSubmit(event) {
@@ -62,6 +55,14 @@ class AddSite extends React.Component {
     this.setState(nextState);
   }
 
+  defaultOwner() {
+    const userState = this.props.storeState.user;
+    if (userState.data) {
+      return userState.data.username;
+    }
+    return '';
+  }
+
   render() {
     return (
       <div>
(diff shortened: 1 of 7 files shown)


frontend/components/AddSite/index.js
```diff
@@ -35,21 +38,12 @@ class AddSite extends React.Component {
...
-    }
-  }
-
   onSubmit(event) {
     event.preventDefault();
     siteActions.addSite(this.state);
   }
 
-  onSubmitTemplate(site) {
+  onSubmitTemplate(site) { // eslint-disable-line class-methods-use-this
```
jmhooper: Why did you ignore the rule here? I ran into it in a different component, it'd be a nice one to disable globally.

frontend/components/AddSite/index.js
```diff
@@ -108,20 +113,11 @@ class AddSite extends React.Component {
...
-                  name="engine"
-                  id="engine"
-                  className="form-control"
-                  value={this.state.engine}
-                  onChange={this.onChange}
-                >
-                  <option value="jekyll">Jekyll</option>
-                  <option value="static">Static (just publish the files in the repository)</option>
-                </select>
+                <SelectSiteEngine value={this.state.engine} onChange={this.onChange} />
```
jmhooper: 👍

frontend/components/selectSiteEngine.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const availableEngines = [
+  {
+    label: 'Jekyll',
+    value: 'jekyll',
+  },
+  {
+    label: 'Hugo',
```
jmhooper: Maybe we could change the label to `Hugo (experimental)` or something like that?

frontend/components/site/siteSettings.js
```diff
@@ -89,16 +90,7 @@ class SiteSettings extends React.Component {
...
-            <select
-              name="engine"
-              className="form-control"
-              value={state.engine}
-              onChange={this.onChange}
-            >
-              <option value="jekyll">Jekyll</option>
-              <option value="static">Static (just publish the files in the repository)</option>
-            </select>
+            <SelectSiteEngine value={state.engine} onChange={this.onChange} />
```
jmhooper: 👍 

This is so nice. This component is way overcrowded.

frontend/components/AddSite/index.js
jseppi: I ignored it mostly so I wouldn't have to change the test, which relies on it being an instance method. I don't know...maybe we should just disable it.

frontend/components/selectSiteEngine.jsx
jseppi: Sounds good to me. @wslack?

wslack: sure!

jseppi:
Updated the hugo label and globally disabled `class-methods-use-this` in our `.eslintrc`

jseppi:
oh snap, looks like `npm install` failed in CI. We might need to lock down our React and React-related dependencies in `package.json`. Better solution would be to use a lockfile (ref #779).

jmhooper:
We have an issue for a lockfile. +1 to that

jseppi:
I restarted the Travis build and it seems to be working. We may have just been in-between react releases (v15.6.0 was rolling out yesterday and today) at that exact moment.

Issue #1015, opened 2017-06-19
Site Files Table can have blank file names


jseppi:
<img width="826" alt="screen shot 2017-06-19 at 2 03 02 pm" src="https://user-images.githubusercontent.com/697848/27301229-1bd31f8e-54f8-11e7-8f4b-e66f63b1f975.png">

Need to investigate further, but I think those might be redirect objects. @jmhooper might know off the top of his head.

jmhooper:
Yep, that would be the route redirect object.

jseppi:
Ok, seems like we could just filter those out from the table display.

wslack:
Likely not worth immediate attention, I think?

jseppi:
I'm already doing stuff with tables, so I'll fix it. It's just one line: `.filter(f => !!f.name)`

PR #1014, opened 2017-06-19
Add download of plaintext build logs

jseppi:
- Implement api method and route to download plaintext build logs (via `format=text` query param on existing build log api route) (ref #983).
- Exclude build token from build and buildLog api responses (ref #1013). 
- Extract `SiteBuildLogTable` component from `SiteBuildLogs` component.
- Add download link for plaintext logs to `SiteBuildLogs` component.
- Disable `"no-throw-literal"` eslint rule since it will require a large change to how we craft error responses.
- Lint touched files.
- Fix log table styles (ref PR #1016).

To Do:
- [x] fix tests

UI looks like:

screen shot 2017-06-19 at 10 46 32 am

Downloaded file looks like:

```txt
Source: clone.sh
Timestamp: 2017-06-08T20:48:51.308Z
Output:
[clone.sh] Cloning site
Cloning into '.'...
[clone.sh] Cleaning up

Source: build.sh
Timestamp: 2017-06-08T20:53:03.000Z
Output:
[build.sh] Initializing RVM
/usr/local/nvm/versions/node/v4.8.3/bin:/usr/local/rvm/gems/ruby-2.3.1/bin:/usr/local/rvm/gems/ruby-2.3.1@global/bin:/usr/local/rvm/rubies/ruby-2.3.1/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/rvm/bin
[build.sh] Using ruby version in .ruby-version
Already installed ruby-2.3.1.
To reinstall use:

    rvm reinstall ruby-2.3.1

[build.sh] Ruby version: ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
...
```

Issue #844, opened 2017-04-04
Build Log Table can't handle very long/gross error message and maintain width


wslack:
See how the window doesn't display the right side of the table?
![screen shot 2017-04-04 at 3 40 06 pm](https://cloud.githubusercontent.com/assets/10016577/24675401/002a4e6a-194d-11e7-828c-02207d0dfd1b.png)

PR #1016, opened 2017-06-19
Improve styling of log tables to not overrun the page width

jseppi:
**Open against branch `983-download-logs` (PR https://github.com/18F/federalist/pull/1014)**

This PR adds some new table styling classes and rules so that tables with pre-formatted content do not cause overruns of the page width. Ref #844.

Also filter out files with no names, which are redirect objects, from the `SitePublishedFilesTable` (ref #1015).

Build Table looks like this now:

screen shot 2017-06-19 at 3 07 26 pm

Rows that have long messages look like this, and are scrollable:

screen shot 2017-06-19 at 3 07 53 pm

Build Log Table looks like this now, with scrollable "Output" cells:

screen shot 2017-06-19 at 3 38 07 pm

### DIFF
--- a/frontend/sass/_tables.scss
+++ b/frontend/sass/_tables.scss
@@ -1,0 +1,16 @@
+
+.log-table {
+  font-size: smaller;
+
+  pre {
+    overflow: auto;
+    white-space: pre-wrap;
+    margin: 0;
+    max-height: 200px;
+    max-width: 200px;
+  }
+}
+
+.log-table__site-build-log pre {
+  max-width: 400px;
+}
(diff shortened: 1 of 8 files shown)


frontend/sass/_tables.scss
```diff
@@ -0,0 +1,16 @@
...
+    overflow: auto;
+    white-space: pre-wrap;
+    margin: 0;
+    max-height: 200px;
+    max-width: 200px;
+  }
+}
+
+.log-table__site-build-log pre {
+  max-width: 400px;
```
jmhooper: How did you land on these widths?

jseppi: uh, I just fiddled until I thought they looked ok.

jmhooper: Nice, that works I guess

PR #1019, opened 2017-06-20
Switch from npm to yarn for managing dependencies

jseppi:
- Lock dependencies using `yarn.lock`
- Update all references to `npm` with `yarn` equivalents
- Switch Travis to using `yarn`
- Delete `build-on-ci.sh`, which is unused

To Do:
- [x] Update https://federalist-docs.18f.gov/pages/how-federalist-works/running-federalist-locally/ to include `yarn` as a dependency and change all `npm` commands to `yarn`. ref: https://github.com/18F/federalist-docs/pull/95


.travis.yml
```diff
@@ -33,8 +42,8 @@ before_deploy:
   - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.22.2"
   - tar xzvf $HOME/cf.tgz -C $HOME
   - travis_retry cf install-plugin autopilot -f -r CF-Community
-  - npm install
-  - npm run build
+  - yarn --pure-lockfile
```
jseppi: `--pure-lockfile` means it will fail if `yarn.lock` is not present

scripts/build-on-ci.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
```
jseppi: This file seems completely unused, so I removed it rather than updating references of `npm` to `yarn`. Ref #938.

package.json
```diff
@@ -34,16 +34,16 @@
     "yamljs": "^0.2.8"
   },
   "scripts": {
-    "postinstall": "npm run migrate:up",
+    "postinstall": "yarn migrate:up",
     "build": "webpack -p",
     "start": "node index.js",
     "debug": "node debug index.js",
     "watch": "NODE_ENV=development concurrently --kill-others --raw \"nodemon index.js --config nodemon.json\" \"webpack --progress --colors --watch\"",
-    "test": "npm run test:prepare && npm run test:server && npm run test:client",
```
jseppi: rather than updating all these "internal" uses of `npm run`, I thought about using something like https://github.com/camacho/yarn-or-npm to use whichever was available. It ended up feeling like unnecessary overhead so I opted to just change everything to plain ole `yarn`.

scripts/exportSitesAsCsv.js
```diff
@@ -43,9 +43,9 @@ function consolidateOnSiteId(sites) {
 const args = Array.prototype.slice.call(process.argv);
 const destination = resolveDestination(args[2] || './current-sites.csv');
 console.log('Final output can be found at', destination);
-console.log('\tUse npm run export:sites -- /other/path/file.csv to change');
+console.log('\tUse yarn export:sites -- /other/path/file.csv to change');
 
-return User.findAll({ include: [Site] })
```
jseppi: linter caught this errant `return` outside of a function.

.travis.yml
```diff
@@ -11,16 +12,24 @@ services:
   - postgresql
 before_install:
   - rvm install ruby-1.9.3-p551
+  # Repo for Yarn
+  - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
+  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+  - sudo apt-get update -qq
+  - sudo apt-get install -y -qq yarn=$YARN_VERSION
```
jmhooper: This [Travis blog post seems](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn) to suggest that Yarn is pre-installed for you if you have a `yarn.lock`

README.md
```diff
@@ -62,18 +62,20 @@ This file's usage is discussed in length in "Building the Server" below.
 
 ### Build the server
 
+This project uses [`yarn`](https://yarnpkg.com/) for managing JavaScript dependencies and running tasks specified in `package.json`.
```
jmhooper: 💪

package.json
jmhooper: :+1:

jmhooper:
It's a bit of a bummer that my review comments on a given line don't show the @jseppi's comments on the same line so I come across as an emoji madman.

.travis.yml
jseppi: Travis does have yarn pre-installed, but it is an old version and I don't think there is a way to specify a newer one. I got these install script lines from https://yarnpkg.com/en/docs/install-ci#travis-tab

Unfortunately, that means `yarn` is installed twice with every Travis run. But it doesn't add that much to the build time, and we'll be switching to Circle soon enough so I think we shouldn't worry too much about it.

jmhooper: Got it

PR #1021, opened 2017-06-20
Fix newrelic.js to pull config vals from user-provided-service env

jseppi:
**This is currently open against PR #1019**

Also:
- delete NEW_RELIC_APP_NAME from manifest.yml because it should be in the user-provided service instead
- make small modification to logging line in app.js
- upgrade newrelic to latest version

### DIFF
--- a/app.js
+++ b/app.js
@@ -10,7 +10,7 @@ logger.add(logger.transports.Console, { colorize: true });
 const env = require('./services/environment.js')();
 
 if (env.NEW_RELIC_APP_NAME && env.NEW_RELIC_LICENSE_KEY) {
-  logger.info('Activating New Relic: ', env.NEW_RELIC_APP_NAME);
+  logger.info(`Activating New Relic: ${env.NEW_RELIC_APP_NAME}`);
   require('newrelic'); // eslint-disable-line global-require
 } else {
   logger.warn('Skipping New Relic Activation');
--- a/manifest.yml
+++ b/manifest.yml
@@ -13,7 +13,6 @@ services:
 - federalist-production-env
 - federalist-site-wide-error
 env:
-  NEW_RELIC_APP_NAME: federalist-prod
   NODE_ENV: production
   APP_ENV: production
   APP_HOSTNAME: https://federalist.18f.gov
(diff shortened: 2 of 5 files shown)


jseppi:
I've also updated staging's `federalist-staging-env` service to include the required New Relic config vars so that once this gets merged down to `staging` the monitoring will come online.

The production `federalist-production-env` service already had the config values.

jseppi:
Oh, also, there's some unused apps listed in our New Relic account that can/should be deleted (including one new one I made to test locally):

screen shot 2017-06-20 at 3 36 39 pm

All of those except `federalist-prod` should be deleted, but I am not an admin on the Federalist New Relic account, so I can't do it. @wslack: are you an admin?

wslack:
I can't delete the one @jseppi made and I think we should continue to monitor the builder absent other guidance, but the others are gone!

jseppi:
Thanks! Filed https://github.com/18F/federalist-builder/issues/28 to track it in `federalist-builder`

jseppi:
After merging, we should be able to activate the New Relic Browser monitor via the "Add More" button at https://rpm.newrelic.com/accounts/1078427/browser

manifest.yml
```diff
@@ -13,7 +13,6 @@ services:
 - federalist-production-env
 - federalist-site-wide-error
 env:
-  NEW_RELIC_APP_NAME: federalist-prod
```
jseppi: This isn't used; rather, the value comes from the `federalist-{space}-env` user-provided service.

app.js
```diff
@@ -10,7 +10,7 @@ logger.add(logger.transports.Console, { colorize: true });
 const env = require('./services/environment.js')();
 
 if (env.NEW_RELIC_APP_NAME && env.NEW_RELIC_LICENSE_KEY) {
-  logger.info('Activating New Relic: ', env.NEW_RELIC_APP_NAME);
+  logger.info(`Activating New Relic: ${env.NEW_RELIC_APP_NAME}`);
```
jmhooper: Fancy

wslack:
Ok - James is now an admin

jseppi:
Thanks for Admin!

You can see now in the Federalist New Relic APM page that `federalist-staging` is now reporting data: https://rpm.newrelic.com/accounts/1078427/applications/66104527

PR #1025, opened 2017-06-21
rm unecessary call to `yarn --pure-lockfile` in before_deploy

jseppi:
There's no need to reinstall packages in the `before_deploy` section of the Travis build, and that second call is causing a build error due to some weirdness between Travis+node-sass+yarn+yarn cache, as described in https://github.com/sass/node-sass/issues/1579 and https://github.com/yarnpkg/yarn/issues/3485.

I _think_ removing that second installation will clear up the [`ENOENT` error on deploys from Travis](https://travis-ci.org/18F/federalist/builds/245402569#L1423), and it isn't necessary anyways since all the deps are already installed up in the `install` section of the Travis build.


jseppi:
🤞

jseppi:
Sweet, it worked 💃

PR #1036, opened 2017-06-26
Delete build token in build model toJSON method

jseppi:
Also, updated BuildLog swagger definition to make sure token is not included.

### DIFF
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -1,110 +1,105 @@
-const crypto = require("crypto")
-const URLSafeBase64 = require('urlsafe-base64')
-const SQS = require("../services/SQS")
+const crypto = require('crypto');
+const URLSafeBase64 = require('urlsafe-base64');
+const SQS = require('../services/SQS');
 
 const afterCreate = (build) => {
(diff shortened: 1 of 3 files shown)


api/models/build.js
```diff
@@ -1,110 +1,105 @@
...
-  Object.keys(object).forEach(key => {
+  Object.keys(object).forEach((key) => {
     if (object[key] === null) {
-      delete object[key]
+      delete object[key];
     }
-  })
-  return object
+  });
+  delete object.token;
```
jseppi: Here's the relevant change

api/serializers/build.js
```diff
@@ -6,7 +6,6 @@ function serializeObject(build) {
   json.site = build.Site.toJSON();
   delete json.User;
   delete json.Site;
-  delete json.token;
```
jseppi: not needed here since it is taken care of in `toJSON()`

jseppi:
Yup, I locally tested that the validation worked. The error it gave was a little weird, but it caught the `token` being in the response before I added the `delete object.token` line.

PR #1038, opened 2017-06-28
Remove moment.js locales and add webpack-bundle-analyzer

jseppi:
This gets rid of moment.js locales from our JS bundle, which saves us almost 200Kb from our bundle (`649K` before,  `461K` after)!

I also added the `webpack-bundle-analyzer` plugin, which generates a stats.json file that can be viewed using `yarn analyze-webpack` to see what comprises the bundle size. Neat! Results look like this:

screen shot 2017-06-28 at 8 46 58 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ public/styles/styles.*.css*
 public/styles/uswds.*.css*
 webpack-manifest.json
 webpackAssets
+public/stats.json
 
 # test coverage reports
 coverage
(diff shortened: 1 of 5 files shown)


jseppi:
`moment.js` still is a very large dependency, we might want to look into (`date-fns`)[https://date-fns.org/] as a replacement sometime in the future.

jmhooper:
That's neat!

Should we add docs on how to do this kind of analysis, or is the tooling standard enough that we expect people to know how to use it? My thinking is the script in the package.json is enough of a signal.

jseppi:
I think I rebased this properly... :P

Issue #1041, opened 2017-06-29
Use rel="noopener noreferrer" on any target="_blank" links to prevent tabnabbing


jseppi:
As reported by `sp1d3rs` in https://hackerone.com/bugs?report_id=244413&subject=tts-private:

> When the user opens a link in a new tab (target="_blank"), without rel="noopener noreferrer" attribute, the page that opens in a new tab can access the initial (parent) tab and change it's location using the window.opener property.

PR #1042, opened 2017-06-29
Use rel="noopener noreferrer" on all _blank links

jseppi:
Mitigates a low-impact security issue where a linked page could "tabnab" from the source domain. Ref #1041.

### DIFF
--- a/frontend/components/nav.js
+++ b/frontend/components/nav.js
@@ -1,45 +1,48 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Link } from 'react-router';
+import shortid from 'shortid';
 
 const propTypes = {
-  username: React.PropTypes.oneOfType([
-    React.PropTypes.string,
-    React.PropTypes.object
-  ])
+  username: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object,
+  ]),
 };
 
-const getSecondaryNav = (username) => {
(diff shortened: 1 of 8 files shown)


jseppi:
There a good amount of "noise" in the PR since I also fixed linting errors all the files that were touched. The relevant change is that anywhere where `_blank` was used now also includes `rel="noopener noreferrer"`

frontend/components/nav.js
```diff
@@ -1,45 +1,48 @@
...
+  return [
+    <a href="https://federalist-docs.18f.gov" target="_blank" rel="noopener noreferrer">Documentation</a>,
+    <a href="https://github.com/18F/federalist/issues/new" target="_blank" rel="noopener noreferrer">Contact Us</a>,
+    <a href="/auth/github">Log in</a>,
+  ];
 }
 
+function getSecondaryNav(username) {
+  const links = getSecondaryNavLinks(username);
+  return links.map(link => <li key={shortid.generate()}>{link}</li>);
```
jseppi: using `shortid` to generate unique keys for the links, which was something the linter caught. Alternatively, we could just include the surrounding `<li>...</li>` tags in the arrays of anchors up in `getSecondaryNavLinks`.

jseppi: wait, we don't even need `key` on those. Removing.

Issue #1039, opened 2017-06-28
Add refresh button to build and build log pages to refresh data without a full page reload


jseppi:
This is an alternative, or more near-term, solution for https://github.com/18F/federalist/issues/886

My thought is that we could have a refresh button present on the the build and build logs pages that would cause a re-fetch of the API data for those pages, thus updating the page state with the newest data without requiring the user to do a browser refresh. Just updating the state will be a faster and better user experience. We could even show a nice toasty message upon the refresh success (or failure).

PR #1043, opened 2017-06-29
Implement refresh buttons on Builds and Build Logs pages

jseppi:
Ref #1039

Looks like this: **Edit** I added the `usa-button` class to the buttons so they are square instead of having the rounded edges like in the screenshots below, so imagine them like that :)

screen shot 2017-06-29 at 3 35 41 pm

screen shot 2017-06-29 at 3 35 31 pm

screen shot 2017-06-29 at 3 35 57 pm

### DIFF
--- a/frontend/components/site/refreshBuildLogsButton.jsx
+++ b/frontend/components/site/refreshBuildLogsButton.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import autoBind from 'react-autobind';
+
+import buildLogActions from '../../actions/buildLogActions';
+
+
+class RefreshBuildLogsButton extends React.Component {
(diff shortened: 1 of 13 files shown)


frontend/components/site/refreshBuildLogsButton.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import autoBind from 'react-autobind';
```
jmhooper: I think we still need to add `react-autobind` to the `package.json`

jseppi: whoops, must have screwed that up on a rebase or something.

jseppi:
Build is fixed, @jmhooper

PR #1045, opened 2017-06-30
Remove unused image-manager styles

jseppi:
These styling rules appear to be remnants of a long-gone feature, so there is no reason to keep them around.


wslack:
Back when GitHub had no file upload feature.

PR #1048, opened 2017-07-03
remove org access from SBST team

wslack:
Given our new method for managing access, this was technical debt.


jseppi:
I can push a commit to this PR with the linting fixes if you'd like.

wslack:
Users are already added to `federalist-users`.

wslack:
@jseppi please do

Issue #1047, opened 2017-07-03
Stored XSS possible with domain and demoDomain fields


jseppi:
Reported by `sp1d3rs` in https://hackerone.com/bugs?subject=tts-private&report_id=245172.

The `domain` and `demoDomain` fields are not properly validated and can be used as part of a stored XSS attack because their values are used directly as the `href` attribute of links (the View Website button and the demo branch View link).

> ### POC & Reproduction steps
> 
> 1. Login to the federalist and go to the some instance http://localhost:1337/sites/<siteid>/settings
> 2. Fill the Custom Domain field by the `javascript:alert(document.domain)`
> and Demo domain `javascript:alert(document.domain);`
> (it cannot be the same so we bypass the check by adding ;)
> 3. Save and press View Website button. You will be XSSed.
> 4. Go to the http://localhost:1337/sites/<siteid>/published - and press view on the demo site to test second Stored XSS

PR #1049, opened 2017-07-03
Validate that domain and demoDomain are urls that start with https://

jseppi:
These changes are a mitigation of an XSS vulnerability described in #1047

- Use url-type inputs for domain and demoDomain on the site settings page, and check that they start with **`https://`**
- Add url validation to site model's domain and demoDomain fields
- ~Ensure that no invalid urls are returned in the site model's toJSON method~
- Add a migration to update all `domain` and `demoDomains` so that they start with `"https://"`
- Upgrade db-migrate package
- Add/modify tests

To Do:
- [x] only accept `https://` URLs
- [x] add a migration that cleans existing `domain` and `demoDomain` fields.
- [x] remove newly added logic from `Site.toJSON` since it will be superseded by the above change.

### DIFF
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -44,22 +44,6 @@ const beforeValidate = (site) => {
   }
 };
 
-
-function isEmptyOrUrl(value, throws = true) {
-  const validUrlOptions = {
-    require_protocol: true,
-    protocols: ['http', 'https'],
-  };
-
-  if (value && value.length && !validator.isURL(value, validUrlOptions)) {
-    if (throws) {
-      throw new Error('URL must start with http:// or https://');
-    }
-    return false;
-  }
-  return true;
-}
-
 function toJSON() {
   const object = Object.assign({}, this.get({
     plain: true,
@@ -67,16 +51,6 @@ function toJSON() {
 
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
-
-  // prevent any invalid domain or demoDomain URLs from being returned
-  // or used in the creation of viewLink or demoViewLink
-  if (!isEmptyOrUrl(object.domain, false)) {
-    object.domain = null;
-  }
-
-  if (!isEmptyOrUrl(object.demoDomain, false)) {
-    object.demoDomain = null;
-  }
 
   const s3Config = config.s3;
   object.siteRoot = `http://${s3Config.bucket}.s3-website-${s3Config.region}.amazonaws.com`;
@@ -107,6 +81,17 @@ function viewLinkForBranch(branch) {
     return `${s3Root}/demo/${this.owner}/${this.repository}`;
   }
   return url.resolve(config.app.preview_hostname, `/preview/${this.owner}/${this.repository}/${branch}`);
+}
+
+function isEmptyOrUrl(value) {
+  const validUrlOptions = {
+    require_protocol: true,
+    protocols: ['https'],
+  };
+
+  if (value && value.length && !validator.isURL(value, validUrlOptions)) {
+    throw new Error('URL must start with https://');
+  }
 }
 
 module.exports = (sequelize, DataTypes) => {
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -1,4 +1,6 @@
 const url = require('url');
+const validator = require('validator');
+
 const config = require('../../config');
 
 const afterValidate = (site) => {
@@ -12,6 +14,14 @@ const afterValidate = (site) => {
     error.status = 403;
     throw error;
   }
+};
+
+const validationFailed = (site, options, validationError) => {
+  const messages = validationError.errors.map(err => `${err.path}: ${err.message}`);
+
+  const error = new Error(messages.join('\n'));
+  error.status = 403;
+  throw error;
 };
 
 const associate = ({ Site, Build, User }) => {
@@ -34,6 +44,22 @@ const beforeValidate = (site) => {
   }
 };
 
+
+function isEmptyOrUrl(value, throws = true) {
+  const validUrlOptions = {
+    require_protocol: true,
+    protocols: ['http', 'https'],
+  };
+
+  if (value && value.length && !validator.isURL(value, validUrlOptions)) {
+    if (throws) {
+      throw new Error('URL must start with http:// or https://');
+    }
+    return false;
+  }
+  return true;
+}
+
 function toJSON() {
   const object = Object.assign({}, this.get({
     plain: true,
@@ -41,6 +67,16 @@ function toJSON() {
 
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
+
+  // prevent any invalid domain or demoDomain URLs from being returned
+  // or used in the creation of viewLink or demoViewLink
+  if (!isEmptyOrUrl(object.domain, false)) {
+    object.domain = null;
+  }
+
+  if (!isEmptyOrUrl(object.demoDomain, false)) {
+    object.demoDomain = null;
+  }
 
   const s3Config = config.s3;
   object.siteRoot = `http://${s3Config.bucket}.s3-website-${s3Config.region}.amazonaws.com`;
@@ -80,6 +116,9 @@ module.exports = (sequelize, DataTypes) => {
     },
     demoDomain: {
       type: DataTypes.STRING,
+      validate: {
+        isEmptyOrUrl,
+      },
     },
     config: {
       type: DataTypes.STRING,
@@ -90,6 +129,9 @@ module.exports = (sequelize, DataTypes) => {
     },
     domain: {
       type: DataTypes.STRING,
+      validate: {
+        isEmptyOrUrl,
+      },
     },
     engine: {
       type: DataTypes.ENUM,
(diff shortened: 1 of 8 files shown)


frontend/components/site/siteSettings.js
```diff
@@ -116,7 +116,8 @@ class SiteSettings extends React.Component {
                 <input
                   name="domain"
                   className="form-control"
-                  type="text"
+                  type="url"
+                  pattern="https?://.+"
```
jseppi: Using `type="url"` and `pattern="https?://.+"` are primarily so that feedback is provided to a user if they enter an invalid URL. They do not mitigate the XSS issue on their own because an attacker could still set "bad" URLs directly via the API.

test/api/unit/services/SQS.test.js
```diff
@@ -80,8 +80,6 @@ describe('SQS', () => {
     context("building a site's default branch", () => {
       it('should set an empty string for BASEURL in the message for a site with a custom domain', (done) => {
         const domains = [
-          'www.example.com',
```
jseppi: Removed all the protocol-less domains in the SQS tests since they are no longer valid.

api/models/site.js
```diff
@@ -80,6 +116,9 @@ module.exports = (sequelize, DataTypes) => {
     },
     demoDomain: {
       type: DataTypes.STRING,
+      validate: {
+        isEmptyOrUrl,
```
jseppi: The `validate` options added here prevent Site models from being created or updated with invalid URLs.

api/models/site.js
```diff
@@ -42,6 +68,16 @@ function toJSON() {
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
 
+  // prevent any invalid domain or demoDomain URLs from being returned
+  // or used in the creation of viewLink or demoViewLink
+  if (!isEmptyOrUrl(object.domain, false)) {
+    object.domain = null;
+  }
+
+  if (!isEmptyOrUrl(object.demoDomain, false)) {
```
jseppi: These checks prevent any *currently* existing Site models from returning invalid URLs for `domain` and `demoDomain` (since the new `validate` config below will only happen on model creation or update).

@jmhooper checked our production and staging databases and found no maliciously-invalid `domain` or `demoDomain` fields, however.

konklone: How about rewriting existing URLs in-place within the database to normalize them all to what you want to enforce going forward? Then you don't need special-case logic for it.

test/api/unit/models/site.test.js
```diff
@@ -1,102 +1,125 @@
-const expect = require("chai").expect
-const sinon = require("sinon")
-const config = require("../../../../config")
-const factory = require("../../support/factory")
-const { Site } = require("../../../../api/models")
+const expect = require('chai').expect;
```
konklone: The many changes to single/double quotes prevent easy detections of the changes to this file. Could you describe them in a sentence or two, or point to them in comments?

api/models/site.js
jseppi: Sure, though according to @jmhooper the only values in prod or staging that aren't valid URLs are values like `"test1"`, so they can't quite be turned into valid values. We could just remove them. Do you think that would be better? I don't really like mucking around in prod databases personally, but that should be an easy enough thing to do.

test/api/unit/models/site.test.js
```diff
@@ -1,102 +1,125 @@
...
+      defaultBranch: 'preview',
+      demoBranch: 'preview',
+    }).catch((err) => {
+      expect(err.status).to.equal(403);
+      expect(err.message).to.equal('Default branch and demo branch cannot be the same');
+      done();
+    }).catch(done);
+  });
+
+  it('should validate that the domain is a valid URL', (done) => {
```
jseppi: This test was added

test/api/unit/models/site.test.js
```diff
@@ -1,102 +1,125 @@
...
+      repository: 'repository',
+      domain: 'boop://beep.com',
+    }).catch((err) => {
+      expect(err.status).to.equal(403);
+      expect(err.message).to.equal('domain: URL must start with http:// or https://');
+      done();
+    }).catch(done);
+  });
+
+  it('should validate that the demoDomain is a valid URL', (done) => {
```
jseppi: This test was added

test/api/unit/serializers/site.test.js
```diff
@@ -1,37 +1,69 @@
...
+        .then(site => SiteSerializer.serialize(site))
+        .then((object) => {
+          const result = validateJSONSchema(object, siteSchema);
+          expect(result.errors).to.be.empty;
+          done();
+        })
+        .catch(done);
+    });
+
+    it('should not include invalid domain or viewLink', (done) => {
```
jseppi: This test was added

test/api/unit/serializers/site.test.js
```diff
@@ -1,37 +1,69 @@
...
+        })
+        .then((object) => {
+          expect(object.domain).to.be.empty;
+          expect(object.viewLink.indexOf(badUrl)).to.equal(-1);
+          done();
+        })
+        .catch(done);
+    });
+
+    it('should not include invalid demoDomain or demoViewLink', (done) => {
```
jseppi: This test was added

api/models/site.js
konklone: Yeah, I think that would be better, rather than complicating the logic of a security-relevant bug fix to avoid touching it. We could definitely get bugs that require us to migrate production data, so this just seems like something that falls into that class.

jseppi: okie dokie

test/api/unit/services/SQS.test.js
jmhooper: Are we sure? This is what I'm looking at:

```javascript
> validator.isURL('www.example.com')
true
```

So I'd think a domain without a protocol could still make it's way into database.

jseppi: I'm using a custom function that sets `protocol_required`. See https://github.com/18F/federalist/pull/1049/files#diff-5a61a2437a3a240173dd03188316466aR48

jseppi: And also this test: https://github.com/18F/federalist/pull/1049/files#diff-db4b881064ac845fd46df4812df5e033R118

jmhooper: Got it. Okay, so we'll want to also stick a protocol onto all the domains without protocols when we work on the db.

jseppi: Should we add the protocol, or just delete `domain`s/`demoDomain`s that do not have a protocol like `http(s)://`?

jmhooper: We should definitely add the protocol. Otherwise we risk breaking sites that depend on having a custom domain present.

jseppi: How can those sites work without a protocol? I must be missing something.

Like if a site currently has `boop.com` in their `domain` field, the "View Website" button currently links to `https://federalist-staging.18f.gov/sites/53/boop.com`, which doesn't work.

jmhooper: Yes, but it still affects the value of `BASEURL` in the build settings. Changing that in the db could cause the `BASEURL` value to be incorrect for some live sites.

jseppi: ah, so a protocol-less custom domain is a valid `BASEURL`?

jmhooper: Yeah, there should be a test for the SQS service that verifies that it sets the baseurl correctly for protocol-less domains and protocol-less domains without a trailing slash.

api/models/site.js
```diff
@@ -34,6 +44,22 @@ const beforeValidate = (site) => {
   }
 };
 
+
+function isEmptyOrUrl(value, throws = true) {
+  const validUrlOptions = {
+    require_protocol: true,
+    protocols: ['http', 'https'],
```
jmhooper: Maybe we should change this to `['https']`? That is what the frontend validation is looking for. If we do that. We can also drop the `http` domain in the SQS test.

jseppi: The front end validation is looking for either `http` or `https`. Note the `?` after the `s` in the regex `https?://.+`

jseppi: Though I am fine requiring all custom domains to be `https` if that's actually the case :)

jmhooper: Ah, didn't notice that, cool. I think we may want to open an issue to discuss enforcing an HTTPS requirement, but that's scope creep in the PR?

konklone: Definitely scope creep for the PR, but definitely a good idea, since Federalist already gives free HTTPS to all customers. 😺

jmhooper: I opened #1050 to discuss that, @konklone. I think we're all on agreement on that front.

jseppi:
@jmhooper - could you verify that the changes to `SQS.test.js` are keeping with (new) intended behavior?

test/api/unit/services/SQS.test.js
jseppi: Follow-up: The migration added to this PR will add `https://` to the beginning of all protocol-less values of `domain` and `demoDomain`. It will change any that start with `http://` to `https://`

Issue #1051, opened 2017-07-05
Missing 'rel="noopener noreferrer"' on "View Website" button in the header


jmhooper:
It looks like that button got left off the PR to add those.

jseppi:
Needs to be added to `linkButton.js`

PR #1052, opened 2017-07-05
set rel='noopener noreferrer' if target='_blank' in LinkButton

jseppi:
Completes the mitigation of #1041 by ensuring the `<LinkButton>` component results in an anchor tag with `rel='noopener noreferrer'` when it has `target='_blank'`.

Ref #1051

### DIFF
--- a/frontend/components/linkButton.js
+++ b/frontend/components/linkButton.js
@@ -1,26 +1,45 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Link } from 'react-router';
 
 const propTypes = {
-  alt: React.PropTypes.string,
-  children: React.PropTypes.oneOfType([
-    React.PropTypes.object,
-    React.PropTypes.array
+  alt: PropTypes.string,
+  children: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.array,
   ]),
-  className: React.PropTypes.string,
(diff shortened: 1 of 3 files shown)


jmhooper:
This is an interesting one.

My first instinct is to think there's a problem with the `LinkButton` component. You'd think (and we both did) it would pass arbitrary props to the child `Link` component and just add some styles. So initially my thinking is that the change should be to enable that behavior, and keep the `rel` setting where we had it.

On the other hand, I like enforcing `noreferer` / `noopener` on all buttons with a `target='_blank'` prop.

On the other other hand, there's a part of me that wants the `LinkButton` component's API to mimic the `Link` components API as much as possible, only adding style changes.

A few options:

1. Leave this the way it is. It works
2. Change the `LinkButton` component to forward props to the `Link` child and just apply the style changes
3. Remove the `LinkButton` component all together and replace instances of it with `Link` components.

Thoughts?

jseppi:
I'm pretty happy to get rid of `LinkButton`, so I will file a different PR that does that :)

jmhooper:
Nice, this works in the meantime.

jmhooper:
Or wait, did you mean close this one and file another one? That works too.

jseppi:
Planning to close this one and open a different one.

jseppi:
Superseded by #1053

Issue #1054, opened 2017-07-05
Authenticated user can restart builds for a site they don't have access to


jseppi:
As reported by `sp1d3rs` in https://hackerone.com/bugs?subject=tts-private&report_id=245872

> ### Steps to reproduce
> 
> 1. Login to the Federalist with your test account (we call it user1).
> 1. Restart the build on one of the sites belongs to you.
> 1. Catch the request to the http://192.168.119.128:1337/v0/build/ and change site parameter to the siteID, which was created on some other account (we call it user2).
> 1. Execute the request. It will be accepted.
> 1. Login as user2 and go to this site's builds. You will notice that build was restarted by another user.
> 
> ### Suggested fix
> 
> It looks like other endpoint (e.g. site settings modify) checks the site id correctly. The fix is add additional check too on this specific endpoint.
>

jmhooper:
We do have a test to make sure this doesn't happen. For some reason, it isn't holding up in staging prod: https://github.com/18F/federalist/blob/45ee891e37d14c2fdb731bd6aa278b1187e0a892/test/api/requests/site.test.js#L132

We'll want to look into why that test is failing while fixing this. From there, we should check on other, similar tests to make sure they don't suffer from the same failing.

jseppi:
Should be this test: https://github.com/18F/federalist/blob/staging/test/api/requests/build.test.js#L124

jseppi:
Here's the problem line: https://github.com/18F/federalist/blob/staging/api/authorizers/build.js#L6

Basically the first part of that if-condition will pass just if the `build.site` is truthy, so it is not actually doing a proper check to make sure the user has access to the specified site id.

PR #1060, opened 2017-07-06
Fix build authorizer to properly check that user has permissions on site

jseppi:
This site fixes an issue with the build authorizer that would allow _any_ user to create a build model (or restart one). The problem was a buggy if-condition.

Ref #1054

To Do:
- [x] fix tests
- [x] fix logic


test/api/unit/authorizers/build.test.js
```diff
@@ -1,89 +1,107 @@
...
-      }).then(({ user, build }) => {
-        return authorizer.create(user, { user: user.id, site: site.id })
-      }).then(() => {
-        done(new Error("Expected authorization error"))
-      }).catch(err => {
-        expect(err).to.equal(403)
-        done()
-      }).catch(done)
-    })
+    it("should reject if the build is not associated with one of the user's sites", (done) => {
```
jseppi: This is the test whose logic has been amended. It wasn't previously making a site that did belong to the user, so it wasn't really running through the logic of the target method.

api/authorizers/build.js
```diff
@@ -1,25 +1,27 @@
-const { User, Site } = require("../models")
+const { User, Site } = require('../models');
 
-const authorize = (user, build) => {
-  return User.findById(user.id, { include: [Site] }).then(user => {
-    for (site of user.Sites) {
-      if (build.site || build.Site.id === site.id) {
```
jseppi: This was the problem line. It was resolving the promise just if `build.site` was truthy, but not actually checking to see if the user was associated with the given site.

api/authorizers/build.js
```diff
@@ -1,25 +1,27 @@
...
-        return Promise.resolve()
+const authorize = (user, build) => (
+  User.findById(user.id, { include: [Site] })
+    .then((userModel) => {
+      const buildSiteId = build.site || build.Site.id;
+      const matchingUserSite = userModel.Sites.find(site => buildSiteId === site.id);
```
jseppi: These two lines are the essential part of th efix

api/controllers/build.js
```diff
@@ -50,14 +50,14 @@ module.exports = {
...
     authorizer.create(req.user, params)
-    .then(() => Build.create(params))
-    .then(build =>
-      GithubBuildStatusReporter.reportBuildStatus(build)
-      .then(() => build)
-    )
-    .then(build => buildSerializer.serialize(build))
-    .then(buildJSON => res.json(buildJSON))
-    .catch(res.error);
+      .then(() => Build.create(params))
```
jseppi: just linting fixes in this file

test/api/unit/services/S3PublishedFileLister.test.js
```diff
@@ -1,143 +1,146 @@
...
-        ])
-        done()
-      }).catch(done)
-    })
-
-    it("should reject with an error if S3.listObjects is unsuccessful", done => {
-      AWSMocks.mocks.S3.listObjects = (params, cb) => cb(new Error("Test error"))
-
-      factory.site().then(model => {
-        return S3PublishedFileLister.listPublishedFilesForBranch(site, "preview")
```
jseppi: Interestingly, this test started failing after fixing the build authorizer. If you see here in the old code, `site` was not actually defined! `model` was the name of the argument to the `then(...)` callback, so I'm not totally sure how this test ever worked o_O

test/api/unit/services/S3PublishedFileLister.test.js
```diff
@@ -1,143 +1,146 @@
...
+        done();
+      }).catch(done);
+    });
+
+    it('should reject with an error if S3.listObjects is unsuccessful', (done) => {
+      AWSMocks.mocks.S3.listObjects = (params, cb) => cb(new Error('Test error'));
+
+      factory.site()
+        .then(
+          site => S3PublishedFileLister.listPublishedFilesForBranch(site, 'preview')
```
jseppi: Here's the fix to the problem described in the above comment.

jseppi:
ready for review, @jmhooper

test/api/unit/authorizers/build.test.js
```diff
@@ -1,89 +1,107 @@
...
+          done();
+        })
+        .catch(done);
+    });
 
-    it("should reject if the build is not associated with one of the user's sites", done => {
-      Promise.props({
-        user: factory.user(),
-        site: factory.site(),
-      }).then(({ user, build }) => {
```
jmhooper: It looks like maybe there was also a typo here in my code? Looks like `build` is not in the original object and `site` is undefined...

test/api/unit/services/S3PublishedFileLister.test.js
jmhooper: My suspicion would be that maybe somewhere `site = "asdf"` exists without a `var`, `let`, or `const` declaration 😳.

api/authorizers/build.js
jmhooper: Good catch. Good fix. 👍.

jseppi:
Welp, I screwed that up. 🙃

jmhooper:
Did you... delete all the commits?

jseppi:
...yes, but can recover

PR #1061, opened 2017-07-06
Fix build authorizer to properly check if user is associated to site

jseppi:
This site fixes an issue with the build authorizer that would allow any user to create a build model (or restart one). The problem was a buggy if-condition. Ref #1054

Also:
- fix S3PublishedFileLister.test.js that broke due to proper build auth checking
- fix associated build authorizer and request tests
- lint files

Supersedes PR #1060 that I screwed up with a bad rebase

### DIFF
--- a/api/authorizers/build.js
+++ b/api/authorizers/build.js
@@ -1,25 +1,26 @@
-const { User, Site } = require("../models")
+const { User, Site } = require('../models');
 
-const authorize = (user, build) => {
-  return User.findById(user.id, { include: [Site] }).then(user => {
-    for (site of user.Sites) {
-      if (build.site || build.Site.id === site.id) {
-        return Promise.resolve()
+const authorize = (user, build) => (
+  User.findById(user.id, { include: [Site] })
+    .then((userModel) => {
+      const buildSiteId = build.site || build.Site.id;
+      const matchingUserSite = userModel.Sites.find(site => buildSiteId === site.id);
+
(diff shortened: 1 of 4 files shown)


api/authorizers/build.js
```diff
@@ -1,25 +1,26 @@
-const { User, Site } = require("../models")
+const { User, Site } = require('../models');
 
-const authorize = (user, build) => {
-  return User.findById(user.id, { include: [Site] }).then(user => {
-    for (site of user.Sites) {
-      if (build.site || build.Site.id === site.id) {
```
jseppi: This was the bad if-condition. Note it was checking just if `build.site` was truthy.

api/authorizers/build.js
```diff
@@ -1,25 +1,26 @@
...
-        return Promise.resolve()
+const authorize = (user, build) => (
+  User.findById(user.id, { include: [Site] })
+    .then((userModel) => {
+      const buildSiteId = build.site || build.Site.id;
+      const matchingUserSite = userModel.Sites.find(site => buildSiteId === site.id);
```
jseppi: This line and the one before it are the main parts of the fix.

jmhooper:
Appreciate documenting the relevant changes after merge 😄

jseppi:
hehe, yeah, just in case someone comes back to look at this and can't tell what is going on

PR #1071, opened 2017-07-18
.about.yml maintenance

jseppi:
Various fixes/clarifications to our `.about.yml`.

Most significant change is an update to the `team` list. I made previous folks "emeritus" members and added the three current team members (@wslack, @jmhooper, and me).

I also removed the `snyk.io` badge from README.


jmhooper:
+1 to the Gemnasium addition

Issue #1073, opened 2017-07-18
Setup scanjs rules via eslint


jseppi:
See https://before-you-ship.18f.gov/security/static-analysis/#javascript for info

Also ref https://github.com/18F/before-you-ship/issues/320 for changes that might be necessary to that guidance

jseppi:
Will want #1070 merged first so we can make sure adding `scanjs` doesn't break CodeClimate, or, if it does, how to tell CodeClimate to not use that plugin.

PR #1070, opened 2017-07-18
Configure CodeClimate

jseppi:
This PR should get code quality reporting in Code Climate.

I decided to punt on test coverage reporting and setting up `scanjs` with `eslint` for separate PRs that can build off of this baseline functionality. Furthermore, using the `nodesecurity` engine in CodeClimate requires a `shrinkwrap.json` be present, but we are using `yarn`, so we don't have one :\ I think I'll see what shakes out at today's ATO meeting to decide what to do regarding static security analysis.

To Do:
- [x] Request and wait for default branch setting change by CodeClimate team
- ~Add nodesecurity engine: https://docs.codeclimate.com/v1.0/docs/nodesecurity~ Won't work because we don't use `npm shrinkwrap`
- ~Figure out test coverage reporting.~ Punt to a separate PR.

PR #1075, opened 2017-07-19
Setup scanjs rules in eslint

jseppi:
As required by the [Before You Ship guide](https://before-you-ship.18f.gov/security/static-analysis/#javascript).

ref #1073

- Change .eslintrc files into JavaScript .eslintrc.js files
- Only include scanjs and no-unsanitized plugins and rules when scanjs-rules is able to be `require`'d. This prevents eslint in Code Climate from attempting to run those plugins/rules, because they are not available there and will error out.

- [x] Should wait to merge until #1070 is merged so we can make sure it works with CodeClimate


jseppi:
screen shot 2017-07-19 at 10 29 00 am

😬

jseppi:
Trying to look into this, but Code Climate's website gives me a `504 Gateway Timeout` when I try to  access the [details link](https://codeclimate.com/github/18F/federalist/pull/1075).

jseppi:
Code Climate doesn't support the scanjs plugin, so I am guessing that's why so many errors are generated. And will all the errors their server is bombing out trying to load the details page.

Here's the list of supported plugins: https://docs.codeclimate.com/v1.0/docs/eslint#section-eslint-3

I'm not sure what to do now. Seems we're at an impasse with what Before You Ship recommends versus what is possible.

jseppi:
Going to try this approach:

- Make new `.eslintrc.codeclimate.json` file that does not have scanjs
- Point `.codeclimate.yml` at that file
- Change our root `.eslintrc` to extend `.eslintrc.codeclimate.json` and keey the scanjs rules in that file

This might cause some weirdness due to our descendent `test/.eslintrc` but I'll see what happens.

jseppi:
Latest commit got rid of the 21000 errors, but now having problems with `mocha` stuff not being defined. I could add `env: mocha` to the new `.eslintrc.codeclimate.json` file, but then we'll have multiple configs that could end up out-of-sync :\

jseppi:
What do you think of this approach, @jmhooper? It seems a bit silly to me to have to maintain separate configs, but I'm not really sure what else to do.

jmhooper:
I agree. I'd wait to hear back from infrastructure. I'd think it defeats the purpose to have ScanJS configured, but not use it all. I'm also not a fan of the duplication between the configs.

jseppi:
It's still used by our Travis config and local editor plugins

jseppi:
@jmhooper: Let me know if you think this is 👍 and then I'll clean up the commit history and implement a similar thing over in federalist-builder.

jseppi:
Neither did I. I learned it from Code Climate actually, which was reporting that our extension-less `.eslintrc` files were deprecated and to please use an extension.

Issue #1058, opened 2017-07-06
Error pages linked-to from GitHub checks can appear blank if user doesn't have proper permissions


toolness:
When working on https://github.com/18F/web-design-standards/pull/2015, the federalist build that was triggered failed and the "Details" link pointed at https://federalist.18f.gov/sites/62/builds/9365/logs.  However, for me this page was almost entirely blank:

> ![blank-federalist](https://user-images.githubusercontent.com/124687/27918034-c01e8fa6-623b-11e7-80cb-deb29fbe2043.png)

I talked to @jseppi and he said that this was because I didn't have permission to view the logs. I thought it might be useful to say as much on this page, instead of being blank, thus I am filing this issue.

wslack:
triaged to top of PR2 backlog. Most of the people who might click on these links aren't authorized for Federalist, so you're in a middle case.

wslack:
This is beautifully fixed!

PR #1077, opened 2017-07-21
Show error message when user has no site data for given site id param

jseppi:
Ref #1058

This PR modifies the `SiteContainer` component to render an error message after sites have loaded if there is no `site` with the same `id` as in the URL params, which generally means either such a site does not exist, or the user is not authorized to view that site.

`SiteContainer` was also previously untested, so this adds tests for that component.

Looks like this:

screen shot 2017-07-21 at 12 48 18 pm

### DIFF
--- a/frontend/components/siteContainer.js
+++ b/frontend/components/siteContainer.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import SideNav from './site/SideNav/sideNav';
 import PagesHeader from './site/pagesHeader';
 import AlertBanner from './alertBanner';
+import LoadingIndicator from './loadingIndicator';
 
 const propTypes = {
   params: PropTypes.shape({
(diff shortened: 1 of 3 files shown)


test/frontend/components/siteContainer.test.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { expect } from 'chai';
+
+import SiteContainer from '../../../frontend/components/siteContainer';
+
+let props;
+
+describe('<SiteContainer/>', () => {
```
jmhooper: 👍 for adding tests for this component

frontend/components/siteContainer.js
```diff
@@ -1,53 +1,95 @@
...
+      publishedBranches,
+      publishedFiles,
+      githubBranches,
+    };
 
     const pageTitle = this.getPageTitle(location.pathname);
 
     if (!site) {
-      return null;
+      if (storeState.sites.isLoading) {
```
jmhooper: This works, but the nested conditionals make me a little antsy. Any way we could extract the missing site case into another function or otherwise flatten out the conditional?

jseppi: sure thing

jseppi:
@jmhooper - I de-nested the `if` statements and rebased

Issue #1079, opened 2017-07-24
Use robots.txt to block search engines from indexing federalist-staging


jseppi:
Currently we have the same `robots.txt` served on both [production](https://federalist.18f.gov/robots.txt) and [staging](https://federalist-staging.18f.gov/robots.txt). We should use a different `robots.txt` on staging to block search spiders from indexing the staging domain  because we don't want that to come up when people search for "federalist."

wslack:
I don't know....deliberately desyncing https://github.com/18F/federalist/blob/master/public/robots.txt between branches seems unwise. I'm not opposed to it coming up in Google.

wslack:
Googling "federalist staging" doesn't show the URL. I think Google may already do this for us.

jseppi:
It wouldn't be de-syncing code branches. It would be delivering different robots.txt content based on the `APP_ENV` setting. In fact, we might want to do it based on request url, so that `federalist.fr.cloud.gov` isn't indexed.

I can't think of any way Google would know not to index it as it is currently setup.

![image](https://user-images.githubusercontent.com/697848/28536765-527f331a-706e-11e7-992e-f875684236d5.png)

jseppi:
(also this is quite low effort to implement)

wslack:
Will be closed after confirmed this worked on Prod.

jseppi:
Confirming that now the production route work has been done (#1094) and prod has been deployed (#1104), our `robots.txt` content is properly returned.

Disallow only `/preview` when using the prod hostname:

<img width="415" alt="screen shot 2017-08-07 at 9 20 08 am" src="https://user-images.githubusercontent.com/697848/29030794-db2f6bec-7b51-11e7-94e8-488bd39850ff.png">

Disallow all if not using the prod hostname:

<img width="436" alt="screen shot 2017-08-07 at 9 19 56 am" src="https://user-images.githubusercontent.com/697848/29030801-e034e630-7b51-11e7-8592-5da3e45ac47f.png">

wslack:
👏

PR #1088, opened 2017-07-27
Dynamically return robots.txt based on app_env and req.hostname

jseppi:
This PR makes it so `robots.txt` with our production rules is only returned if `app_env == 'production' && request.hostname == 'federalist.18f.gov'`

Otherwise a `robots.txt` that denies all is returned.

Ref #1079

### DIFF
--- a/api/controllers/main.js
+++ b/api/controllers/main.js
@@ -43,7 +43,7 @@ module.exports = {
 
     // If this is the production instance and the request came to the production hostname
     if (config.app.app_env === 'production' &&
-      config.app.hostname === `${req.protocol}://${req.hostname}`) {
+      config.app.hostname === `https://${req.hostname}`) {
       // then send the production robots.txt content
       return res.send(PROD_CONTENT);
     }
--- a/api/controllers/main.js
+++ b/api/controllers/main.js
@@ -34,4 +34,21 @@ module.exports = {
 
     res.render('index.html', context);
   },
+
+  robots(req, res) {
+    const PROD_CONTENT = 'User-Agent: *\nDisallow: /preview\n';
+    const DENY_ALL_CONTENT = 'User-Agent: *\nDisallow: /\n';
+
+    res.set('Content-Type', 'text/plain');
+
+    // If this is the production instance and the request came to the production hostname
+    if (config.app.app_env === 'production' &&
+      config.app.hostname === `${req.protocol}://${req.hostname}`) {
+      // then send the production robots.txt content
+      return res.send(PROD_CONTENT);
+    }
+
+    // otherwise send the "deny all" robots.txt content
+    return res.send(DENY_ALL_CONTENT);
+  },
 };
--- a/api/routers/main.js
+++ b/api/routers/main.js
@@ -3,5 +3,6 @@ const MainController = require('../controllers/main');
 
 router.get('/', MainController.index);
 router.get('/sites(/*)?', MainController.index);
+router.get('/robots.txt', MainController.robots);
 
 module.exports = router;
--- a/manifest.yml
+++ b/manifest.yml
@@ -3,7 +3,9 @@ applications:
 - name: federalist
 buildpack: nodejs_buildpack
 stack: cflinuxfs2
-domain: fr.cloud.gov
+routes:
+- route: federalist.18f.gov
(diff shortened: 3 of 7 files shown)


api/controllers/main.js
```diff
@@ -34,4 +34,21 @@ module.exports = {
...
+
+  robots(req, res) {
+    const PROD_CONTENT = 'User-Agent: *\nDisallow: /preview\n';
+    const DENY_ALL_CONTENT = 'User-Agent: *\nDisallow: /\n';
+
+    res.set('Content-Type', 'text/plain');
+
+    // If this is the production instance and the request came to the production hostname
+    if (config.app.app_env === 'production' &&
+      config.app.hostname === `${req.protocol}://${req.hostname}`) {
```
jmhooper: Do `req.hostname` and `req.protocol` return the values we're expecting them to behind the cloud.gov router?

jseppi: Hrm, great question.  Any idea how to test?

jmhooper: The thing that comes to mind is a sandbox app that logs `req.hostname` and `req.protocol`.

jmhooper: Though that wouldn't be behind the CDN so... hmm...

jseppi: blurglelburg :\

jseppi: oof, looks like when visiting `https://federalist-staging.18f.gov/robots.txt` we get 

```
protocol: http
hostname: federalist-staging.fr.cloud.gov
```
 
inside the app.

jseppi: I think we need to do `app.enable('trust proxy')` so that express will use `X-Forwarded-Host` as the `req.hostname` when it is available. CloudFront implements this safely -- it will not allow clients to set `X-Forwarded-*` headers.

jseppi: No dice with that either. It might be something that needs to be turned on in CloudFront. I'll see if I can check `X-Forwarded-Host` directly.

jseppi: So something is actually up with the `cdn-route` service. @jcscottiii and I are going to pair next week to figure out what the deal is, but it doesn't appear to be something application-level.

jseppi:
Update: this now has some changes to our manifest files to map the real URLs (`federalist.18f.gov` and `federalist-staging.18f.gov`) to the corresponding apps. We'll also need to update the production `cdn-route` service to map the domain to the real URL:

```sh
cf update-service federalist-route -c '{"domain": "federalist.18f.gov,federalist.fr.cloud.gov"}'
```

api/controllers/main.js
```diff
@@ -34,4 +34,21 @@ module.exports = {
...
+
+  robots(req, res) {
+    const PROD_CONTENT = 'User-Agent: *\nDisallow: /preview\n';
+    const DENY_ALL_CONTENT = 'User-Agent: *\nDisallow: /\n';
+
+    res.set('Content-Type', 'text/plain');
+
+    // If this is the production instance and the request came to the production hostname
+    if (config.app.app_env === 'production' &&
+      config.app.hostname === `https://${req.hostname}`) {
```
jseppi: `req.protocol` is always `http` when deployed to cloud.gov since http is used for internal routing within cloud.gov, but our `APP_ENV` environment variable is always `https` since that is what is needed for external clients. Thus, I've hardcoded the protocol as `https://` here in this if-condition.

jseppi:
Here's the order of operations for production:

```sh
cf t -s production
cf create-route production 18f.gov --hostname federalist-staging
cf map-route federalist 18f.gov --hostname federalist
cf update-service federalist-route -c '{"domain": "federalist.18f.gov"}'
```

jmhooper:
Looks like you have a typo in the first one. Shouldn't `federalist-staging` be `federalist`? Otherwise that looks 👌, let's do it :+1:

jseppi:
Good catch - fixed!

jseppi:
Are you saying to make the `production` route changes now(-ish)?

jmhooper:
No, let's wait until we're ready to deploy everything to prod to make the route changes.

jseppi:
Issue is a good idea. Maybe I can merge this then mark #1079 as blocked until the new issue is resolved.

Issue #1092, opened 2017-08-01
Implement rate limiting on Federalist API


wslack:
Per https://hackerone.com/bugs?report_id=249319

We should implement a basic rate limiting package.

PR #1093, opened 2017-08-02
add express-rate-limit for rate limiting

jseppi:
ref #1092 

Adds simple rate limiting using [express-rate-limit](https://github.com/nfriedly/express-rate-limit) for all requests.

Also:
- refactor router into separate routers for improved understandability/readability:
  - `apiRouter` for all API routes, ie, those that start with `/v0`
  - `mainRouter` for all other routes
- fix linting errors on touched files

To Do:
- [x] ensure `req.ip` is different per user -- ie, that it is not the cloud.gov router IP
  - Update: `req.ip` is indeed the IP of the proxy, but we can use `app.enable('trust proxy')` to get the correct ip in `req.ip`
- [x] add `app.enable('trust proxy')`
- [x] figure out good parameters to use for the rate limiter
- [x] move rate limit options to config

### DIFF
--- a/app.js
+++ b/app.js
@@ -24,6 +24,8 @@ const session = require('express-session');
 const PostgresStore = require('connect-session-sequelize')(session.Store);
 const responses = require('./api/responses');
 const passport = require('./api/services/passport');
+const RateLimit = require('express-rate-limit');
+const router = require('./api/routers');
 
 const app = express();
 const sequelize = require('./api/models').sequelize;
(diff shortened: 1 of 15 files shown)


jseppi:
lol, the rate limiting broke the tests

app.js
```diff
@@ -65,8 +72,17 @@ app.use(expressWinston.errorLogger({
   ],
 }));
 
-const routers = require('./api/routers');
+const limiter = new RateLimit({
```
jmhooper: One thought I have is we could add this to a config, something like `config.rateLimitting`. Then you could configure production like this, and do this for the default:

```
{
  max: 0,
  delayAfter: 0,
  delayMs: 0
}
```

That should enable rate limiting in prod, but disable it in dev and test. But we could still get and idea of how the app behaves when requests are passed through the middleware in the test and dev environment.

Thoughts?

jseppi:
Yeah, that's a good idea!

jseppi:
ready for review again, @jmhooper

PR #1098, opened 2017-08-03
report test coverage with nyc and codeclimate-test-reporter

jseppi:
This PR fixes code coverage reporting by removing `istanbul` and instead using its cli, which is called`nyc`. Now when we run tests we'll see a nice table of coverage results.

In addition, I attempted to hook up Code Climate's test coverage reporter. The Travis build for this PR said that the coverage report was successfully submitted, but I'm not sure where to see it. The Code Climate docs for code coverage are extremely confusing, but I'm hoping that once merged we'll start to see coverage reports in our PRs. I somehow got this working for CALC, but now I'm 😖. 

Oh, also I put our `CODECLIMATE_REPO_TOKEN` in our TravisCI settings as a private env var. It's not really necessary to make it private because really its just a unique id and can only be used to post test results, not really a private "token" in any security sense:

> Since we don't consider a repo's test reporter ID a sensitive value, regenerating it shouldn't be necessary in most cases. However you may regenerate the ID in the repo's "Settings" section under "Test Coverage". After regenerating the ID, any existing CI jobs reporting test coverage will stop working until you update their configuration to use the new test reporter ID.

- https://docs.codeclimate.com/v1.0/docs/test-coverage-troubleshooting-tips#section--should-i-keep-my-test-coverage-token-secret-


jmhooper:
Looks like code climate is saying this:

```
Sending test coverage results to codeclimate.com ...
Test coverage data sent.
```

...but I don't see the coverage status check running against this PR. Is that something we need to enable, or will it come on once we have test coverage info about staging?

jseppi:
Right, that's what I can't quite figure out. I'm hoping something will happen once it is moved to `staging`, but I dunno. I also don't want to spend much time on this since it was just supposed to be a quick fix to the coverage task.

PR #1099, opened 2017-08-03
Test test coverage

jmhooper:
I deleted some tests to see if code climate would get mad about declining test coverage


jseppi:
While it is not reporting it as a GitHub PR status, the CodeClimate Dashboard now includes test coverage: https://codeclimate.com/github/18F/federalist/coverage

jmhooper:
@jseppi: It looks like we may need to run `cc-test-reporter before-build` before the build starts to notify Code Climate of a pending report:

Ref: https://docs.codeclimate.com/v1.0/docs/setting-up-test-coverage

jseppi:
This is something I was confused about. There is `cc-test-reporter` and also the [`codeclimate-test-reporter`](https://docs.codeclimate.com/v1.0/docs/travis-ci-javascript-test-coverage) (which is what I used in my last PR. Do we not need the `codeclimate-test-reporter` if we use `cc-test-reporter`?

jseppi:
Also, we _didn't_ use `cc-test-reporter` in CALC (see our travis config: https://github.com/18F/calc/blob/develop/.travis.yml), yet somehow we still had PR test coverage integration:

screen shot 2017-08-03 at 3 50 25 pm

-- from https://github.com/18F/calc/pull/1557

jseppi:
Closing since we got it working in #1101

PR #1101, opened 2017-08-04
Switch to cc-test-reporter to enable PR coverage statuses

jseppi:
use cc-test-reporter and add nyc config file

- cc-test-reporter will enable PR test coverage status reports from Code Climate
- don't output test summary table (it's kind of annoying and too much info)
- make sure to include .jsx files in coverage
- enable nyc cache
- remove unnecessary `coverage` task from `package.json`

To Do:

- [x] extract the `.nycrc` stuff from #1102 since that PR seems to be getting bigger. ~wait on #1102 to be merged since it adds an `.nycrc` file~
- [x] remove the `coverage` task in `package.json` and the call to it in `.travis.yml` because it won't be needed after that merge
- [x] clean up this PR's commit history to just include the changes that made it work
- [ ] After merge: remove `CODECLIMATE_REPO_TOKEN` from https://travis-ci.org/18F/federalist/settings since it is superseded by `CC_TEST_REPORTER_ID`
- ~figure out how to combine back end and front end coverage reports~ Going to punt on this and instead open a new issue for it since I can't figure out any way to do it currently: #1116

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,0 +1,9 @@
+{
+  "reporter": [
+    "lcov"
+  ],
+  "extension": [
+    ".jsx"
+  ],
+  "cache": true
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,13 @@ install:
   - yarn --pure-lockfile
 before_script:
   - "psql -c 'create database travis_ci_test;' -U postgres"
+  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+  - chmod +x ./cc-test-reporter
 script:
   - scripts/lint-travis.sh
   - yarn test
 after_success:
-  - npm install -g codeclimate-test-reporter
-  - yarn coverage # converts to lcov format
-  - codeclimate-test-reporter < ./coverage/lcov.info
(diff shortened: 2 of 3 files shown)


jseppi:
It worked, @jmhooper! Looks like you were right about using that other CLI tool from Code Climate.

jseppi:
Ah, I just noticed the back end and front end coverage reports are not combined, and it looks like the front end report is "winning" out as the one that gets reported. This also happens locally, which can easily be seen when viewing the `html`-formatted coverage report:

screen shot 2017-08-04 at 11 25 45 am

jseppi:
Ready for review @jmhooper

.nycrc
```diff
@@ -0,0 +1,9 @@
+{
+  "reporter": [
+    "lcov"
+  ],
+  "extension": [
+    ".jsx"
```
jmhooper: Do we know if this will prevent it from creating coverage reports for regular `.js` files?

jseppi: Yeah, it always covers `.js` files. `"extension"` just provides _additional_ extensions to instrument. I agree, it's weird.

I just confirmed locally also by using the `"text"` reporter for easily visualizing the files covered. Here's a snippet that shows both `.js` and `.jsx` are included:

screen shot 2017-08-11 at 9 37 14 am

Issue #1018, opened 2017-06-20
Special characters in branch names (like "#") lead to sites that can't be viewed


jseppi:
Characters that are allowed in git branch names, but not allowed, or that have special meaning, in URLs cannot actually be viewed. The build itself (and upload to S3) appears to work, however.

I'm not clear on how we want to solve for this. Some possibilities:

- Strip special characters from the branch name (like we do with repo names) for constructing the URLS
- Do nothing, including not doing a build, and communicate the problem to the user.

wslack:
Flag those with a message in the branch list in the "settings" screen?

jseppi:
Here's where I'm going with this:

<img width="770" alt="screen shot 2017-08-03 at 4 43 33 pm" src="https://user-images.githubusercontent.com/697848/28945163-fba358dc-786a-11e7-8e92-854affec3090.png">

Notice that I also changed the link text to hopefully better communicate the production branch site link and the demo branch site link.

Thoughts, @wslack and @jmhooper?

PR #1102, opened 2017-08-04
Create BranchViewLink component

jseppi:
- render a message instead of a link if a branch has special characters in it (ref #1018)
  - Only allow alphanumeric, `-`, and `_`
- render different link text for `defaultBranch` and `demoBranch`
- adjust text on normal branch preview links
- remove all S3-based site URLs (as in the "View Site" button) in favor of using a `federalist-proxy` url (configured in `config.app.preview_hostname`)

Looks like:

screen shot 2017-08-22 at 3 02 53 pm

and:

screen shot 2017-08-22 at 3 05 32 pm

Somewhat unrelated - add `.nycrc` config file for `nyc`:
- don't output test summary table (it's kind of annoying and too much info)
- make sure to include .jsx files in coverage

To Do:
- [x] use new component in `<SitePublishedBranchesTable>`
- [x] add tests for new component
- [ ] clean up commit history

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,11 +25,20 @@ try {
   const scanjsConfig = require('eslint-plugin-scanjs-rules');
   scanjsRulesConfig = {};
 
-  // bring in all the scanjs rules, prefixed with 'scanjs-rules',
+  // ScanJS rules we don't want to use:
+  const ignoredScanjsRules = [
+    // scanjs is looking for a different `connect` than the one we use often from react-redux
+    'call_connect',
+  ];
+
+  // bring in all the scanjs rules (other than those we want ignored),
+  // prefixed with 'scanjs-rules',
   // but convert them to eslint errors instead of warnings
-  Object.keys(scanjsConfig.rulesConfig).forEach((rule) => {
-    scanjsRulesConfig[`scanjs-rules/${rule}`] = 1;
-  });
+  Object.keys(scanjsConfig.rulesConfig)
+    .filter(rule => ignoredScanjsRules.indexOf(rule) === -1)
+    .forEach((rule) => {
+      scanjsRulesConfig[`scanjs-rules/${rule}`] = 1;
+    });
 } catch (err) {
   // eslint-disable-next-line no-console
   console.log('eslint-plugin-scanjs-rules not found, will not include scanjs rules');
--- a/api/controllers/main.js
+++ b/api/controllers/main.js
@@ -21,11 +21,17 @@ module.exports = {
 
     const siteDisplayEnv = config.app.app_env !== 'production' ? config.app.app_env : null;
 
+    const frontendConfig = {
+      TEMPLATES: config.templates,
+      PREVIEW_HOSTNAME: config.app.preview_hostname,
+    };
+
     const context = {
       siteWideError: null,
       jsBundleName: webpackAssets['main.js'],
       cssBundleName: webpackAssets['main.css'],
       siteDisplayEnv,
+      frontendConfig,
     };
 
     if (req.session.authenticated) {
--- a/api/controllers/preview.js
+++ b/api/controllers/preview.js
@@ -1,10 +1,9 @@
-const logger = require("winston")
-const url = require("url")
-const config = require("../../config")
+const url = require('url');
+const config = require('../../config');
 
 module.exports = {
-  redirect: function(req, res) {
-    const target = url.resolve(config.app.preview_hostname, req.path)
-    res.redirect(target)
-  }
-}
+  redirect(req, res) {
+    const target = url.resolve(config.app.preview_hostname, req.path);
+    res.redirect(target);
+  },
+};
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -1,4 +1,3 @@
-const url = require('url');
 const validator = require('validator');
 
 const config = require('../../config');
@@ -44,6 +43,19 @@ const beforeValidate = (site) => {
   }
 };
 
+
+function siteUrl() {
+  return this.domain || `${config.app.preview_hostname}/site/${this.owner}/${this.repository}/`;
+}
+
+function demoUrl() {
+  return this.demoDomain || `${config.app.preview_hostname}/demo/${this.owner}/${this.repository}/`;
+}
+
+function branchPreviewUrl(branch) {
+  return `${config.app.preview_hostname}/preview/${this.owner}/${this.repository}/${branch}/`;
+}
+
 function toJSON() {
   const object = Object.assign({}, this.get({
     plain: true,
@@ -52,11 +64,10 @@ function toJSON() {
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
 
-  const s3Config = config.s3;
-  object.siteRoot = `http://${s3Config.bucket}.s3-website-${s3Config.region}.amazonaws.com`;
-  object.viewLink = object.domain || [object.siteRoot, 'site', object.owner, object.repository].join('/');
+  object.viewLink = this.siteUrl();
+
   if (object.demoBranch) {
-    object.demoViewLink = object.demoDomain || [object.siteRoot, 'demo', object.owner, object.repository].join('/');
+    object.demoViewLink = this.demoUrl();
   }
 
   Object.keys(object).forEach((key) => {
@@ -69,18 +80,13 @@ function toJSON() {
 }
 
 function viewLinkForBranch(branch) {
-  const s3Root = `http://${config.s3.bucket}.s3-website-${config.s3.region}.amazonaws.com`;
-
-  if (branch === this.defaultBranch && this.domain) {
-    return this.domain;
-  } else if (branch === this.defaultBranch) {
-    return `${s3Root}/site/${this.owner}/${this.repository}`;
-  } else if (branch === this.demoBranch && this.demoDomain) {
-    return this.demoDomain;
+  if (branch === this.defaultBranch) {
+    return this.siteUrl();
   } else if (branch === this.demoBranch) {
-    return `${s3Root}/demo/${this.owner}/${this.repository}`;
+    return this.demoUrl();
   }
-  return url.resolve(config.app.preview_hostname, `/preview/${this.owner}/${this.repository}/${branch}`);
+
+  return this.branchPreviewUrl(branch);
 }
 
 function isEmptyOrUrl(value) {
@@ -148,6 +154,9 @@ module.exports = (sequelize, DataTypes) => {
     instanceMethods: {
       toJSON,
       viewLinkForBranch,
+      siteUrl,
+      demoUrl,
+      branchPreviewUrl,
     },
     hooks: {
       beforeValidate,
--- a/api/routers/preview.js
+++ b/api/routers/preview.js
@@ -1,7 +1,13 @@
-const router = require("express").Router()
-const PreviewController = require("../controllers/preview")
+const router = require('express').Router();
+const PreviewController = require('../controllers/preview');
 
-router.get("/preview/:owner/:repo/:branch", PreviewController.redirect)
-router.get("/preview/:owner/:repo/:branch/*", PreviewController.redirect)
+// These /preview/ routes provide redirects to the proxy app
+// (configured as config.app.preview_hostname).
+// They are used to not break existing links from when
+// Federalist site previews were handled in this application instead
+// of by federalist-proxy.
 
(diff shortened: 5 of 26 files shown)


jseppi:
Feedback on language is very welcome, @wslack.

frontend/components/site/siteGithubBranchesTable.js
```diff
@@ -1,46 +1,31 @@
 import React from 'react';
 
 import LoadingIndicator from '../loadingIndicator';
+import BranchViewLink from '../branchViewLink';
 import { SITE, GITHUB_BRANCHES } from '../../propTypes';
 
-const tableHeader = () => (
```
jseppi: I got rid of some of these functions and just moved the JSX markup into the `renderTable` method because I didn't feel like all the functions were increasing clarity/readability.

jseppi:
Oh, I just realized I should use the same component over in `<SitePublishedBranchesTable>` as well. Changing back to WIP.

jseppi:
Hrm, the links in `<SitePublishedBranchesTable>` don't have exactly the same behavior as the linsk in the branch table in Site Settings. The links in <SitePublishedBranchesTable>` always point to `branch.viewLink`.

Should the links in these two tables perform the same way?

wslack:
Long term, I want all of these to point to URLs like https://federalist-proxy.app.cloud.gov/preview/18f/federalist-docs/967-switch-to-yarn with consistant behavior, after we reorganize the menus and such per the card sort.

jseppi:
Cool, is this correct then?:

For a given branch:
- default behavior: link to the `federalist-proxy` URL
- if branch is the default branch and there is a custom URL: link to the custom URL
- if branch is the demo branch and there is a custom demo URL: link to the custom demo URL

jseppi:
Notes:
- never use the direct s3 link
- add comment to `/preview/` route to indicate that it is there for legacy support
- site settings branch table comes from github, but published branches table comes from our api, so there might be some mis-match there.

api/routers/preview.js
```diff
@@ -1,7 +1,13 @@
...
+const router = require('express').Router();
+const PreviewController = require('../controllers/preview');
 
-router.get("/preview/:owner/:repo/:branch", PreviewController.redirect)
-router.get("/preview/:owner/:repo/:branch/*", PreviewController.redirect)
+// These /preview/ routes provide redirects to the proxy app
+// (configured as config.app.preview_hostname).
+// They are used to not break existing links from when
+// Federalist site previews were handled in this application instead
+// of by federalist-proxy.
```
jseppi: added explanatory text

api/controllers/preview.js
```diff
@@ -1,10 +1,9 @@
-const logger = require("winston")
-const url = require("url")
-const config = require("../../config")
+const url = require('url');
```
jseppi: just linting fixes in this file

api/models/site.js
```diff
@@ -44,6 +43,19 @@ const beforeValidate = (site) => {
   }
 };
 
+
+function siteUrl() {
```
jseppi: created these 3 methods for reuse in `toJSON()` and `viewLinkForBranch()`

api/serializers/published-branch.js
```diff
@@ -1,19 +1,15 @@
...
-    const object = serializeObject(site, branch)
-    return Promise.resolve(object)
-  }
-}
-
-const serializeObject = (site, name) => {
-  return {
-    name,
-    site: site.toJSON(),
-    viewLink: site.viewLinkForBranch(name),
```
jseppi: removed the `viewLink` property; rather, this logic will be performed by the new front end `BranchViewLink` component

frontend/components/branchViewLink.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
jseppi: This is the new component for use in both `SiteGithubBranchesTable` and `SitePublishedBranchesTable`

frontend/components/branchViewLink.js
```diff
@@ -0,0 +1,43 @@
...
+import { SITE } from '../propTypes';
+import appConfig from '../../config/app';
+
+
+const isDefaultBranch = (branchName, site) => branchName === site.defaultBranch;
+
+const isDemoBranch = (branchName, site) => branchName === site.demoBranch;
+
+// we only want to link branch names that are alphanumeric plus _ and -
+const isLinkable = s => /^[a-zA-Z0-9_-]+$/.test(s);
```
jseppi: This is the core logic of the fix for #1018

frontend/components/disclaimer.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-const flagUrl = require('uswds/dist/img/us_flag_small.png');
+import flagUrl from 'uswds/dist/img/us_flag_small.png';
```
jseppi: just a minor clean-up -- no need to use `require`

api/models/site.js
```diff
@@ -52,11 +64,10 @@ function toJSON() {
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
 
-  const s3Config = config.s3;
-  object.siteRoot = `http://${s3Config.bucket}.s3-website-${s3Config.region}.amazonaws.com`;
```
jseppi: removed the S3 `siteRoot` from the serialized `site` model since we never want to link to the S3 bucket URL

frontend/components/branchViewLink.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { SITE } from '../propTypes';
+import appConfig from '../../config/app';
```
jseppi: Use the app config settings so we can get `preview_hostname` in the front end

api/models/site.js
jmhooper: Reads much easier this way 👍

frontend/components/site/siteGithubBranchesTable.js
jmhooper: This component looks much nicer

frontend/components/branchViewLink.js
jmhooper: Hmm, this is a little bit clunky. The `preview_hostname` variable in `config/app.js` is hardcoded to this:

```javascript
process.env.FEDERALIST_PREVIEW_HOSTNAME || 'http://localhost:1338'
```

That means that if `FEDERALIST_PREVIEW_HOSTNAME` is not set in the environment where this is built (in our case CI) then this won't work. That makes me a little nervous because that isn't really obvious.

Additionally, pulling configs into the frontend makes me nervous because secrets can land in those and then leak into the frontend code. With that in mind, my ask would be to add a comment to the app config saying as much.

jseppi: Yeah, I kinda agree with feeling a littler nervous about pulling in config to the front end. I saw it was already done in the `AddSite/index.js` component to get the `templates` config.

hrmmmmmmmm....

jseppi: What about instead using a context variable in the `main` controller to pass explicit config values (both `preview_hostname` and the `templates` config for now) to the rendered template. They could then be exposed as JS in a `<script>` block.

jmhooper: :+1:

jseppi: Cool, will bust that up

.eslintrc.js
```diff
@@ -25,11 +25,20 @@ try {
   const scanjsConfig = require('eslint-plugin-scanjs-rules');
   scanjsRulesConfig = {};
 
-  // bring in all the scanjs rules, prefixed with 'scanjs-rules',
+  // ScanJS rules we don't want to use:
+  const ignoredScanjsRules = [
+    // scanjs is looking for a different `connect` than the one we use often from react-redux
+    'call_connect',
```
jseppi: ran into some warnings due to this rule since I added the the use of `react-redux`'s `connect()` function

api/controllers/main.js
```diff
@@ -21,11 +21,17 @@ module.exports = {
 
     const siteDisplayEnv = config.app.app_env !== 'production' ? config.app.app_env : null;
 
+    const frontendConfig = {
```
jseppi: explicitly add the necessary front end config values to the index page's context

frontend/components/AddSite/TemplateSiteList/index.js
```diff
@@ -62,30 +65,30 @@ class TemplateList extends React.Component {
...
-                  handleSubmit={this.props.handleSubmitTemplate}
-                  defaultOwner={this.props.defaultOwner}
-                  {...template} />
-              </div>
-            );
-          })}
-        </div>
-      );
-    });
+    const templateGrid = templateRows.map(row => (
```
jseppi: mostly linting fixes here. I changed the `key` properties so that they wouldn't be loop-index-based as well.

frontend/components/AddSite/TemplateSiteList/index.js
```diff
@@ -102,4 +105,8 @@ class TemplateList extends React.Component {
 
 TemplateList.propTypes = propTypes;
 
-export default TemplateList;
+const mapStateToProps = state => ({
+  templates: state.FRONTEND_CONFIG.TEMPLATES,
```
jseppi: grab the `templates` config from the store!

frontend/components/branchViewLink.js
```diff
@@ -0,0 +1,47 @@
...
+
+BranchViewLink.propTypes = {
+  branchName: PropTypes.string.isRequired,
+  site: SITE.isRequired,
+  previewHostname: PropTypes.string.isRequired,
+};
+
+
+const mapStateToProps = state => ({
+  previewHostname: state.FRONTEND_CONFIG.PREVIEW_HOSTNAME,
```
jseppi: grab the `previewHostname` from the store!

frontend/reducers.js
```diff
@@ -19,4 +19,5 @@ export default {
   user,
   githubBranches,
   notifications,
+  FRONTEND_CONFIG: (state = {}) => state,
```
jseppi: kind of a no-op or pass-through reducer. we have no actions to modify the `FRONTEND_CONFIG` state, nor do we want any.

frontend/store.js
```diff
@@ -18,8 +19,16 @@ const middlewares = [
...
+// FRONTEND_CONFIG is a global variable rendered into the index.html
+// template by the Main controller. We use it to initialize our
+// store's state with configuration values from the server-side.
+const FRONTEND_CONFIG = typeof window !== 'undefined'
+  ? window.FRONTEND_CONFIG
+  : global.FRONTEND_CONFIG;
+
 const store = createStore(
   reducer,
+  { FRONTEND_CONFIG },
```
jseppi: hydrate the store's initial state with the `FRONTEND_CONFIG`

test/api/unit/models/site.test.js
```diff
@@ -42,10 +40,10 @@ describe('Site model', () => {
     });
 
     context('for the demo branch', () => {
-      it('should return an s3 demo link if there is no demo domain', (done) => {
+      it('should return a proxy demo link if there is no demo domain', (done) => {
```
jseppi: modified test to not look for s3 bucket url

views/index.html
```diff
@@ -32,6 +32,10 @@
     <title>
       Federalist<% if (siteDisplayEnv) { %> | <%= siteDisplayEnv %><% } %>
     </title>
+
+    <script>
+      window.FRONTEND_CONFIG = <%- JSON.stringify(frontendConfig) %>;
```
jseppi: here's how we get the `frontendConfig` from the server side to the front end app

jseppi:
arg, the Code Climate status reporters are gone. askdjasdjasdlasjdklasjdsad

jseppi:
Seems like the missing CodeClimate _coverage_ report was just a blip with their system. I triggered a rebuild and it worked.

But the regular CodeClimate status reporter is still missing.

jseppi:
I have no idea what is going on with CC. Maybe it is something on their end. All the integration checks I can do from the [CC dashboard](https://codeclimate.com/repos/56a67732a4a55700430119a2/settings/git_repository/edit) say that the integration and status checks are enabled and working.

jmhooper:
I reviewed this and it LGTM. I'll wait until you get back to talk through it to make sure I understand what's going on then sign off 📝

jseppi:
oh, we can go over this in person tomorrow, @jmhooper!

Issue #1105, opened 2017-08-04
URL inputs should require TLD


jseppi:
We won't have a site that is hosted at a URL without a TLD (the `.whatever` portion of the URL), and right now our back end URL validation does require a TLD while the front end validation does not. We should make the front end require it.

PR #1106, opened 2017-08-04
Require TLD in HttpsUrlInput component

jseppi:
Fixes a minor inconsistency I noticed and described in #1105. Changes our url input pattern to require a TLD of at least 2 characters.

### DIFF
--- a/frontend/components/httpsUrlInput.js
+++ b/frontend/components/httpsUrlInput.js
@@ -13,7 +13,7 @@ const HttpsUrlInput = ({ placeholder, ...props }) => (
   <input
     {...props}
     type="url"
-    pattern="https://.+"
+    pattern="https://.+\.\w{2,}"
     placeholder={placeholder}
     onInvalid={onInvalidUrl}
     onInput={onUrlInput}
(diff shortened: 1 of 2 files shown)


frontend/components/httpsUrlInput.js
```diff
@@ -13,7 +13,7 @@ const HttpsUrlInput = ({ placeholder, ...props }) => (
   <input
     {...props}
     type="url"
-    pattern="https://.+"
+    pattern="https://.+\.\w{2,}"
```
jmhooper: Do we need to add something to match path components (in case someone wants to use a path, like handbook is doing)?

jseppi: That's probably a good idea, but should be a new issue IMHO. We'd also want to validate that on the back end, and probably make it explicit in the docs.

jseppi: oic, this regex blocks paths. We want to block paths, don't we?

wslack: I don't think we can support paths in the CDN, ever. Someone would have to have a server pulling from Federalist, which isn't compliant. (so no paths is ok)

jmhooper: Okay, cool. We have some old code enabling paths in the SQS service. Also, we'll want to validate that on the backend.

jseppi:
issues created!

PR #1120, opened 2017-08-15
Server-side rendered templates and flash messages

jseppi:
Ref #958.

Worked this up on the plane yesterday ✈️. This PR moves common site elements (basic page structure, header, nav, footer, etc) into server-rendered templates and out of the React app, and adds server-side auth protection to the `/sites` view. I also switched to `nunjucks` for templating since it allows blocks and inheritance, which `ejs` does not.

Benefits include:
- not loading the front end code for every user on the home (brochure) page, 
- nice separation of logged-in vs logged-out functionality, 
- faster page loads since client side rendered elements are reduced, and 
- slightly smaller (reduced by about ~16K) JS bundle.

I think it'll also make a number of other features easier to implement.

To Do:
- [x] adjust tests
- [x] clear out TODO comments
- [x] use a message flash middleware for when login fails or when private route is visited while not authenticated
- [x] squash commit history before merging


jseppi:
huh, the "regular" Code Climate check is back in this PR:

screen shot 2017-08-15 at 6 27 57 am

jmhooper:
:+1: I am a fan of this

api/controllers/main.js
```diff
@@ -11,34 +11,62 @@ function loadAssetManifest() {
...
+  if (process.env.NODE_ENV === 'development') {
+    // reload the webpack assets during development so we don't have to
+    // restart the server for front end changes
+    webpackAssets = loadAssetManifest();
+  }
+
+  const siteDisplayEnv = config.app.app_env !== 'production' ? config.app.app_env : null;
+
+  const messages = {
+    errors: req.flash('error'),
```
jseppi: Decided to structure this way so that we could more easily add other message types (like `info`) in the future, but keep them all contained in the `messages` object.

api/models/user.js
```diff
@@ -1,37 +1,37 @@
 const associate = ({ User, Build, Site }) => {
```
jseppi: just linting fixes in this file

api/services/passport.js
```diff
@@ -60,7 +60,12 @@ passport.callback = (req, res) => {
         }
       });
     } else {
-      res.redirect('/?login_failed=1');
```
jseppi: don't need this query parameter-based messaging any more with the new message flash stuff

frontend/components/disclaimer.js
```diff
@@ -1,19 +0,0 @@
-import React from 'react';
```
jseppi: 💥

frontend/components/header.js
```diff
@@ -1,21 +0,0 @@
-import React from 'react';
```
jseppi: 💥

frontend/components/home.html
```diff
@@ -1,138 +0,0 @@
-<div id="home">
```
jseppi: 💥

frontend/sass/_actions.scss
```diff
@@ -1,3 +1,7 @@
+a {
+  text-decoration-skip: ink;
```
jseppi: snuck in this minor styling change. It makes the underline skip descenders, like on the `g` here:

screen shot 2017-08-24 at 12 27 14 pm

nodemon.json
```diff
@@ -1,5 +1,5 @@
 {
   "ignore": ["frontend/", ".tmp/", "public/", "node_modules/", "webpack-manifest.json"],
-  "ext": "js",
+  "ext": "js,njk",
```
jseppi: restart when `.njk` files change too

Issue #1127, opened 2017-08-22
Site List ordering can change, which is slightly confusing


jseppi:
There is no ordering enforced on the Site List page (https://federalist.18f.gov/sites) so sometimes the order of sites changes which is a little confusing (to me at least). 

It's a very easy fix to just order by `site.id` in the `SiteList` component.

wslack:
They change? Huh. I think this only matters to extreme power users but it it only take 5 minutes to fix, would be fine. Put in backlog for now.

jseppi:
More like 30 seconds:) I'll probably just roll it into some other PR

PR #1129, opened 2017-08-24
Order sites in SitesList ascending by id

jseppi:
Ref #1127


jseppi:
Yeah, I think we could do a lot more to make this page easier to navigate for folks, especially those with lots of sites. Maybe like grouping repos by GitHub org.

I chose `id` so it would be basically in order of when sites were added. Happy to change to something else though.

jmhooper:
Don't feel obligated. This make it consistent which was the goal.

wslack:
I honestly like id to maintain the order of addition but could maybe be persuaded if other users disagree. Most users will not have so many sites that alphabetizing is really that helpful.

Issue #1143, opened 2017-09-05
Branch preview links should allow `.` character


jseppi:
Right now only alphanumeric, _ and - are allowed. Dots (`.`) should also be allowed. Ref #1018.

PR #1148, opened 2017-09-06
allow dots `.` in branch preview links

jseppi:
ref #1143

### DIFF
--- a/frontend/components/branchViewLink.js
+++ b/frontend/components/branchViewLink.js
@@ -8,8 +8,8 @@ const isDefaultBranch = (branchName, site) => branchName === site.defaultBranch;
 
 const isDemoBranch = (branchName, site) => branchName === site.demoBranch;
 
-// we only want to link branch names that are alphanumeric plus _ and -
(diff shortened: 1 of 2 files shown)


frontend/components/branchViewLink.js
```diff
@@ -8,8 +8,8 @@ const isDefaultBranch = (branchName, site) => branchName === site.defaultBranch;
 
 const isDemoBranch = (branchName, site) => branchName === site.demoBranch;
 
-// we only want to link branch names that are alphanumeric plus _ and -
-const isLinkable = s => /^[a-zA-Z0-9_-]+$/.test(s);
+// we only want to link branch names that are alphanumeric plus _, -, and .
+const isLinkable = s => /^[a-zA-Z0-9._-]+$/.test(s);
```
jmhooper: I was gonna comment to ask if you needed to escape the dot, but it looks like it works without escaping. Didn't know you could do that.

jseppi: yeah, no escape necessary because it is within a character class (`[...]`). I think our eslint config let me know not to escape.

Issue #1142, opened 2017-09-05
Reduce padding in GitHub branches table


jseppi:
It's too big after the change of the `p` font size to 16px

jseppi:
We could either reduce the `margin-top/bottom` or get rid of the wrapping `<p>` tags. None of the other tables wrap their `<td>` content in `<p>`s, so this might be the more consistent option.

wslack:
Concur with getting rid of the <p>.

PR #1151, opened 2017-09-07
remove p tags surrounding branch names

jseppi:
and also adjust the `margin-top` of the table to bring the associated `h4.label` closer to the table (and make it consistent with form element + label spacing).

ref #1142

Now looks like this:

screen shot 2017-09-07 at 9 34 08 am


jseppi:
cc @wslack

Issue #1163, opened 2017-09-15
autobind maybe has performance issues


jseppi:
Consider using the autobind decorator instead of autobinding (from `react-autobind`) on whole Component classes

https://github.com/andreypopp/autobind-decorator#performance

jseppi:
The `react-autobind` lib actually doesn't bind _everything_ in the class it's used in. It avoids binding all the React lifecycle methods, for instance, so it probably doesn't have the perf problems outlined in `autobind-decorator`.

PR #1169, opened 2017-09-18
only bind necessary methods

jseppi:
Assuages #1163 by only binding necessary methods.

At this point `react-autobind` is just used as a slightly nicer helper to bind functions instead of having to use `this.somefunc = this.somefunc.bind(this);`.

### DIFF
--- a/frontend/components/site/siteBuilds.js
+++ b/frontend/components/site/siteBuilds.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import autoBind from 'react-autobind';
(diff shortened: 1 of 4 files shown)


frontend/components/site/siteBuilds.js
```diff
@@ -44,11 +43,6 @@ class SiteBuilds extends React.Component {
     /* eslint-enable jsx-a11y/href-no-hash */
   }
 
-  constructor(props) {
-    super(props);
-    autoBind(this);
```
jseppi: wasn't actually necessary in this component

PR #1174, opened 2017-09-21
amend props in tests to pass component proptype validation

jseppi:


test/frontend/components/site/siteBuildLogs.test.js
```diff
@@ -10,13 +10,13 @@ describe('<SiteBuildLogs/>', () => {
...
       params: {
-        buildId: 123,
+        buildId: '123',
       },
       buildLogs: {
         isLoading: false,
         data: [
-          { build: { id: 1 }, output: 'output 1', source: 'source 1' },
-          { build: { id: 1 }, output: 'output 2', source: 'source 2' },
+          { id: 1, output: 'output 1', source: 'source 1', createdAt: '2017-02-11' },
```
jseppi: @jmhooper I missed this previously and we still had a prop validation error in the test logs. Fixed in most recent commit.

Issue #1027, opened 2017-06-21
New IA for Federalist


wslack:
Current IA
<img width="330" alt="screen shot 2017-06-21 at 3 53 48 pm" src="https://user-images.githubusercontent.com/10016577/27403875-21ef0ab0-569a-11e7-9edd-9477586fb347.png">

Current Target IA:
<img width="680" alt="screen shot 2017-06-21 at 3 55 07 pm" src="https://user-images.githubusercontent.com/10016577/27403876-220def16-569a-11e7-9cb0-9f489b9a68c9.png">

wslack:
@ASaporta - per your request, are would helping to enact this be interesting? Another option is #940.

jseppi:
@wslack - I don't recall: Should the new Advanced Settings be in a totally separate page from Settings, or in a demarcated "area" (that is perhaps expandable/collapsable) of a single Settings page?

wslack:
@jseppi we said an expandable/collapsable area - so it doesn't need another set of "view site" links

jseppi:
Great - thanks!

jseppi:
Remaining tasks:

- Ability to rebuild a specific branch (from the SiteGitHubBranches table) - #1029 
- Add branch name to Build Logs view - https://github.com/18F/federalist/issues/858

PR #1180, opened 2017-09-22
Create view of site users and associated back end changes

jseppi:
- Include array of associated `User` models in `Site` responses
- Create front end view for users associated with the targeted site.

Looks like:

screen shot 2017-09-25 at 8 06 51 am

Note that there is not currently a UI link to the new user table -- that will come in a separate PR as part of the navigation changes for #1027. It can be visited directly at `/sites/<id>/users` though.

ref #1181.

### DIFF
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -61,6 +61,8 @@ function toJSON() {
     plain: true,
   }));
 
+  delete object.site_users__user_sites;
+
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
 
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -17,6 +17,7 @@ function toJSON() {
   delete object.githubAccessToken;
   delete object.githubUserId;
   delete object.signedInAt;
+  delete object.site_users__user_sites;
 
   object.createdAt = object.createdAt.toISOString();
   object.updatedAt = object.updatedAt.toISOString();
--- a/api/serializers/site.js
+++ b/api/serializers/site.js
@@ -12,9 +12,6 @@ const serializeObject = (site) => {
 };
 
 const serialize = (serializable) => {
-  // TODO: Why does this (and every other serializer) re-query for the model?
-  // In addition to seeming unnecessary, it also
-  // removes the loaded associations if any were specified
   const include = [User];
 
   if (serializable.length !== undefined) {
--- a/api/serializers/site.js
+++ b/api/serializers/site.js
@@ -1,26 +1,30 @@
-const { Site } = require("../models")
-
-const serialize = (serializable) => {
-  if (serializable.length !== undefined) {
-    const siteIds = serializable.map(site => site.id)
-    const query = Site.findAll({ where: { id: siteIds } })
-
-    return query.then(sites => {
-      return sites.map(site => serializeObject(site))
-    })
-  } else {
-    const site = serializable
-    const query = Site.findById(site.id)
-
-    return query.then(site => {
-      return serializeObject(site)
-    })
-  }
-}
+const { Site, User } = require('../models');
 
 const serializeObject = (site) => {
-  const json = site.toJSON()
-  return json
-}
+  const json = site.toJSON();
 
-module.exports = { serialize }
+  if (json.Users) {
+    json.users = site.Users.map(u => u.toJSON());
+    delete json.Users;
+  }
+
+  return json;
+};
+
+const serialize = (serializable) => {
+  // TODO: Why does this (and every other serializer) re-query for the model?
+  // In addition to seeming unnecessary, it also
+  // removes the loaded associations if any were specified
+  const include = [User];
+
+  if (serializable.length !== undefined) {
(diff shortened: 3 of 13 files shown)


api/serializers/site.js
```diff
@@ -1,26 +1,30 @@
...
-    const site = serializable
-    const query = Site.findById(site.id)
+  return json;
+};
 
-    return query.then(site => {
-      return serializeObject(site)
-    })
+const serialize = (serializable) => {
+  // TODO: Why does this (and every other serializer) re-query for the model?
```
jseppi: @jmhooper: any insight into this question? It looks like all the serializers actually end up re-fetching models from the database when invoked. This seems a little odd since they are all called with database models already, so it seems like a wasted DB call(s).

jmhooper: IIRC this was in here to maintain some Waterline behavior which would load associated objects when you called `req.json(model)`.

jseppi: hrm, ok, so I guess then it would be ok to just move that behavior into the controller methods as applicable.

Or maybe there is some way in Sequelize to tell it to always load some associations. I'll look into it.

api/models/site.js
```diff
@@ -61,6 +61,8 @@ function toJSON() {
     plain: true,
   }));
 
+  delete object.site_users__user_sites;
```
jseppi: no need to include this

api/models/user.js
```diff
@@ -17,6 +17,7 @@ function toJSON() {
   delete object.githubAccessToken;
   delete object.githubUserId;
   delete object.signedInAt;
+  delete object.site_users__user_sites;
```
jseppi: no need to include this

api/serializers/site.js
```diff
@@ -1,26 +1,27 @@
...
-    const site = serializable
-    const query = Site.findById(site.id)
+  return json;
+};
 
-    return query.then(site => {
-      return serializeObject(site)
-    })
+const serialize = (serializable) => {
+  const include = [User];
```
jseppi: Added this `include` in order to also fetch a `Site`'s associated `User`s

api/serializers/user.js
```diff
@@ -1,30 +1,26 @@
-const { Build, User, Site } = require("../models")
+const { Build, User, Site } = require('../models');
```
jseppi: just linting fixes here

frontend/components/site/SiteUsers.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
jseppi: new component to display a table of users associated with current site

frontend/components/site/SiteUsers.jsx
```diff
@@ -0,0 +1,55 @@
...
+              <td>
+                <a
+                  href={`https://github.com/${user.username}`}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  title={`Visit GitHub profile for ${user.username}`}
+                >
+                  {user.username}
+                </a>
+                {' '}
```
jseppi: little trick to force a space

frontend/components/site/SiteUsers.jsx
```diff
@@ -0,0 +1,55 @@
...
+    const bName = b.username.toLowerCase();
+    if (aName === bName) { return 0; }
+    if (aName > bName) { return 1; }
+    return -1;
+  });
+
+  return (
+    <div>
+      <h4 className="label">Federalist users associated with this site</h4>
+      <table className="usa-table-borderless">
```
jseppi: Even though it currently only has one column, I made it a table in anticipation of having some kind of "remove user" action user in the near future.

frontend/components/site/SiteUsers.jsx
```diff
@@ -0,0 +1,55 @@
...
+                <a
+                  href={`https://github.com/${user.username}`}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  title={`Visit GitHub profile for ${user.username}`}
+                >
+                  {user.username}
+                </a>
+                {' '}
+                {user.username.toLowerCase() === site.owner.toLowerCase() ? '(owner)' : ''}
```
jmhooper: The `owner` badge is interesting to me. It looks like it is displayed next to the person who's GH the repo is on. That means for sites owned by an organization there is no owner.

I was confused by this for a bit b/c I did not connect to GH before looking at the code. I'm a little concerned about it confusing users. Interested in @wslack's thoughts.

jseppi: I see your point. We could definitely remove it!

What do you think about either demarcating or even not displaying the current user (like `jseppi` in the screenshot) in the table, since they'd be implied?

jmhooper: That is also interesting 🤔. I like demarcating them, but not sure about removing them.

This will get a lot more interesting when we add the "Remove user" button.

jseppi: ok, I removed the "owner" badge and added a "you"  badge

Issue #1171, opened 2017-09-19
Add tests for SiteSettings component


jseppi:
It doesn't have any :(

jseppi:
Should do during/after the settings page changes for #1027

jseppi:
This is actually done (in #1185) and has already been part of sprint review and is in prod 🎉

Issue #1001, opened 2017-06-12
Cancel button on Site Settings page doesn't do anything 


jseppi:
It should reload the page, thus reverting all changes, I think.

wslack:
What "cancel" button?

wslack:
Oh, this cancel button:
<img width="815" alt="screen shot 2017-06-12 at 7 18 31 pm" src="https://user-images.githubusercontent.com/10016577/27059339-efc2549a-4fa3-11e7-82ed-d1014e41f27a.png">

wslack:
Agreed it should just reload

jmhooper:
Is there anything stopping us from removing the cancel button? The browser already has an 🔄 button?

jseppi:
It's kind of a nice user affordance to have the cancel button there, even
if what it actually does is cause a refresh. I think we could also
relatively easily have it just reload the state and show a status message
that that happened, but I'm not sure if it is worth it.

wslack:
Agreed. Can we say "cancel changes (reloads page)" as the text or is there something shorter we could use?

jseppi:
This has been fixed as of #1185. Closing.

Issue #908, opened 2017-05-12
Merge custom domain and "default branch" config in the settings screen


wslack:
The "demo" options are clearer than the legacy options for live sites.

PR #1185, opened 2017-09-25
Split Site Settings to Settings and Advanced Settings views

jseppi:
- Split settings into two forms/components: `BasicSiteSettings` and `AdvancedSiteSettings` that are both on the same page (wrapped in the `SiteSettings` component)
  - `AdvancedSiteSettings` are wrapped in an `Accordion`
- Use `ReduxForm` for redux-based form interactions instead of the previous custom `state`-based implementation
- Remove the `SiteGitHubBranches` table from the settings page
- Change broken (ref #1001)  "Cancel" button to "Reset" button that will reset the values in the form when clicked

Looks mostly like this except now the Advanced Settings section looks more like a USWDS accordion:

new-settings

To Do:
- [x] confirm interaction is correct
- [x] bug: clicking anywhere in the Advanced Settings section collapses the accordion
- [x] look into using [USWDS accordion](https://standards.usa.gov/components/accordions/)
- [x] reorganize default branch and custom domain into a well, similar to demo branch/demo domain (ref https://github.com/18F/federalist/issues/908)
- [x] tests! (related: #1171)

Ref #1184

Part of #1027


jseppi:
@wslack: want to check out the interaction here? Or I can do a vid call demo for you since that'll probably be more information than watching the gif above.

@jmhooper: I still need to finish up the tests, but the majority of the implementation is done.

jseppi:
hrm, that is annoying that the total coverage went down. Looking into it.

jseppi:
I better tested the base `SiteSettings/index.jsx` component to bring the coverage back up.

Issue #866, opened 2017-04-17
Possible for commits to fail because "Payload Too Large"


wslack:
Whatever library handles inbound webhooks replied with "Payload Too Large" to a 136kb payload. That's unusually large, but it would be nice for Federalist to note a "bad webhook" or something in the builds page, or to have guidance in the docs about these types of issues.

jmhooper:
One tripping point may be that in order to know which site the webhook is for, we need to process the too request body, which is too large.

Another note: It looks like this error [originates in the body-parser middlerware](https://github.com/expressjs/body-parser#request-entity-too-large). We can adjust the `limit` and `length` properties to allow for a larger request body.

wslack:
hmm any harm in adjusting those?

jmhooper:
We should be able to tune them within a reasonable range.

If we go absurdly high then we open ourselves up to a DOS while parsing a huge JSON request body.

wslack:
@shawnbot what's the largest hook you can picture sending us?

shawnbot:
Honestly, I don't know. 200K? How long is too long to be parsing JSON? @jmhooper maybe you could consider using a streaming JSON parser, like [JSONStream](https://github.com/dominictarr/JSONStream)?

wslack:
@jmhooper will _do the math_ on what size we can bump to. This has never affected external customers.

shawnbot:
I could see this being a problem on larger projects, particularly when merging big change sets from, say, staging or develop branches to `master`. Just something to consider.

wslack:
@shawnbot I think that large customers can always to a tiny nonsense commit as a workaround, but hat's not optimal.

shawnbot:
Yeah, or you could manually restart the build. Without any feedback (see #663) it's hard to know what went wrong, though. Also, nonsense commits (especially on the production branch) are generally not good. 😉

PR #1200, opened 2017-09-28
increase bodyParser json size limit to 2mb

jseppi:
to mitigate nearly silent github webhook failures due to payloads being larger than the default size [limit of 100kb](https://github.com/expressjs/body-parser/blob/master/README.md#limit-1)

cc @konklone 

ref https://github.com/18F/federalist/issues/866


jseppi:
It would be cool if we could a max payload size that GitHub would send, but I haven't had any luck finding such a limit in their documentation or elsewhere.

PR #1177, opened 2017-09-21
[WIP] Navigation updates for new IA

jseppi:
**work-in-progress**

ref #1027

To Do:
- [x] test - Original coverage is only at 60% so we should bring that up as part of this PR.

PR #1201, opened 2017-10-02
add users link to sidenav

jseppi:
This PR includes some of the changes from #1177 in order to add the "users" link to the SideNav. I wanted to get that link in before today's review/deploy so that folks would be able to navigate to the new Users page.

### DIFF
--- a/frontend/components/site/SideNav/sideNav.js
+++ b/frontend/components/site/SideNav/sideNav.js
@@ -2,17 +2,18 @@ import React from 'react';
 import SideNavItem from './sideNavItem';
 
 const propTypes = {
-  siteId: React.PropTypes.string
+  siteId: React.PropTypes.number.isRequired,
 };
 
 export const sideNavPaths = {
   SETTINGS: 'settings',
+  USERS: 'users',
   BUILDS: 'builds',
   PUBLISHED: 'published',
 };
 
 class SideNav extends React.Component {
-  getUrl(id, path='') {
+  getUrl(id, path = '') {
     // strip trailing slashes, just in case.
     return `/sites/${id}/${path}`.replace(/\/$/, '');
   }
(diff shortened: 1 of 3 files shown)


frontend/components/site/SideNav/sideNav.js
```diff
@@ -25,17 +26,22 @@ class SideNav extends React.Component {
...
           />
           <SideNavItem
+            href={this.getUrl(siteId, sideNavPaths.USERS)}
+            icon="gear"
+            linkText={sideNavPaths.USERS}
+          />
+          <SideNavItem
             href={this.getUrl(siteId, sideNavPaths.PUBLISHED)}
-            icon='media'
+            icon="media"
```
wslack: why do we need double quotes and not single?

frontend/sass/_icons.scss
```diff
@@ -97,3 +97,16 @@ $twelfth-column: 0;
...
+.icon.icon-gear {
+  &:before {
+    @include icon();
+    height: 1em;
+    background-position: $second-column $second-row;
+    content: 'Gear icon';
+  }
+
+  &:hover:before {
+    background-position-x: $fifth-column;
```
wslack: why is this needed?

wslack:
@jseppi Jonathan is in a login.gov fire so I checked it. Two questions.

frontend/components/site/SideNav/sideNav.js
jseppi: That's a linting thing. In JSX, we use double-quotes for attribute values because that is what our preferred config (`eslint-config-airbnb`) is set to use. Either way doesn't matter, but consistency is good so we just go with what the config specifies :)

frontend/sass/_icons.scss
jseppi: that's for "swapping" out the black icon to the blue icon on mouse hover.

wslack: ah! clever.

PR #1213, opened 2017-10-03
update link color to allow for required contrast

wslack:


maya:
This works, but if you want to use a darker tone of the existing color, you could use: #007d98. If you have a visual designer on hand to check that would be nice, otherwise whichever you prefer.

wslack:
I'll check in staging after merge, I thank. Thanks!

jseppi:
With color @wslack used:
slack color

With color @maya suggested:
maya color

jseppi:
I prefer the first option

wslack:
@jseppi do you think its "too blue" compared to the more gray colors on the page?

jseppi:
Looks fine to these untrained eyes

Issue #1206, opened 2017-10-03
Branch input styling is different from url input styling


jseppi:
Minor issue.

The branch inputs have a wide light gray outline, while the URL inputs have a thin black outline. They should probably be consistent:

<img width="490" alt="screen shot 2017-10-03 at 4 22 16 pm" src="https://user-images.githubusercontent.com/697848/31149887-31cc2666-a857-11e7-937e-6e53b42494f7.png">

wslack:
Is this how the USWDS handles text boxes vs URL boxes, @maya?

maya:
USWDS default styling is the latter, but I think a previous designer styled it to look like the former (thicker outline).

wslack:
Cause is here: https://github.com/18F/federalist/blob/staging/frontend/sass/_forms.scss#L15

jseppi:
Thanks. Just needs a `input[type="url"],` added there

wslack:
Wouldn't it be `input[type="text","url"]`?

jseppi:
No, that won't work. Needs to be a separate target.

wslack:
I think I prefer the USWDS styling. Do you like the gray more?

jseppi:
No preference other than consistency. If we change back to default, we also should change the `<textarea>` styling back to that as well.

PR #1214, opened 2017-10-04
make URL field look like other fields

wslack:
for #1206

### DIFF
--- a/frontend/sass/_forms.scss
+++ b/frontend/sass/_forms.scss
@@ -11,11 +11,7 @@ fieldset {
 }
 
 input[type="text"],
-textarea {
-  border: 3px solid #ccc;
-}
-
-input[type="url"]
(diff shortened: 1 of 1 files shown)


frontend/sass/_forms.scss
```diff
@@ -15,6 +15,11 @@ textarea {
   border: 3px solid #ccc;
 }
 
+input[type="url"]
```
jseppi: Just move the `input[type="url"]` target up to line `14`. You don't need to recopy the `textarea` target and style rule. So, final result, starting at current line 13 should be:

```css
input[type="text"],
input[type="url"],
textarea {
  border: 3px solid #ccc;
}
```

jseppi:
screen shot 2017-10-04 at 12 06 55 pm

Issue #834, opened 2017-03-29
Reorganize Federalist Menus with card sort results


wslack:

wslack:
@wslack to record outcomes from Murally.

wslack:
![screen shot 2017-07-31 at 12 16 16 pm](https://user-images.githubusercontent.com/10016577/28787409-1540f390-75ea-11e7-979c-86df2e5e7738.png)

wslack:
Order:

Build History
GitHub Branches
Data on S3
Users
Settings

jseppi:
Done as part of #1216

PR #1216, opened 2017-10-05
Implement Site GitHub Branches view and finish Side Nav changes

jseppi:
- create SiteGitHubBranches component and tests
- update and refactor SideNav component and add tests
  - set the default site route to "builds" instead of "settings"
  - reorder SideNav links

Looks like this:

screen shot 2017-10-06 at 11 58 04 am

To Do:
- [x] make build history the default view instead of site settings
- [x] reorder SideNav links
- [x] test that github branch fetch happens in component mount
- [x] in the branches table, order the rows so first one is the default branch, and the second one is the demo branch (if it exists)
- [x] change the wording in the "error" state to be closer to the previous branch table's wording:

```js
const renderErrorState = () => (
  <p>
    An error occurred while downloading branch data from Github.
    Often this is because the repo is private or has been deleted.
  </p>
);

const SiteGithubBranchesTable = ({ site, branches }) => {
  if (branches.isLoading) {
    return renderLoadingState();
  } else if (branches.error || !branches.data || branches.data.length === 0) {
    return renderErrorState();
  }
  // ...
```

ref #834 and #1217


frontend/actions/actionCreators/githubBranchActions.js
```diff
@@ -1,23 +1,23 @@
-const githubBranchesFetchStartedType = "SITE_GITHUB_BRANCHES_FETCH_STARTED"
```
jseppi: just linting fixes here

frontend/actions/githubBranchActions.js
```diff
@@ -1,30 +1,29 @@
-import github from '../util/githubApi'
```
jseppi: just linting fixes

frontend/actions/siteActions.js
```diff
@@ -11,14 +11,14 @@ import {
 } from './dispatchActions';
```
jseppi: just linting fixes

frontend/components/GitHubRepoLink.jsx
```diff
@@ -3,21 +3,36 @@ import PropTypes from 'prop-types';
...
-    rel="noopener noreferrer"
-  >
-    <GitHubMark />
-  </a>
-);
+const GitHubRepoLink = ({ owner, repository, branch }) => {
+  let href = `https://github.com/${owner}/${repository}`;
+  let title = 'View repository';
+
+  if (branch) {
```
jseppi: modified this component to render a link to branch on github if the `branch` prop is supplied

frontend/components/site/SideNav.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router';
+
+const SideNav = ({ siteId }) => (
```
jseppi: simplified this component by just listing the links explicitly instead of via a config and child components

wslack:
THIS IS SO GREAT!

wslack:
@jseppi are you going to flip the build history and the settings screen in another PR?

jseppi:
Oh whoops, forgot about that. Will do it here -- should be very simple.

jseppi:
Updated the screenshot with the new side navigation ordering, @wslack

jseppi:
I modified the order you had in https://github.com/18F/federalist/issues/834 very slightly to put `Data on S3` above `GitHub Branches`.

wslack:
Any reason to flip? I was trying to make an order according to how often they would be needed but am rather agnostic.

jseppi:
heh, same reason I changed it. I thought `Data on S3` would be used more, but I'm admittedly less of a user than you are. Happy to change it!

wslack:
Yep, let's please flip it back; users will use the branch list much more . Can we also call it "Published Files" instead of "Data on S3" please?

jseppi:
done!

wslack:
Excellent. Now just pending review.

frontend/components/site/SiteGitHubBranches.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+import { SITE, GITHUB_BRANCHES } from '../../propTypes';
+import LoadingIndicator from '../loadingIndicator';
+import GitHubRepoLink from '../GitHubRepoLink';
+import BranchViewLink from '../branchViewLink';
+import githubBranchActions from '../../actions/githubBranchActions';
+
+export class SiteGitHubBranches extends React.Component {
+  componentDidMount() {
+    githubBranchActions.fetchBranches(this.props.site);
```
jseppi: I'm not a huge fan of putting this here, but couldn't figure a cleaner place to put it. I wanted to put it in the associated route's `onEnter` prop, but couldn't because it relies on `site`, which may have not been loaded at the time the route is visited.
A better side-effect library (like `redux-sagas`) might help out with this kind of thing, but this is the only weird place right now, so I punted on it.

frontend/routes.js
```diff
@@ -9,17 +9,26 @@ import SiteBuildLogs from './components/site/siteBuildLogs';
...
+import userActions from './actions/userActions';
+
+const fetchInitialData = () => {
+  userActions.fetchUser();
+  siteActions.fetchSites();
+};
+
 export default (
-  <Route path="/" component={App}>
+  <Route path="/" component={App} onEnter={fetchInitialData}>
```
jseppi: I found out about this `onEnter` callback while reading the react-router docs and thought this was a cleaner place to initiate the data fetch than where it was in `main.jsx`

jmhooper: Nice, that's fancy 🤵

wslack:
I'm so excited to see this on `staging`.

frontend/components/site/SiteGitHubBranches.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!site || githubBranches.isLoading) {
+      return <LoadingIndicator />;
+    }
+
+    if (!githubBranches.data || !githubBranches.data.length) {
+      return (
+        <p>
+          No branches were found for this repository.
+          There may have been an error communicating with the GitHub API.
```
jmhooper: Should we put something in here about private repos?

I remember seeing something about adding a recommendation about private repositories to our docs. If something like that happens, another idea may be to link to that here.

jseppi: Good question. I'm actually not sure what would happen with a private repo!

jmhooper: The GH API renders a 404.

In the old branch table, we handled the error by displaying a message that said something along these lines with a note that a private repo could also be to blame.

jmhooper:
I like this a lot and am happy that a lot of the code is much cleaner after this.

One thing that would be nice is if the branches were ordered with the default branch at the top, then the demo, and then the rest. That's how it's done on the published branches table view. But that's def a lagniappe though and I'll approve if you don't want to add it in this PR.

frontend/components/site/SiteGitHubBranches.jsx
jseppi: Ah, cool. I'll change the wording and logic to be closer to the previous table's handling of this case.

jseppi:
Great idea about the ordering. I'll make that change in addition to "error state" wording change.

test/frontend/components/site/SiteGitHubBranches.test.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const wrapper = shallow(<SiteGitHubBranches {...props} />);
+    expect(wrapper.find('table')).to.have.length(0);
+    expect(wrapper.find('p')).to.have.length(1);
+    expect(wrapper.find('p').contains(
+      'No branches were found for this repository. There may have been an error communicating with the GitHub API.')
+    ).to.be.true;
+  });
+
+  it('should render an loading state site is null', () => {
```
jseppi: typo: `an loading` -> `a loading`

wslack: fixed this one

jseppi:
How's this look, @wslack and @jmhooper?

I added little parenthesized-text on the special branches to differentiate them a little:

screen shot 2017-10-06 at 10 50 51 am

wslack:
endorsed!

jseppi:
Alright, changes made, @jmhooper!

Issue #1115, opened 2017-08-09
UX for adding a repo for an existing Federalist site is confusing


jseppi:
When a user would like to add an _existing_ Federalist site to their Federalist account, the UX is a bit confusing because the user is asked to specify the "Static Site Engine" and "Default Branch" in addition to the repository information. This is a bit scary to the user, and can cause problems if they select values different from those already set for the existing site.

An easy improvement would be to remove those two fields completely from the Add New Site page, and instead have them only on the site settings page. I think we should be able to default the values to `Jekyll` and `master` for sites that are actually "new" to Federalist.

wslack:
Let's add an interstitial when people actually add new sites to get their branch to be published and build engine. Otherwise, keep it simple.

jseppi:
Noting idea so I don't forget:

Upon submitting org and repo name, check response from back end to see if it is a new site or not.
If it is a new site, then display the default branch and build engine form fields before allowing user to continue.
If it is not a new site, then continue on the normal flow.

PR #1221, opened 2017-10-06
Improve Add Site UX

jseppi:
First show the owner and repository fields in the Add Site flow.
If a site exists in Federalist (that the user can access), then add the current
user to that site.
Otherwise, present the user with the engine and primary branch fields so
that a completely new site will be added to Federalist.

also:
- extract AddRepoSiteForm and use redux-form for it
- create new back end method to add a user to a Federalist site based on owner/repo
- lots of linting and tests

To Do:
- [x] BUG: Site engine selection does not seem to be saved. Ie, "jekyll" is selected during add but "static" shows up in the added site's settings.
- [x] specifically look for a `404` in https://github.com/18F/federalist/pull/1221#discussion-diff-144615332R43

ref #1115

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,6 +1,7 @@
 {
   "reporter": [
-    "lcov"
+    "lcov",
+    "html"
   ],
   "extension": [
     ".jsx"
--- a/api/services/SiteCreator.js
+++ b/api/services/SiteCreator.js
@@ -1,152 +1,157 @@
-const GitHub = require("./GitHub")
-const config = require("../../config")
-const { Build, Site, User } = require("../models")
+const GitHub = require('./GitHub');
+const config = require('../../config');
+const { Build, Site, User } = require('../models');
 
-const createSite = ({ user, siteParams }) => {
-  const template = siteParams.template
-  siteParams = paramsForNewSite(siteParams)
+function paramsForNewSite(params) {
+  return {
+    owner: params.owner ? params.owner.toLowerCase() : null,
+    repository: params.repository ? params.repository.toLowerCase() : null,
+    defaultBranch: params.defaultBranch,
+    engine: params.engine,
+  };
+}
 
-  if (template) {
-    return createSiteFromTemplate({ siteParams, template, user })
-  } else {
-    return createSiteFromExistingRepo({ siteParams, user })
+function templateForTemplateName(templateName) {
+  const template = config.templates[templateName];
+  if (!template) {
+    throw new Error(`No such template: ${templateName}`);
   }
+  return template;
 }
 
-checkExistingGithubRepository = ({ user, owner, repository, site }) => {
-  return GitHub.getRepository(user, owner, repository).then(repo => {
-    if (!repo) {
-      throw {
-        message: `The repository ${owner}/${repository} does not exist.`,
-        status: 400,
-      }
-    }
-    if (!repo.permissions.admin && !site) {
-      throw {
-        message: "You do not have admin access to this repository",
-        status: 400,
-      }
-    }
-    if (!repo.permissions.push) {
-      throw {
-        message: "You do not have write access to this repository",
-        status: 400,
-      }
-    }
-    return true
-  })
+function paramsForNewBuildSource(templateName) {
+  if (templateName) {
+    const template = templateForTemplateName(templateName);
+    return { repository: template.repo, owner: template.owner };
+  }
+  return null;
 }
 
-checkForExistingSiteErrors = ({ site, user }) => {
-  const existingUser = site.Users.find(candidate => candidate.id === user.id)
+function paramsForNewBuild({ user, site, template }) {
+  return {
+    user: user.id,
+    site: site.id,
+    branch: site.defaultBranch,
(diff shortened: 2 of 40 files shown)


jseppi:
ready for review, @jmhooper. This turned out to be pretty huge. I tried to keep most of the linting changes in a one commit, and all the functional changes in another.

jseppi:
omg coverage went _down_ somehow ?!?1

wslack:
That makes no sense given https://codeclimate.com/github/18F/federalist/compare/1e0dba7c5bd4a1b561f7fc46ba0b27e831692fcc...8aa747aaf0403deb25b39f97112e34df484cc904#ratings

jseppi:
Unfortunately that's the quality ratings (ie, no lint errors) for those files, rather than test coverage. Test coverage is in a different tab: https://codeclimate.com/github/18F/federalist/coverage?sort=covered_percent&sort_direction=asc

I see some places that I missed now. I'll add more coverage later. I think the "meat" of this PR is good to be reviewed though.

jseppi:
I added more tests. I did, in fact, miss some new stuff in the tests I wrote. All fixed now -- and even added some additional tests for older code. The Code Climate test coverage PR status check isn't showing up for some reason. Here's the report I generated locally, though:

screen shot 2017-10-13 at 8 19 59 am

jseppi:
oh, now the PR check showed up! Up to 96% coverage!

api/services/SiteCreator.js
```diff
@@ -1,152 +1,184 @@
-const GitHub = require("./GitHub")
```
jmhooper: Not a blocker, but I think there's a good opportunity to split this out into 2 services. One for creating sites and one for adding a user to a site.

jseppi: yeahhhhhhh....that whole SiteCreator thing is pretty huge. I was trying to not mess with it as much as possible since it is pretty intricate. Could definitely stand for some refactoring.

frontend/actions/siteActions.js
```diff
@@ -31,6 +33,16 @@ export default {
...
   },
 
+  addUserToSite({ owner, repository }) {
+    return federalist.addUserToSite({ owner, repository })
+      .then(dispatchUserAddedToSiteAction)
+      .then(updateRouterToSitesUri)
+      // rather than display an alert error for this action
+      // we'll instead want to show the additional fields necessary
+      // for adding a completely new site to Federalist
+      .catch(dispatchShowAddNewSiteFieldsAction);
```
jmhooper: We may want to look out for bugs here. Could we check the error code to make sure it's a 404 or something like that? I don't think we want to swallow other bugs, like "you don't have read access" or something like that.

jseppi: That's a good idea. The way it is currently written it won't get a read access bug, but there maybe could be a case where something else server side throws an exception (like some bad code that triggers a `500` or something).

api/services/SiteCreator.js
jmhooper: I actually don't see where we render an error if a site already exists?

jseppi: If a site already exists, the user gets added to the site (_if_ they have the valid github permissions). It is the same behavior that we already had.

jseppi: If they don't have the proper GitHub permissions, then then the error alert gets shown on the Site List page that the user is redirected to (same behavior as before).

jmhooper: Ah got it, so the site fails to validate because it's missing the `engine` and `defaultBranch` props. That could get sticky even if we check the status code since a validation error on `owner` or `repo` would cause these fields to be displayed 🤔

jmhooper: We do have some frontend validations on those though... Maybe something we could tie up if we come back after this PR and refactor the SiteCreator?

jseppi: I'm not quite following.
When just the owner+repo fields are shown, the submission triggers the "add user to site" flow. If that fails (because the site doesn't exist in Federalist), then the form changes into the regular Add New Site form (with the owner, repo, engine, branch fields), and submitting it triggers the regular "add site" flow.

jmhooper: Okay, I've been thinking about it backwards. I see now. I think we're good as long as we check for a 404. I still think there's a chance for a permissions error to get swallowed if we don't.

jmhooper: ...since `.addUserToSite` calls `. checkExistingGithubRepository` which will raise a permissions error and pass that down to the frontend.

jseppi: yeah, you're right

jseppi:
@jmhooper: I've addressed issues found during review. Ready for another look!

.nycrc
```diff
@@ -1,6 +1,7 @@
 {
   "reporter": [
-    "lcov"
+    "lcov",
+    "html"
```
jseppi: This outputs a handy local coverage report in the `/coverage` folder that you can explore in your web browser to see all the coverage info.

Issue #1222, opened 2017-10-11
Change "add site" to use a link to the repo


wslack:
Instead of this, we should just ask for a URL to the GitHub repo:

![screen shot 2017-10-11 at 12 42 07 pm](https://user-images.githubusercontent.com/10016577/31453878-9e4a08f0-ae81-11e7-91b1-7beafc59fff9.png)

jseppi:
We can split the repo URL on `/` and grab the last two items (which correspond to the `owner` and `repo` values.

wslack:
👍  (one of those obvious things we should have done a while ago)

We should give an example when changing thing, like:

`Example: https://github.com/18F/federalist`

PR #1224, opened 2017-10-13
Change to using a GitHub repo URL instead of owner+repository inputs in Add Site

jseppi:
ref: #1222 

Looks like:

screen shot 2017-10-20 at 9 33 16 am

screen shot 2017-10-20 at 9 33 25 am

To Do:
- [x] test validator functions
- [x] test renderField function/component
- [x] figure out how to test that validators are actually being used in the main component


jseppi:
@jmhooper: most relevant changes are in https://github.com/18F/federalist/pull/1224/commits/d16dda695a5a246d6503a081a589dff500965ae5

The other commits are just some "bonus" component clean-ups.

frontend/components/GitHubRepoUrlField/RenderField.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RenderField = ({
```
jseppi: This is a bit of a weird separation of components in my opinion, but I couldn't figure a better way to setup for testing.

jmhooper: Yeah, their may be another name we can use besides `RenderField`, like `GithubUrlInput` or something like that?

jmhooper:
Are the changes to `HttpsUrlInput` changes coincidental? I didn't see anywhere it was being used differently?

jmhooper:
Ah, I noticed the component cleanup comment.

frontend/components/GitHubRepoUrlField/RenderField.jsx
jseppi: yeah, I picked `RenderField` since that is the prop it is passed into in the "main" component. Also it's in a sub-directory, so I think keeping it called `RenderField` for this and future similar components might be good for consistency.

jmhooper: Cool deal, if you like it then it works for me.

test/frontend/components/GitHubRepoUrlField.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  it('has validators specified', () => {
+    const props = {
+      id: 'boopId',
+      name: 'boopName',
+    };
+
+    const wrapper = shallow(<GitHubRepoUrlField {...props} />);
+    const validateProp = wrapper.props().validate;
+    expect(validateProp.length).to.equal(2);
```
jseppi: I also don't love this way of testing that the validation functions are "hooked" up to the component. Couldn't think of anything better unfortunately.

jseppi:
Sweet, looks like the new combined test coverage reporting worked 🎉 

The `total-coverage` went down because now it is properly including the server side test coverage, not because this PR in particular is missing tests (there were no server side changes in this PR).

Issue #1107, opened 2017-08-06
Coverage reports only include the front end tests


jseppi:
Since the front end tests are run second by our `yarn test` task, they overwrite and take the place of the back end tests.

I can't figure out a way to get `nyc` to make different coverage reports, which I think would be a solution to this problem. Seems like this issue (https://github.com/istanbuljs/nyc/issues/531) captures the missing functionality.

wslack:
If this is fixed, we can put in controls to alert us when we lose significant test coverage.

jseppi:
Got some advice in #javascript:

* from @cmc333333: Can submit multiple reports to CodeClimate via their CLI. See https://github.com/18F/omb-eregs/blob/master/.circleci/config.yml#L71-L75
* from @LinuxBozo: Use this to merge multiple reports: https://github.com/mweibel/lcov-result-merger

jseppi:
I think @LinuxBozo's suggestion won't be totally straightforward because `nyc` doesn't allow specification of different coverage output filenames (istanbuljs/nyc#531), so it would require some more scripting on our part. i.e., run back-end tests, rename coverage file, run front-end tests, merge coverage reports.

PR #1227, opened 2017-10-20
Combined test coverage

jseppi:
Attempt to create, combine, and submit coverage reports for both our client-side and server-side tests. 🤞 

Currently open against `1222-repo-url-input` (#1224). 

ref #1107


jseppi:
_Looks_ like it worked: 

screen shot 2017-10-20 at 10 01 28 am

I think this either has to make it into the `staging` branch or be re-targeted against the `staging` branch to really see if the PR status checks work.

PR #1236, opened 2017-10-26
add report template

wslack:


jseppi:
Looks like this:

screen shot 2017-10-26 at 11 57 13 am

wslack:
We really need to make sure that the boxes behaving more intelligently

jseppi:
How would you like them to look?

wslack:
I thought we had it set up to automatically grid them if there were 4. It's also wonly looking because they are different sizes. Not a huge deal though.

PR #1246, opened 2017-10-27
Break SiteMembershipCreator off of SiteCreator

jmhooper:
The SiteCreator service was doing double duty, creating sites and adding
users to sites. This commit creates a new service and moves the code to
add a user to a site into that service.

One change in the API as a result of this is that now attempting to
create a site that already exists renders an error instead of adding a
user to the site. The addUser endpoint will need to be used now to add
users to a site.

### DIFF
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -1,0 +1,55 @@
+const GitHub = require('./GitHub');
+const { Site, User } = require('../models');
+
+const checkGithubRepository = ({ user, owner, repository }) =>
+  GitHub.getRepository(user, owner, repository).then((repo) => {
+    if (!repo) {
+      throw {
+        message: `The repository ${owner}/${repository} does not exist.`,
+        status: 400,
+      };
+    }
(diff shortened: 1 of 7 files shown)


api/services/SiteMembershipCreator.js
```diff
@@ -0,0 +1,55 @@
+const GitHub = require('./GitHub');
+const { Site, User } = require('../models');
+
+const checkGithubRepository = ({ user, owner, repository }) =>
+  GitHub.getRepository(user, owner, repository).then((repo) => {
+    if (!repo) {
+      throw {
```
jseppi: Looks like this condition is not covered by tests:

screen shot 2017-10-30 at 9 47 59 am

jmhooper: Ah, good catch. I'll put one in there.

jseppi: Not sure if I mentioned it to you, but html coverage reports are written to your local `coverage/` directory. You can launch a static html server in that directory to explore the reports. Very helpful!

PR #1256, opened 2017-10-31
update location of report template

wslack:
The new site is live but will not be set to serve at the correct address until this is merged.

### DIFF
--- a/config/templates.js
+++ b/config/templates.js
@@ -1,5 +1,5 @@
 module.exports = {
-    report: {
+  report: {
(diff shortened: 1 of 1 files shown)


jseppi:
Looks like you have some formatting issues to fix. I think deleting two spaces before `report:` and two spaces before `uswds:` should do it.

wslack:
Going to go ahead and merge this; just need to remember that the link won't work until the federalist config change.

Issue #1270, opened 2017-11-09
GitHub Repo URL input doesn't allow periods `.`


jseppi:
Which prevents repos like `AminPIC/meganlooff.github.io` from being added.

jseppi:
@wslack - going to close this one since we already validated and deployed.

PR #1271, opened 2017-11-09
allow periods in org and repo names

jseppi:
ref #1270


wslack:
Does the dot match any character except line breaks?

wslack:
Answered my own question, this is good.

jseppi:
@wslack not when it is inside of `[` `]`. It is just taken literally as a `.` in that case.

wslack:
let's merge!

PR #1273, opened 2017-11-09
Update templates.js

davidewarren:
Update titles - change single page and multi page to - landing page and modern team


jseppi:
This looks like it was made directly against `master` instead of `staging`. Generally we open all PRs against `staging`, then go over changes during Sprint Review and make a single "deploy" PR to `master`.

PR #1299, opened 2017-11-21
remove Docker dependency in README

jseppi:
Docker is not needed for running this application locally.


jseppi:
Makes sense. Should be covered now by https://github.com/18f/federalist-garden-build-py#development

PR #1300, opened 2017-11-21
add "contact us" link

wslack:
Per @line47


jseppi:
oooooh I think we could make that `90` a "live" value like our number of builds in the past week.

wslack:
Alas, no - the number of live sites is inherited from compliance, not from any old repos that were added.

wslack:
however, we can update that every two weeks

line47:
That would be rad if we could use a `live` number

Issue #1251, opened 2017-10-27
After adding a site, the Site List page is scrolled to the bottom


jseppi:
The scroll position should be at the top so that errors that happened during the add will be visible.

Not sure why it is scrolled to the bottom.

wslack:
Maybe because that's where the site should be listed if everything worked?

line47:
When I was demoing Federalist for @micahtaylor and @jenniferthibault the thought came up about where you are taken after creating a new website. The consensus was that it would be awesome to land on the  **builds page** after a new site creation instead of the *my sites* list.

jseppi:
That sounds like a good idea to me -- and it would be very easy to implement

wslack:
I'm good for trying this; however, jumping immediately will not show a build - since the first operation is to clone the repo, which takes a few seconds.

Could we combine this with a message if the table is blank explaining what is happening?

line47:
I ❤️ that idea @wslack a message showing something is happening would be 💯

wslack:
If we can do this in under an hour, 👍 . If longer, let's focus on the frontpage and converting more leads.

wslack:
This worked beautifully in staging.

jseppi:
Just tested myself, too! Nice!

PR #1317, opened 2017-11-30
Redirect to new site's builds view after add

jseppi:
ref https://github.com/18F/federalist/issues/1251#issuecomment-348259193

Looks like this when there are no builds:

screen shot 2017-11-30 at 5 39 03 pm

To Do:
- [x] Show message if the table is blank explaining what is happening (per https://github.com/18F/federalist/issues/1251#issuecomment-348291650)


line47:
Is there a class we could use to draw attention to that text? Something like this?

screen shot 2017-12-01 at 10 58 55 am

jseppi:
Yeah, that's a good idea. None of the other similar messages have any styling applied to them though. I think that should be handled in a separate issue/PR so we can do them all at once/consistently.

line47:
@jseppi that sounds good I'll add an issue now

Issue #1266, opened 2017-11-09
Repo name and github icon link overrun when owner+repo is too long


jseppi:
Happens on all `/sites/` pages:

<img width="813" alt="screen shot 2017-11-09 at 8 24 07 am" src="https://user-images.githubusercontent.com/697848/32610148-793da53c-c527-11e7-8b90-689749615c5b.png">

jseppi:
Might be another good one for @line47 to get his feet wet.

line47:
That sounds good to me !
Related to this I was confused at first to find that the github icon was in fact a link to the repo. I think this could be made more clear buy associating some text like "View the repo" or something similar next to the icon.

jseppi:
Great idea!

wslack:
Hm. I wouldn't want to do that all over the site since it will take up so much space, but we should have microcopy to explain this somewhere.

jseppi:
here's how CodeClimate does it:

![githublink](https://user-images.githubusercontent.com/697848/32619888-9cc02ede-c541-11e7-9d10-4664bd3830fd.gif)

line47:
Here is a before/after screenshot of moving a few things around to give more breathing room to the repo name and titles closer with navigation items and page content.

I've included some layout thoughts below: 

### Before screenshot
![screen shot 2017-11-09 at 11 58 56 am](https://user-images.githubusercontent.com/1449852/32619996-6a08f2e2-c54a-11e7-835d-948df83ba5de.png)

### After screenshot
![screen shot 2017-11-09 at 11 59 20 am](https://user-images.githubusercontent.com/1449852/32620002-70fdc172-c54a-11e7-8cdc-1905738d056f.png)

**Here is what is different in the after screenshot**
- Moved the view website button to the left hand nav to give some space for the title and also group the view website with other site actions
- Moved page title of "builds" down closer to the table listing the builds and also updated the naming to match the title in the left hand nav "Build history" 
- Moved the link to github onto it's own line under the site title and included some text "View repo" to make more clear that there is a link to follow.

line47:
Currently, this is hard to test locally as I don't have any test data :(

jseppi:
You should be able to add your own repositories to your local instance. So you could create a repo called `line47/super-duper-really-long-repo` and add it to your local Federalist.

line47:
Awesome! Thanks!

line47:
[...]

PR #1319, opened 2017-11-30
Display commit SHA in build table

el-mapache:
* Adds link to the commit

**Screenshot**
screen shot 2017-12-01 at 9 08 38 am

### DIFF
--- a/test/frontend/components/site/SiteGitHubBranches.test.jsx
+++ b/test/frontend/components/site/SiteGitHubBranches.test.jsx
@@ -29,15 +29,24 @@ describe('<SiteGitHubBranches />', () => {
     };
 
     const wrapper = shallow(<SiteGitHubBranches {...props} />);
+
     expect(wrapper.exists()).to.be.true;
     expect(wrapper.find('table')).to.have.length(1);
     const rows = wrapper.find('tbody tr');
+
     expect(rows).to.have.length(2);
     expect(rows.at(0).find('td').at(0).text())
       .to.have.string('branch-one');
     expect(rows.at(1).find('td').at(0).text())
       .to.have.string('branch-two');
-    expect(rows.find('GitHubRepoLink')).to.have.length(2);
+
+    // TODO: Nested component selectors, like the `Connect(BranchViewLink)` below, don't
+    // work in this case because the rendered component has  name of `Component`, which enzyme
+    // cant seem to find. Figure out why this is happening
+   expect(rows.find({
+      owner: props.site.owner,
+      repository: props.site.repository
+    })).to.have.length(2);
     expect(rows.find('Connect(BranchViewLink)')).to.have.length(2);
   });
 
--- a/test/frontend/components/site/SiteGitHubBranches.test.jsx
+++ b/test/frontend/components/site/SiteGitHubBranches.test.jsx
@@ -40,9 +40,8 @@ describe('<SiteGitHubBranches />', () => {
     expect(rows.at(1).find('td').at(0).text())
       .to.have.string('branch-two');
 
-    // TODO: Nested component selectors, like the `Connect(BranchViewLink)` below, don't
-    // work in this case because the rendered component has  name of `Component`, which enzyme
-    // cant seem to find. Figure out why this is happening
+    // Workaround to select the GitHubIconLink component.
+    // See https://github.com/18F/federalist/issues/1325
     expect(rows.find({
       owner: props.site.owner,
       repository: props.site.repository,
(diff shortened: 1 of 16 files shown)


el-mapache:
I modified the `GitHubRepoLink` component a bit, let me know if y'all would rather see it split up into different components, e.g. an external link component that the github link component could import

jseppi:
I do think a separate component(s) might be a good idea at this point. Could you whip that up?

el-mapache:
yes, definitely!

jseppi:
@el-mapache: you should be able to locally run `yarn lint:diff` to see the same linting errors that CodeClimate will find

el-mapache:
@jseppi I did, but it seemed to only show two errors. Trying again!

el-mapache:
@jseppi All fixed up!

test/frontend/components/site/SiteGitHubBranches.test.jsx
```diff
@@ -29,15 +29,24 @@ describe('<SiteGitHubBranches />', () => {
...
     const rows = wrapper.find('tbody tr');
+
     expect(rows).to.have.length(2);
     expect(rows.at(0).find('td').at(0).text())
       .to.have.string('branch-one');
     expect(rows.at(1).find('td').at(0).text())
       .to.have.string('branch-two');
-    expect(rows.find('GitHubRepoLink')).to.have.length(2);
+
+    // TODO: Nested component selectors, like the `Connect(BranchViewLink)` below, don't
```
jseppi: Let's file an issue for this and reference it here in the comment

el-mapache: sure thing

line47:
maybe "view commit" for the text would be good with `title` text containing the commit value that way we don't have commit numbers breaking the table

el-mapache:
@jseppi I Opened an issue here https://github.com/18F/federalist/issues/1325, take a look and see if it seems reasonable. I already pushed up a commit linking to it in the code.

el-mapache:
@line47 That sounds good to me!

jseppi:
@el-mapache: issue looks good to me, thanks!

el-mapache:
Updated code based on @line47's suggestion!
screen shot 2017-12-04 at 5 25 54 pm

wslack:
Why don't we put the "view commit" link in parenthesis next to the branch name?

Something like this but with a carriage return so the link is always below the branch name:
screen shot 2017-12-04 at 5 33 38 pm

el-mapache:
@wslack Seems fine, the majority of users probably don't need to know what the SHA is, just that they can get back to the commit. @line47 what do you think?

wslack:
The most compact way to do it would be to turn the branch name into a link to the commit but it wouldn't be obvious to users that the links pointed differently.

line47:
@el-mapache yeah I think that sounds OK
@wslack we definitely want to make sure the linking experience is consistent throughout the application UI to avoid confusion.

el-mapache:
@line47 then, should we be using the :octocat: icon for these links as well? The other external github links don't have the links on a separate line, they have a line of text with the :octocat: serving as the link.

wslack:
@line47 agreed but these are different than links to the repo in their behavior and so I think (View Commit) without a GitHub icon is fine.

line47:
I feel the GitHub icon used in conjunction with a link would be most clear approach and avoid confusion on what _is_ and _is not_ a link to GitHub.

**Good example**
[GitHub link](www.federalist.18f.gov) :octocat: 
In this example the link and icon are both inside the same `<a>` tag

**Confusing example**
GitHub link [:octocat:](www.federalist.18f.gov)
In this example the text is not a link and the icon is a link.

Either direction we go let's be consistent and either use GitHub icons in conjunction with a link or don't use Github icons at all and just use link text. I think the icon with text is more clear you will be taken to GitHub.

el-mapache:
@line47 I agree with that, I think that having the icon be the actual link could potentially be confusing to users who don't know much about github, or might not realize that you can click it.

Either the plain link or the 'good' example above seems like a good direction!

line47:
Also wanted to point out this issue which seems to co-inside with this PR. https://github.com/18F/federalist/issues/1266 

@el-mapache if you had some time to pair on making those changes in that PR that would be awesome as I'd like some extra eyes on making those changes.

el-mapache:
@line47 yes, definitely!

el-mapache:
@wslack @jseppi @line47 What do yall think about merging this PR for the functionality? I can put in a separate PR for normalize the link styles (or we can roll it into the PR Scott linked above).

line47:
I vote for merge and work and open a new PR 🎉

wslack:
I would only want us to merge this into staging if we can get the change done before Friday. I don't think I want us to push a new column to the builds table live - instead, this should be a plain link, with no GitHub icon.

Right now, I honestly feel like having a prose link and a Github icon just takes up extra space.

jseppi:
Going to update this branch with `staging` now that #1323 has been merged in. Hopefully it'll kick CodeClimate into getting unstuck.

el-mapache:
Here is the updated 'view commit' link
screen shot 2017-12-06 at 12 49 35 pm

el-mapache:
@line47 @jseppi @wslack commit spam!

wslack:
I'm good with this but someone else should review the code updates

PR #1323, opened 2017-12-04
Dockerize Federalist web app

jseppi:
Sets up this repository to use Docker Compose as the primary development environment.

To Do:
- [x] fix analyze-webpack task/instructions


README.md
```diff
@@ -59,29 +56,32 @@ This will be the file that holds your S3 and SQS configurations.
...
-If you are working on the front-end of the application, the things you need to know are:
+If you are working on the front end of the application, the things you need to know are:
 
-1. It is a React based application
+1. It is a React and Redux application
 1. It is built with `webpack`
 1. It lives in `/frontend`
 
-To analyze the contents of the front end JavaScript bundle, use `yarn analyze-webpack` after a build. This will launch a browser window showing a visualization of all the code that makes up the bundle.
+To analyze the contents of the front end JavaScript bundle, use `docker-compose run --service-ports app yarn analyze-webpack` after a build. Then visit http://127.0.0.1:8888 to see a visualization of the the bundle contents.
```
jseppi: This doesn't work yet. I think it is because I need to change the args to `webpack-bundle-analyzer` to use localhost instead of the 127.0.0.1 IP.

wslack:
What's the 40,000 ft explanation of what we're doing here? (since Docker is a scary word sometimes at GSA). Just making it easier to run the app locally?

jseppi:
Yes, that's correct. This is for running a local development instance of Federalist's web application. It obviates needing a globally installed database and node runtime of particular versions (since they now exist with the Docker Compose environment).

I'm fairly sure I recall the engineering chapter saying that Docker is the recommended way to setup projects for development (but I can't find that guidance anywhere right now :\ ). I hope it's not scary for anyone -- it should be much easier than the alternative. Why do you have the impression of it being scary?

wslack:
Dockerizing an app for local dev is fine. The word "Docker" just has baggage from past issues.

This is what I needed! As long as it changed nothing about the app running on cloud.gov, we're totally fine.

> This is for running a local development instance of Federalist's web application. It obviates needing a globally installed database and node runtime of particular versions (since they now exist with the Docker Compose environment).

jseppi:
Ah, I see. Yup, no changes to the deployed application.

README.md
jseppi: works now -- had to set it to use `0.0.0.0`

config/local.sample.js
```diff
@@ -3,29 +3,31 @@ if (process.env.NODE_ENV !== 'test') {
...
+          clientSecret: 'GITHUB_OAUTH_CLIENT_SECRET',
           callbackURL: 'http://localhost:1337/auth/github/callback',
         },
         organizations: [
-          1234567,
+          1234567, // YOUR GITHUB ORGANIZATION ID
         ],
       },
     },
     postgres: {
```
jmhooper: We can move this out of `local.js` now that these are going to be constant, right?

jseppi: it still needs to be different from the test and production configs. would putting it in `config/postgres.js` do that?

jmhooper: Yeah, it should. The environment configs should still override it there. Just make sure to drop the `postgres` key.

jmhooper:
I'm seeing this error when trying to start it up locally:

```
app_1  | $ NODE_ENV=development concurrently --kill-others --raw "nodemon index.js --config nodemon.json" "webpack --progress --colors --watch"
app_1  | /bin/sh: 1: concurrently: not found
```

jseppi:
@jmhooper: did you run `docker-compose run app yarn && docker-compose run app yarn build` first?

jmhooper:
I did not. Missed that step.

Thoughts on adding that to the Dockerfile so it will be kicked off by `docker-compose build`?

jseppi:
I decided not to do that because 

1) we'd have to change our `postinstall` hook which needs the database present (because it runs the database migration). The database is not present during the Docker image build.
2) any time `package.json` or `yarn.lock` changes, we'd need to run `docker-compose run app yarn` to get the new dependencies anyway (or do a new `docker-compose build`)

jseppi:
@wslack: How do we want to roll this out to the current team? I could go over it at the end of stand-up tomorrow.

wslack:
SGTM!

jseppi:
I'm going to merge this right before stand-up, @wslack

wslack:
@jseppi ok! Seems like folks could use it on the current branch but :man_shrugging:.

jseppi:
Yeah, we can do that also to get folks warmed up without messing with their currently open PR branches.

jseppi:
Going to close and reopen this PR in hopes of getting CodeClimate to restart

jseppi:
Adding a commit made CodeClimate re-run and fix itself.

PR #1326, opened 2017-12-04
Update template selection for multiple screen sizes

line47:
### Small screens (1 up)
screen shot 2017-12-04 at 5 10 19 pm

### Medium screens (2 up)
screen shot 2017-12-04 at 5 10 08 pm

### Large screens (4 up)
screen shot 2017-12-04 at 5 09 55 pm


wslack:
Why doesn't the image in the 1-up scale to full width?

line47:
@wslack that's a conversation I was hoping to address in this PR. 
At smaller screens is a 100% width image optimal use of screen real estate or could we cale it back or not include it at all. 

See the screenshot for a few examples:

localhost_1337_sites_new iphone 6

wslack:
I would say that 100% width is totally ok - looking at it, I'm more concerned about how the "view template" link looks orphaned.

line47:
@wslack Cool, I've updated the images so they run 100% at small screens 👍

PR #1329, opened 2017-12-04
add a space in the text box about the configs

wslack:
Seems like there should be another way to do this that doesn't upset the linter....help?

### DIFF
--- a/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
+++ b/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
@@ -57,7 +57,9 @@ export const AdvancedSiteSettings = ({
         <div className="well">
           <h3 className="well-heading">Site configuration</h3>
           <p className="well-text">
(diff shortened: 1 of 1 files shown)


frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
```diff
@@ -57,8 +57,7 @@ export const AdvancedSiteSettings = ({
         <div className="well">
           <h3 className="well-heading">Site configuration</h3>
           <p className="well-text">
-            Add additional configuration in yaml to be added to your
-            <code>_config.yml</code> file when we build your site&apos;s primary branch.
+            Add additional configuration in yaml to be added to your <code>_config.yml</code> file when we build your site&apos;s primary branch.
```
jseppi: I think this should work:

```jsx
Add additional configuration in yaml to be added to your
{' '}
<code>_config.yml</code> file when we build your site&apos;s primary branch.
```

PR #1339, opened 2017-12-06
(WIP)Autofill github url so user doesnt have to type so much

el-mapache:
This is just a work in progress I put together because I was getting tired of typing the entire github URL every time I added a test site. Let me know if this seems useful and I will write some tests!

auto-fill-form

Alternatively, it might be nice to make the user only type their `username/reponame`, and we can just prepend the github url automatically

### DIFF
--- a/frontend/components/AddSite/AddRepoSiteForm.jsx
+++ b/frontend/components/AddSite/AddRepoSiteForm.jsx
@@ -11,6 +11,7 @@ const propTypes = {
 
   initialValues: PropTypes.shape({
     engine: PropTypes.string.isRequired,
+    repoUrl: PropTypes.string.isRequired
   }).isRequired,
   // the following props are from reduxForm:
   handleSubmit: PropTypes.func.isRequired,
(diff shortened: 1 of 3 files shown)


jseppi:
My thought with requiring the `https://github.com/` at all was that I expected normal users to be copy-pasting from their repository address bar most of the time instead of typing anything. I _think_ this has played out to be true, right, @wslack?

If real users are mostly typing, then I think we'd be better off just asking for `owner/repo` and leaving off the github.com completely. That's kinda what we were doing before I implemented the input in question, though it was spread across two separate inputs.

jseppi:
Even better way to do this: read all repos from the user and their orgs that have allowed Federalist, and provide the user with a list of each repo and a simple button to click to add that repo.

But probably pretty low priority given the effort required since the Add Site flow isn't really that frequently used.

wslack:
Agreed this is too low priority, but it seems easy enough to replace text that I'm 🤷‍♂️ on adding the default text.

el-mapache:
Cool, the context is helpful! This was a pretty trivial amount of work so I'll shelve it if it's only benefitting the devs 😄

el-mapache:
Closing this, see #1366

PR #1340, opened 2017-12-07
Install git in app container

jseppi:
without it, we are not able to run our `yarn lint:diff` command within the `app` container


jseppi:
Folks will have to re-run `docker-compose build` after this is merged

wslack:
@jseppi should this be connected to an issue or is it small enough to not need that?

jseppi:
It was super small and just a dev ergonomics thing.

PR #1344, opened 2017-12-07
Content2

OpenGlobe:
Revisions to the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,11 @@
 [![Test Coverage](https://codeclimate.com/github/18F/federalist/badges/coverage.svg)](https://codeclimate.com/github/18F/federalist/coverage)
 [![Dependency Status](https://gemnasium.com/badges/github.com/18F/federalist.svg)](https://gemnasium.com/github.com/18F/federalist)
 
-***Federalist is updated regularly.  Contact us via [email]() or [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. Or, check our our [documentation](https://federalist-docs.18f.gov/).***
-
-Federalist is a unified interface for publishing static government websites. It automates common tasks for integrating GitHub and Amazon Web Services to provide a simple way for developers to launch new static websites or more easily manage existing static websites. This repo is home to "federalist-core" - a Node.js app that allows government users to add and configure their Federalist sites.
+***Federalist is updated regularly.  Contact us via [email](federalist-support@gsa.gov) or [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. Or, check our our [documentation](https://federalist-docs.18f.gov/).***
 
 ## About Federalist
 
 Federalist helps government agencies publish basic websites quickly and seamlessly.  It provides a simple interface for managing simple website content that is compliant with federal guidelines.  More complex tasks are automated behind the scenes, providing a way for users to more easily create and launch government websites.  Federalist is built on top of Github and Amazon Web Services.   
-
-Federalist-core is a node.js application that allows government users to add and configure basic websites. 
 
 ## Examples
 
@@ -27,7 +23,7 @@ More examples can be found [here] (https://federalist-docs.18f.gov/pages/about-f
 
 ## Getting started
 
-This section will walk you through how to install Federalist locally onto your own machine or within your own environment.  Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day.  If you need help, [contact us](). 
+This section will walk you through how to install Federalist locally onto your own machine or within your own environment.  Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day.  If you need help, [contact us](federalist-support@gsa.gov). 
 
 ### Dependencies / Tooling
 
@@ -76,7 +72,7 @@ This will be the file that holds your S3 and SQS configurations.
       * in the Settings -> Applications -> federalist view in your GitHub Account
 1. Type `cf login --sso -a https://api.fr.cloud.gov -o gsa-18f-federalist -s staging` in terminal.
 1. Visit https://login.fr.cloud.gov/passcode to get a one time passcode.  You'll need to login using your cloud.gov credentials.
-1. Enter your passcode back into terminal.  (Note: if you have a problem at this point it's likely because you don't have permissions.  [Contact us] for help.)
+1. Enter your passcode back into terminal.  (Note: if you have a problem at this point it's likely because you don't have permissions.  [Contact us] (federalist-support@gsa.gov) for help.)
 1. Type `cf apps`.  This will list the current, running applications. 
 1. Type `cf env federalist-staging` to get environment variables.
 1. Find the section in the listing of environment variables that starts with `"s3": [` and look for the following values and paste those values into the S3 section in your `local.js` file.  (Hint: you may have to scroll up to the top of the file to find the S3 environment variables.)
--- a/README.md
+++ b/README.md
@@ -5,25 +5,25 @@
 [![Test Coverage](https://codeclimate.com/github/18F/federalist/badges/coverage.svg)](https://codeclimate.com/github/18F/federalist/coverage)
 [![Dependency Status](https://gemnasium.com/badges/github.com/18F/federalist.svg)](https://gemnasium.com/github.com/18F/federalist)
 
-***Federalist is updated regularly.  Contact us via [email](federalist-support@gsa.gov) or [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. Or, check our our [documentation](https://federalist-docs.18f.gov/).***
+***Federalist is updated regularly. [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. You can also check our our [documentation](https://federalist-docs.18f.gov/) to learn more.***
 
 ## About Federalist
 
-Federalist helps government agencies publish basic websites quickly and seamlessly.  It provides a simple interface for managing simple website content that is compliant with federal guidelines.  More complex tasks are automated behind the scenes, providing a way for users to more easily create and launch government websites.  Federalist is built on top of Github and Amazon Web Services.   
+Federalist helps government agencies publish static websites with fixed content quickly and seamlessly. It provides a user-friendly interface for managing simple website content that is compliant with federal guidelines. More complex tasks are automated behind the scenes, providing a way for users to more easily create and launch government websites. Federalist is built on top of cloud.gov, GitHub, and Amazon Web Services. 
 
 ## Examples
 
-Partner agencies across the federal government use Federalist to host websites.  A few examples include: 
+Partner agencies across the federal government use Federalist to host websites. A few examples include: 
 
 - [College Scorecard] (https://collegescorecard.ed.gov/)
 - [Natural Resources Revenue Data] (https://revenuedata.doi.gov/)
-- [NSF Small Business Innovation Research program] (https://seedfund.nsg.gov)
+- [NSF Small Business Innovation Research program] (https://seedfund.nsf.gov)
 
 More examples can be found [here] (https://federalist-docs.18f.gov/pages/about-federalist/example-sites/). 
 
 ## Getting started
 
-This section will walk you through how to install Federalist locally onto your own machine or within your own environment.  Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day.  If you need help, [contact us](federalist-support@gsa.gov). 
+This section will walk you through how to install Federalist locally onto your own machine or within your own environment. Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day. If you need help, [contact us](federalist-support@gsa.gov). 
 
 ### Dependencies / Tooling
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,27 +5,51 @@
...
+***Federalist is updated regularly.  Contact us via [email](federalist-support@gsa.gov) or [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. Or, check our our [documentation](https://federalist-docs.18f.gov/).***
 
-Federalist is a unified interface for publishing static government websites. It automates common tasks for integrating GitHub and Amazon Web Services to provide a simple way for developers to launch new static websites or more easily manage existing static websites. This repo is home to "federalist-core" - a Node.js app that allows government users to add and configure their Federalist sites.
+## About Federalist
+
+Federalist helps government agencies publish basic websites quickly and seamlessly.  It provides a simple interface for managing simple website content that is compliant with federal guidelines.  More complex tasks are automated behind the scenes, providing a way for users to more easily create and launch government websites.  Federalist is built on top of Github and Amazon Web Services.
```
jseppi: I'm not sure the terms `basic` or `simple` are necessary or wanted. Is the issue that `static` is unclear? It is kind of a nuanced technical term, but I'm not sure of a better word for it.

jseppi: `Github` -> `GitHub` (that's how the company stylizes it)

jseppi: I wonder if we should also mention cloud.gov? 

`Federalist is built on top of cloud.gov, Amazon Web Services, and GitHub.`

README.md
```diff
@@ -5,27 +5,51 @@
...
+
+- [College Scorecard] (https://collegescorecard.ed.gov/)
+- [Natural Resources Revenue Data] (https://revenuedata.doi.gov/)
+- [NSF Small Business Innovation Research program] (https://seedfund.nsg.gov)
+
+More examples can be found [here] (https://federalist-docs.18f.gov/pages/about-federalist/example-sites/). 
 
 ## Getting started
 
+This section will walk you through how to install Federalist locally onto your own machine or within your own environment.  Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day.  If you need help, [contact us](federalist-support@gsa.gov).
```
jseppi: I'm not totally sure we want to offer external support for development setup.

README.md
```diff
@@ -5,27 +5,51 @@
...
+This section will walk you through how to install Federalist locally onto your own machine or within your own environment.  Depending on your familiarity with the dependencies and process, this may take anywhere between an hour or two to the better part of a day.  If you need help, [contact us](federalist-support@gsa.gov). 
+
 ### Dependencies / Tooling
 
 Before you start, ensure you have the following installed:
 
 - [Cloud Foundry CLI](https://docs.cloudfoundry.org/cf-cli/install-go-cli.html)
 - [Docker Compose](https://docs.docker.com/compose/install/#install-compose)
 
+Most of these dependencies can be installed relatively quickly if you don't already have them.  Most can also be installed via terminal using popular package managers like [brew](https://brew.sh/).
```
jseppi: I think this is out of date since now there are only two requirements. The word `most` feels a little too "big." I think we could just remove these sentences.

README.md
```diff
@@ -5,27 +5,51 @@
...
-Federalist uses Docker Compose for local development.
+### Setting up a local development environment on your own machine
 
+This section will walk you through downloading the Federalist code, making sure you are running the correct version of node, and creating a copy of the configuration file you'll edit later. Federalist uses Docker Compose for local development.
+
+1. To get started, you'll need to first create your own copy of the Federalist code.  Download or clone the `18F/federalist` repository from Github using command line in terminal or by clicking on the *clone or download* button above. 
 1. Clone the `18F/federalist` repository from Github and `cd` to that directory.
+
+#### Editing the configuration file 
+In this section we'll walk you through the steps necessary to get all the varialbes you'll need to put in the `local.js` configuration file you've just created.
```
jseppi: `varialbes` -> `variables`

README.md
```diff
@@ -5,27 +5,51 @@
...
+#### Editing the configuration file 
+In this section we'll walk you through the steps necessary to get all the varialbes you'll need to put in the `local.js` configuration file you've just created.  
+
 1. Make a copy of `config/local.sample.js` and name it `local.js` and place it in the `config` folder. You can do this by running `cp config/local{.sample,}.js`
 This will be the file that holds your S3 and SQS configurations.
-1. [Register a new OAuth application on GitHub](https://github.com/settings/applications/new). You'll want to use `http://localhost:1337/auth` as the "Authorization callback url". Once you have created the application, you'll see a `Client ID` and `Client Secret`. Add these values to `config/local.js`.
+
+1. [Register a new OAuth application on GitHub](https://github.com/settings/applications/new). You'll want to use `http://localhost:1337/auth` as the "Authorization callback url". Once you have created the application, you'll see a `Client ID` and `Client Secret`on the next screen. Add these values to `config/local.js`.  Next, add 'http://localhost:1337/auth/github/callback' as the value for the 'callbackURL'.   Your file should look something like this:
```
jseppi: too many spaces before `Your file`

README.md
```diff
@@ -47,21 +71,30 @@ This will be the file that holds your S3 and SQS configurations.
       * a first-time login with your GitHub credentials
       * in the Settings -> Applications -> federalist view in your GitHub Account
 1. Type `cf login --sso -a https://api.fr.cloud.gov -o gsa-18f-federalist -s staging` in terminal.
-1. Visit https://login.fr.cloud.gov/passcode to get a one time passcode.
-1. Enter your passcode back into terminal.
-1. Type `cf apps`.
+1. Visit https://login.fr.cloud.gov/passcode to get a one time passcode.  You'll need to login using your cloud.gov credentials.
+1. Enter your passcode back into terminal.  (Note: if you have a problem at this point it's likely because you don't have permissions.  [Contact us] (federalist-support@gsa.gov) for help.)
```
jseppi: Again, not sure we want to offer email support like this. I'd suggest that these instructions are targeted at 18F/TTS folks who are working on Federalist (so maybe that should be started at the start of this local development section).

README.md
```diff
@@ -47,21 +71,30 @@ This will be the file that holds your S3 and SQS configurations.
...
+1. Type `cf apps`.  This will list the current, running applications. 
 1. Type `cf env federalist-staging` to get environment variables.
-1. Find the section in the listing of environment variables that starts with `"s3": [` and look for the following values and paste those values into the S3 section in your `local.js` file.
+1. Find the section in the listing of environment variables that starts with `"s3": [` and look for the following values and paste those values into the S3 section in your `local.js` file.  (Hint: you may have to scroll up to the top of the file to find the S3 environment variables.)
```
jseppi: `top of the file` -> `top of your terminal window`

README.md
```diff
@@ -47,21 +71,30 @@ This will be the file that holds your S3 and SQS configurations.
       * a first-time login with your GitHub credentials
       * in the Settings -> Applications -> federalist view in your GitHub Account
 1. Type `cf login --sso -a https://api.fr.cloud.gov -o gsa-18f-federalist -s staging` in terminal.
-1. Visit https://login.fr.cloud.gov/passcode to get a one time passcode.
-1. Enter your passcode back into terminal.
-1. Type `cf apps`.
+1. Visit https://login.fr.cloud.gov/passcode to get a one time passcode.  You'll need to login using your cloud.gov credentials.
+1. Enter your passcode back into terminal.  (Note: if you have a problem at this point it's likely because you don't have permissions.  [Contact us] (federalist-support@gsa.gov) for help.)
+1. Type `cf apps`.  This will list the current, running applications.
```
jseppi: This isn't really necessary, just perhaps helpful. Maybe strike it.

README.md
jseppi: Something else to maybe note: the S3 credentials are changed every time a commit to `staging` happens, so folks will need to repeat this and the preceding step with some frequency.

README.md
```diff
@@ -5,27 +5,51 @@
...
 
 - [Cloud Foundry CLI](https://docs.cloudfoundry.org/cf-cli/install-go-cli.html)
 - [Docker Compose](https://docs.docker.com/compose/install/#install-compose)
 
+Most of these dependencies can be installed relatively quickly if you don't already have them.  Most can also be installed via terminal using popular package managers like [brew](https://brew.sh/). 
+
 ### Running locally (development)
 
-Federalist uses Docker Compose for local development.
+### Setting up a local development environment on your own machine
```
jseppi: This heading is the same level `###` as the preceding one. Maybe delete the preceding one.

Is `on your own machine` needed? I think `local` implies that.

README.md
```diff
@@ -47,21 +71,30 @@ This will be the file that holds your S3 and SQS configurations.
...
+1. Next, we'll enter some environment variables for SQS.  Under credentials (listed just under the S3 section of the document), find the following values and paste those values into the SQS section in your `local.js` file.
     - `FEDERALIST_AWS_BUILD_KEY` is `accessKeyId`
     - `FEDERALIST_SESSION_SECRET` is `secretAccessKey`
     - `FEDERALIST_SQS_QUEUE` is `queue`
+You're done!  You've now successfully set all the environment variables for your new Federalist development environment.  Next, we'll set up docker to complete the process.  
+
+#### Setting up Docker   
+     
 1. Run `docker-compose build`.
 1. Run `docker-compose run app yarn && docker-compose run app yarn build` to install dependencies and build the app initially.
```
jseppi: Maybe add a note that any time the node dependencies are changed (like from a recently complete new feature), then `docker-compose run app yarn` will need to be re-run after pulling the new code from GitHub.

README.md
```diff
@@ -47,21 +71,30 @@ This will be the file that holds your S3 and SQS configurations.
...
-1. Run `docker-compose up` to start the development environment.  You should now be able to see Federalist running at http://localhost:1337/. Local file changes will cause the server to restart and/or the front end bundles to be rebuilt.
+1. Run `docker-compose up` to start the development environment.  
+
+### Check to see if everything is working correctly
```
jseppi: Should this be a level 4 heading (`####`) instead?

README.md
OpenGlobe: My thought here is that not all users may understand what 'static' means, as not all users will necessarily have a technical background.   Because we'd (I think) want to appeal to the broadest audience possible, more generic terms might be useful.  We could however certainly figure out another place to delineate static.  

👍 to Github styling, and mentioning cloud.gov however!

README.md
OpenGlobe: I would find this frustrating if I didn't know where to go other than the docs for help, and I was a prospective customer.  But I understand your point.  Do we have a sense of how often this might be likely to occur - is it worth testing out?

README.md
OpenGlobe: Super.  We can edit.

README.md
OpenGlobe: I think 'local' implies it to a person familiar with development, but perhaps not someone less familiar.  But I think ties back to some broader content strategy items, it's a good observation.

README.md
jseppi: Hrm, I don't think a prospective customer would be setting up a development instance of the Federalist application though. It's really only for 18F developers, and maybe security researchers or extremely curious open source aficionados.

README.md
OpenGlobe: 👍

README.md
OpenGlobe: 👍

README.md
OpenGlobe: I included this because I wondered if it might be a bit frustrating for someone to get all the way to this point, and then not have cloud.gov credentials.  Perhaps we should add it as a requirement before they start above?  The permissions were an issue I ran into - if this is unusual, we could strike this.

README.md
OpenGlobe: Ooof - that's good to know!

README.md
OpenGlobe: 👍

README.md
jseppi: I meant the whole step of running `cf apps`.

I agree adding some info about needing access to our cloud.gov space before getting started. That probably ties in with the whole audience thing.

Maybe something at the start like:

>This section is primarily for 18F/TTS developers working on the Federalist project. Before you get started, make sure you have been fully on-boarded, including getting access to the Federalist cloud.gov space.

README.md
wslack: This is a README and not intended as a sales document. It seems like we shouldn't sacrifice any technical detail here IMHO - which user are we writing for?

README.md
wslack: Agreed with @jseppi - Federalist is not made to be run by customers. There's no functionality used by customers that involves running locally.

README.md
wslack: I am comfortable for the README to be aimed at a developer; standard users shouldn't refer to it.

README.md
wslack: Agreed; we shouldn't put instructions for 18F/TTS in here. Everything should apply to someone who wants to run their own indepedent Federalist instance.

README.md
```diff
@@ -5,27 +5,51 @@
 [![Test Coverage](https://codeclimate.com/github/18F/federalist/badges/coverage.svg)](https://codeclimate.com/github/18F/federalist/coverage)
 [![Dependency Status](https://gemnasium.com/badges/github.com/18F/federalist.svg)](https://gemnasium.com/github.com/18F/federalist)
 
-***Under active development. Everything is subject to change. Learn more at the [documentation site](https://federalist-docs.18f.gov/). Interested in talking to us? [Join our public chat room](https://chat.18f.gov/).***
+***Federalist is updated regularly.  Contact us via [email](federalist-support@gsa.gov) or [Join our public chat room](https://chat.18f.gov/) to talk to us and stay informed. Or, check our our [documentation](https://federalist-docs.18f.gov/).***
```
wslack: I don't want to put the support e-mail out in a public README - it's for partners only.

README.md
```diff
@@ -5,27 +5,51 @@
...
+
+Federalist helps government agencies publish basic websites quickly and seamlessly.  It provides a simple interface for managing simple website content that is compliant with federal guidelines.  More complex tasks are automated behind the scenes, providing a way for users to more easily create and launch government websites.  Federalist is built on top of Github and Amazon Web Services.   
+
+## Examples
+
+Partner agencies across the federal government use Federalist to host websites.  A few examples include: 
+
+- [College Scorecard] (https://collegescorecard.ed.gov/)
+- [Natural Resources Revenue Data] (https://revenuedata.doi.gov/)
+- [NSF Small Business Innovation Research program] (https://seedfund.nsg.gov)
```
wslack: NSG should be "NSF" in that link.

README.md
wslack: We should put content like this in the TTS handbook.

README.md
jseppi: Unfortunately because of the tight coupling to the cloud.gov/AWS services, 18F/TTS folks working on Federalist will need these steps. I think we should prioritize their use of these instructions since we really want to make getting a local instance setup for a newly onboarded person as easy as possible.

Maybe we could just prefix these cloud.gov steps with `**For 18F/TTS developers:**`?

wslack: That sounds good to me.

README.md
annahsebok: Editing this a bit to remove the support email.

README.md
annahsebok: Based on @wslack's most recent comment and my understanding of your users, I think static works here, especially if you define what static means. I'm also making a few copy edits to enhance the language (for example, not using the word simple twice in one sentence). Also adding cloud.gov and editing GitHub style.

README.md
annahsebok: editing

README.md
annahsebok: @wslack Are you suggesting that this content move to the TTS handbook **and** this section of the Federalist README should point people there? Also, it sounds like y'all are contemplating potentially moving this entire section of instructions. Is that from **Getting started** all the way through **Testing and linting**?

README.md
annahsebok: removing

README.md
annahsebok: Removing **Running locally (development)** header as well as "on your machine" per the comments here.

README.md
annahsebok: editing

README.md
annahsebok: editing

README.md
annahsebok: editing **For 18F/TTS developers:** in.

README.md
annahsebok: I'm editing in the copy @jseppi wrote right after the next heading for this section (per the comment thread above this).

README.md
annahsebok: editing

README.md
annahsebok: editing

README.md
annahsebok: Yup! Editing.

annahsebok:
I just submitted a new commit to content2 that addresses a number of the comments and edits identified in this PR. There are a few comments still outstanding for clarification.

jseppi:
Closing in favor of #1582

PR #1347, opened 2017-12-07
Adds endpoint and handler to remove a user from a  site

el-mapache:


api/routers/site.js
```diff
@@ -9,6 +9,7 @@ router.use(csrfProtection);
 
 router.get('/site', sessionAuth, SiteController.findAllForUser);
 router.post('/site/user', sessionAuth, SiteController.addUser);
+router.delete('/site/:site_id/user/:user_id', sessionAuth, SiteController.removeUser);
```
el-mapache: I think I'd like to break this up into a separate router/controller pair, along with the POST method above. I think that route could also be normalized to `/site/:site_id/user`.

That would need to be a separate refactoring PR, since parts of the front end would have to change as well.

api/controllers/site.js
```diff
@@ -95,6 +99,37 @@ module.exports = {
...
+      }
+
+      if (site.Users.length === 1) {
+        throw {
+          status: 400,
+          message: 'A site must have at least one user',
+        };
+      }
+
+      return authorizer.destroy(req.user, site).then(() => site);
```
jmhooper: This authorizer is for checking whether a user can delete a site. We may want to add a new action to the authorizer for removing users.

el-mapache: I think the only stipulation to a user being able to delete a site is that they have access to the site (and have write permissions on GitHub). Would the new authorizer action just check for both of those?  I guess I'm saying that the check the authorizer is performing still applies in this case.

jmhooper: Yeah, even if the new action just calls the old action it works. I think what we'd want to avoid is a situation where we change the logic we use to check whether a user can delete a site, and inadvertently change the logic we use to check whether a user can delete another user from a site. Adding a new action to the authorizer makes it explicit that these are 2 distinct activities.

el-mapache: Ok, since the authorizer also overloads the `create` action for adding a new user, we should create a separate method to handle that as well, correct?

jmhooper: Yeah, I think so

wslack:
Good work on this.

PR #1351, opened 2017-12-08
Frontend to remove users from a site

el-mapache:
* Adds link to remove user from a site via the `Users` section in site detail view

### DIFF
--- a/api/responses/siteErrors.js
+++ b/api/responses/siteErrors.js
@@ -2,4 +2,5 @@ module.exports = {
   USER_REQUIRED: 'A site must have at least one user. If you want to remove the last user, delete the site from Settings -> Advanced.',
   OWNER_REMOVE: 'You cannot remove the owner of a site.',
   WRITE_ACCESS_REQUIRED: 'You do not have write access to this repository',
+  NO_ASSOCIATED_USER: 'That user is not associated with this site.',
 };
--- a/api/responses/siteErrors.js
+++ b/api/responses/siteErrors.js
@@ -1,0 +1,6 @@
+module.exports = {
+  USER_REQUIRED: 'A site must have at least one user. If you want to remove the last user, delete the site from Settings -> Advanced.',
+  OWNER_REMOVED_SELF: 'You cannot remove yourself from a site that you own.',
+  OWNER_REMOVE: 'You cannot remove the owner of a site.',
+  WRITE_ACCESS_REQUIRED: 'You do not have write access to this repository',
+};
--- a/api/responses/siteErrors.js
+++ b/api/responses/siteErrors.js
@@ -1,6 +1,5 @@
 module.exports = {
   USER_REQUIRED: 'A site must have at least one user. If you want to remove the last user, delete the site from Settings -> Advanced.',
-  OWNER_REMOVED_SELF: 'You cannot remove yourself from a site that you own.',
   OWNER_REMOVE: 'You cannot remove the owner of a site.',
   WRITE_ACCESS_REQUIRED: 'You do not have write access to this repository',
 };
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -66,6 +66,13 @@ const revokeSiteMembership = ({ user, site, userId }) =>
     .then(() => {
       const userToRemove = site.Users.find(u => u.id === Number(userId));
 
+      if (!userToRemove) {
+        throw {
+          message: siteErrors.NO_ASSOCIATED_USER,
+          status: 404,
+        };
+      }
+
       if (userToRemove.username.toLowerCase() === site.owner.toLowerCase()) {
         throw {
           message: siteErrors.OWNER_REMOVE,
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -53,9 +53,9 @@ const createSiteMembership = ({ user, siteParams }) => {
 };
 
 const revokeSiteMembership = ({ user, site, userId }) => {
-  if (user.username.toLowerCase() === site.owner.toLowerCase()) {
+  if (userId === user.id && user.username.toLowerCase() === site.owner.toLowerCase()) {
     throw {
-      message: 'You cannot remove yourself from a site that you are the owner of.',
+      message: 'You cannot remove yourself from a site that you own.',
       status: 400,
     };
   }
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -53,15 +53,8 @@ const createSiteMembership = ({ user, siteParams }) => {
   ).then(() => site);
 };
 
-const revokeSiteMembership = ({ user, site, userId }) => {
-  if (userId === user.id && user.username.toLowerCase() === site.owner.toLowerCase()) {
-    throw {
-      message: siteErrors.OWNER_REMOVED_SELF,
-      status: 400,
-    };
-  }
-
-  return GitHub.checkPermissions(user, site.owner, site.repository)
+const revokeSiteMembership = ({ user, site, userId }) =>
+  GitHub.checkPermissions(user, site.owner, site.repository)
     .then((permissions) => {
       if (!permissions.push) {
         throw {
@@ -82,6 +75,5 @@ const revokeSiteMembership = ({ user, site, userId }) => {
 
       site.removeUser(userToRemove);
     });
-};
 
 module.exports = { createSiteMembership, revokeSiteMembership };
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -80,7 +80,7 @@ const revokeSiteMembership = ({ user, site, userId }) =>
         };
       }
 
-      site.removeUser(userToRemove);
+      return site.removeUser(userToRemove);
     });
 
 module.exports = { createSiteMembership, revokeSiteMembership };
--- a/api/services/SiteMembershipCreator.js
+++ b/api/services/SiteMembershipCreator.js
@@ -1,5 +1,6 @@
 const GitHub = require('./GitHub');
 const { Site, User } = require('../models');
+const siteErrors = require('../responses/siteErrors');
 
 const checkGithubRepository = ({ user, owner, repository }) =>
   GitHub.getRepository(user, owner, repository).then((repo) => {
@@ -11,7 +12,7 @@ const checkGithubRepository = ({ user, owner, repository }) =>
     }
     if (!repo.permissions.push) {
       throw {
-        message: 'You do not have write access to this repository',
+        message: siteErrors.WRITE_ACCESS_REQUIRED,
         status: 400,
       };
     }
@@ -55,7 +56,7 @@ const createSiteMembership = ({ user, siteParams }) => {
 const revokeSiteMembership = ({ user, site, userId }) => {
   if (userId === user.id && user.username.toLowerCase() === site.owner.toLowerCase()) {
     throw {
-      message: 'You cannot remove yourself from a site that you own.',
+      message: siteErrors.OWNER_REMOVED_SELF,
       status: 400,
     };
   }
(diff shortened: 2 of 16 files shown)


frontend/components/site/SiteUsers.jsx
```diff
@@ -44,6 +52,9 @@ const SiteUsers = ({ site, user }) => {
                 {' '}
                 {u.username.toLowerCase() === user.username.toLowerCase() ? '(you)' : ''}
               </td>
+              <td>
+                <ButtonLink clickHandler={handleClick(user)}>Remove user</ButtonLink>
```
jseppi: BUG: this needs to be `handleClick(u)`. `u` is the user object in the `.map` callback; `user` is the current Federalist use who is signed in. Maybe we should rename `u` to `rowUser` or something more descriptive, which might help prevent similar issues in the future.

jseppi: Also, I wonder if we should hide the button and disallow the removal on the backend for the owner of the repository. It's a little weird to delete yourself from your own site.

el-mapache: Ah you're right, I had it as `u`, then moved some stuff around, then forgot to change it back...probably means something needs to be tested

jseppi: Oh, maybe you could check the arguments to the `spy` to make sure it is the expected user being passed to the handler

test/frontend/components/site/SiteUsers.test.jsx
```diff
@@ -16,59 +19,72 @@ describe('<SiteUsers/>', () => {
...
+
+    it('should render a `Remove user` link in each action column', () => {
+      const clickSpy = stub(siteActions, 'removeUserFromSite');
+      wrapper = shallow(<SiteUsers {...props} />);
+      const rows = wrapper.find('tbody tr');
+
+      rows.forEach((row) => {
+        const removeUserLink = row.find('td').last().find('ButtonLink').shallow();
+
+        removeUserLink.simulate('click', { preventDefault: () => ({}) });
```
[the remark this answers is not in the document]
el-mapache: I agree, but code climate complained that the method wasn't checked

frontend/components/site/SiteUsers.jsx
el-mapache: You are probably right, so the only time the error in the comment above gets displayed is when the user:

* is the sole user registered in federalist
* is not the original owner of the repo

jseppi: Yeah. I think we need to keep that error for organization-owned repositories (like `18f/federalist-docs`). But for personal account-owned repos (like `jseppi/test-federalist-site`) we probably ever want the site to get to a state where the original account does not have access (unless they've totally deleted it from Federalist via the Advanced Settings).

el-mapache:
@jseppi All updated!

jseppi:
hrm, when I try to remove a different user I get an error that I cannot remove myself:

remove-user

api/services/SiteMembershipCreator.js
```diff
@@ -52,8 +52,15 @@ const createSiteMembership = ({ user, siteParams }) => {
   ).then(() => site);
 };
 
-const revokeSiteMembership = ({ user, site, userId }) =>
-  GitHub.checkPermissions(user, site.owner, site.repository)
+const revokeSiteMembership = ({ user, site, userId }) => {
+  if (user.username.toLowerCase() === site.owner.toLowerCase()) {
```
jseppi: I think this check is incorrect. It is essentially blocking the repo owner from revoking access of any other user.

I think it needs to be something like

```js
if (userId == user.id && user.username.toLowerCase() === site.owner.toLowerCase()) {
   // ...
}
```

el-mapache: hmm ok, I'll have to check the tests again then, they all pass but this behavior is obviously not correct.

test/api/requests/site.test.js
```diff
@@ -684,7 +684,44 @@ describe('Site API', () => {
...
       ).then((response) => {
         validateAgainstJSONSchema('DELETE', path, 400, response.body);
-        expect(response.body.message).to.equal('A site must have at least one user');
+        expect(response.body.message).to.equal('A site must have at least one user. If you want to remove the last user, delete the site from Settings -> Advanced.');
+        done();
+      })
+      .catch(done);
+    });
+
+    it('should respond with a 400 when the site owner attempts to delete the site', (done) => {
```
jseppi: this wording seems a little off. Maybe should be `should respond with a 400 when the site owner attempts to remove themself`

test/api/requests/site.test.js
```diff
@@ -684,7 +684,44 @@ describe('Site API', () => {
...
+      .then(({ user, site, cookie }) => {
+        githubAPINocks.repo({
+          owner: user.username,
+          repository: site.repo,
+          response: [200, {
+            permissions: { admin: false, push: false },
+          }],
+        });
+
+        return request(app).delete(`/v0/site/${site.id}/user/${user.id}`)
```
jseppi: Maybe we need another test very similar to this one but that calls delete for `anotherUser`

```js
request(app).delete(`/v0/site/${site.id}/user/${anotherUser.id}`)
```

el-mapache: That's the test I need to check, I wrote one for that behavior on line 645.

test/api/requests/site.test.js
el-mapache: You're right, I'm not sure what I was thinking here :D

test/api/requests/site.test.js
el-mapache: I think you are right however, we should have an additional test in which the owner tries to delete a non-owner user.

el-mapache:
@jseppi ok! I _think_ this is all good now.

api/services/SiteMembershipCreator.js
```diff
@@ -52,8 +52,15 @@ const createSiteMembership = ({ user, siteParams }) => {
   ).then(() => site);
 };
 
-const revokeSiteMembership = ({ user, site, userId }) =>
-  GitHub.checkPermissions(user, site.owner, site.repository)
+const revokeSiteMembership = ({ user, site, userId }) => {
+  if (userId === user.id && user.username.toLowerCase() === site.owner.toLowerCase()) {
```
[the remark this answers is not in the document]
jseppi: I think to prevent a user from removing the site owner you'd first need grab the user model associated with `userId` and then check that user's username field against the `site.owner`.

api/services/SiteMembershipCreator.js
```diff
@@ -66,5 +73,6 @@ const revokeSiteMembership = ({ user, site, userId }) =>
       const userToRemove = site.Users.find(u => u.id === Number(userId));
```
jseppi: here's I think where you could add a check to make sure  the owner isn't getting removed:

```js
if (userToRemove.username.toLowerCase() === site.owner.toLowerCase()) {
 throw { status: 400, message: 'The site owner cannot be removed' };
}
```

frontend/components/site/SiteUsers.jsx
```diff
@@ -27,22 +38,31 @@ const SiteUsers = ({ site, user }) => {
...
                 </a>
                 {' '}
-                {u.username.toLowerCase() === user.username.toLowerCase() ? '(you)' : ''}
+                {isSiteOwner(user, rowUser) ? '(you)' : ''}
+              </td>
+              <td>
+                {
+                  isSiteOwner(user, rowUser) ? '-' :
+                  <ButtonLink clickHandler={handleClick(rowUser)}>
+                    Remove user
```
jseppi: I've noticed that sometimes (but not always) after clicking this link that the table doesn't always re-rerender without the removed user. Have you noticed that? I haven't seen anything obvious that cause that behavior, but it definitely seems like something is up o_O

jseppi: Maybe something with the `SITE_USER_REMOVED` case in the reducer?

el-mapache: I haven't noticed this, does anything look obviously wrong in the test in `test/frontend/reducers/sitesTest.js` (starting on line 231)?

api/services/SiteMembershipCreator.js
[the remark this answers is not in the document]
el-mapache: Yep, I was planning on adding that in one of the follow up PRs since that criteria wasn't really established until yesterday afternoon, I'll add it now!

frontend/components/site/SiteUsers.jsx
el-mapache: Is there a specific use case that triggers this, or does it seem to randomly happen?

el-mapache:
@jseppi I added code and a check for _any_ user trying to remove an owner. Is it still worthwhile to have a separate check for the owner removing themselves? Like, in the event that an owner somehow interacted with the API on the front end, do we need a special message just for them?

I'm also unable to repro the table re-rendering error you mentioned :( I tried adding multiple users ~ half a dozen time and deleting them, everything seemed to render ok.

frontend/components/site/SiteUsers.jsx
jseppi: Seems to be random. 🤷‍♀️ 
The test looks correct to me.

api/responses/siteErrors.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
jseppi: 👍 nice!

api/services/SiteMembershipCreator.js
```diff
@@ -64,13 +65,21 @@ const revokeSiteMembership = ({ user, site, userId }) => {
...
-          message: 'You do not have write access to this repository',
+          message: siteErrors.WRITE_ACCESS_REQUIRED,
           status: 400,
         };
       }
     })
     .then(() => {
       const userToRemove = site.Users.find(u => u.id === Number(userId));
+
+      if (userToRemove.username.toLowerCase() === site.owner.toLowerCase()) {
```
jseppi: 👍

test/api/requests/site.test.js
```diff
@@ -566,10 +567,11 @@ describe('Site API', () => {
 
   describe('DELETE /v0/site/:site_id/user/:user_id', () => {
     const path = '/site/{site_id}/user/{user_id}';
+    const requestPath = (siteId, userId) => `/v0/site/${siteId}/user/${userId}`;
```
jseppi: nice

jseppi:
Yeah, good point: we don't need a special error for the owner removing themself. That would be a highly weird interaction since the front end blocks it.

el-mapache:
@jseppi Ok, I've removed the extra check, now any user attempting to remove an owner will trigger the same error message.

jseppi:
huh, the no-refresh is definitely still happening.

I had to try it out three times before it happened, but it did. I recorded:

remove-user-no-refresh

el-mapache:
@jseppi can you trigger the rebuild please

wslack:
Wait a sec - does this block people from removing themselves? (it should be possible to remove yourself as long as you aren't the only person left)

el-mapache:
@wslack Its blocks people who are owners of the site from removing themselves, yes.

el-mapache:
@jseppi Is there an error in the console? It doesn't look like anything happened. I think that if react was re-rendering the table, the focus on the link would have been lost.

jseppi:
No error. And the logged states look correct. I really don't get why it is happening. The database change is made, and when I refresh the browser, the table reflects that the user was removed.

el-mapache:
Maybe there is actually a race condition with data getting fetched, does it look like any other actions are firing off? Obviously I can see that this is happening but I cant repro it at all

el-mapache:
Or, do you have identical entries in the join table?

api/services/SiteMembershipCreator.js
```diff
@@ -57,13 +58,28 @@ const revokeSiteMembership = ({ user, site, userId }) =>
...
+      }
+
+      if (userToRemove.username.toLowerCase() === site.owner.toLowerCase()) {
+        throw {
+          message: siteErrors.OWNER_REMOVE,
+          status: 400,
+        };
+      }
+
       site.removeUser(userToRemove);
```
jseppi: As we discussed while pairing, we're pretty sure there needs to be a `return` in front of this statement because it is probably also async. That is, the issue turned out to be a race condition of this removal not completing before the site was refetched and returned to the front end.

el-mapache:
the `site` object is out of scope by the time we get to the end of the promise chain in the `removeUser` handler, so I just passed in the `id` in an object to the serializer

Issue #1307, opened 2017-11-22
Live and demo URL input focus is lost after first character is typed


jseppi:
Steps to reproduce:
1. Click into a blank Live (or Demo) URL input box on the Site Settings page.
1. Type a character. The focus will jump to the Save Basic Settings button.

Expected behavior:
Focus remains in the input box while typing.

![url-focus-lost](https://user-images.githubusercontent.com/697848/33139230-476b1be8-cf72-11e7-8ec9-3c58c2b274f4.gif)

jseppi:
It is due to code like this, which results in React re-rendering the component on change, which then makes the input lose focus (because it's actually a new DOM element):

```jsx
component={p =>
  <HttpsUrlInput
    value={p.input.value}
    onChange={p.input.onChange}
    className="form-control"
    id="domainInput"
  />
}
```

This method was taken from the ReduxForm docs, but those examples are apparently wrong. More explanation at https://github.com/erikras/redux-form/issues/1094

wslack:
Effort to fix?

jseppi:
Fairly low effort. I’m thinking I could roll it in with fixing the weird
error message when anything in the related form’s submission goes wrong.


On Nov 22, 2017, at 12:41 PM, Will Slack <notifications@github.com> wrote:

Effort to fix?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/18F/federalist/issues/1307#issuecomment-346439294>, or mute
the thread

wslack:
Makes sense, seems like we can put in "Ready" for you to pull off when wanted.

PR #1369, opened 2017-12-13
Use ReduxForm Fields for domain inputs in Basic Settings.

jseppi:
This fixes the bug with losing focus after the first character is typed in the domain inputs on the Basic Settings page (#1307).

I did so by turning those inputs into custom ReduxForm Fields (code example at https://redux-form.com/7.2.0/examples/fieldlevelvalidation/), similar to the existing `GitHubRepoUrlField` component. Since the new `HttpsUrlField` and the existing `GitHubRepoUrlField` are very similar, they now share a common `RenderUrlField` component.

Validation errors now look like this (instead of being HTML5 validation error tooltips):

screen shot 2017-12-13 at 12 01 15 pm


frontend/components/Fields/HttpsUrlField.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Field } from 'redux-form';
+import { isURL } from 'validator';
+
+import RenderUrlField from './RenderUrlField';
+
+
+export const isHttpsUrlWithoutPath = (value) => {
+  if (value && value.length) {
+    if (!isURL(value, { protocols: ['https'] }) || value.lastIndexOf('/') !== 7) {
```
el-mapache: Does this method report a valid URL if the two trailing `/` after the `:` are missing?

jseppi: Nope, it reports that is invalid. Just added a couple additional test expectations to show that in 69/commits/4811c

el-mapache: if that's the case, do we need the `||` condition here?

jseppi: that's to make sure that there is no additional path in the url. `7` is the position of the last `/` in `https://`. Any additional would indicate that there is a sub-path, which we don't want.

ie, `https://example.gov` is valid, but `https://example.gov/` and `https://example.gov/whatever` are not valid.

el-mapache: I didnt see the withoutPath part of the method name, makes sense.

PR #1370, opened 2017-12-13
add explanatory text

wslack:


frontend/components/site/SiteSettings/index.jsx
```diff
@@ -59,6 +59,11 @@ class SiteSettings extends React.Component {
 
     return (
       <div>
+        <p>
+          See our doucmentation site for more about
+          <a href="https://federalist-docs.18f.gov/pages/using-federalist/#managing-site-settings">these settings</a> or
```
el-mapache: Do we want these to open in another tab? Most of our external links on federalist currently do so. I think we should also provide a `title` attribute for these.

wslack: Good point

wslack: I should also have target="_blank" rel="noopener noreferrer" in the link right?

el-mapache: yes! You'll probably have to put the properties on a separate line to stay within our linter's line length requirements

jseppi: yes

wslack:
Oh thanks @el-mapache - was on my list to do.

wslack:
Now I just need a review!

frontend/components/site/SiteSettings/index.jsx
```diff
@@ -59,6 +59,11 @@ class SiteSettings extends React.Component {
...
```
jseppi: did you still want these links to open in a new window?

wslack:
Oh right, I have more to do here. One moment.

Issue #1368, opened 2017-12-13
Invalid or non-unique key prop in TemplateList on Add Site page


jseppi:
<img width="1126" alt="screen shot 2017-12-13 at 10 55 40 am" src="https://user-images.githubusercontent.com/697848/33951303-3e51c54c-dff4-11e7-84cc-4f230cbd49b0.png">

PR #1371, opened 2017-12-13
Specify keys in TemplateSite component arrays

jseppi:
ref #1368

### DIFF
--- a/frontend/components/AddSite/TemplateSiteList/index.js
+++ b/frontend/components/AddSite/TemplateSiteList/index.js
@@ -63,11 +63,12 @@ export class TemplateList extends React.Component {
     let index = 0;
 
     const templateGrid = templateRows.map(row => (
-      <div className={`flexbox-grid flexbox-grid-${MAX_CELLS_PER_ROW}`} key={row}>
+      <div className={`flexbox-grid flexbox-grid-${MAX_CELLS_PER_ROW}`} key={row.join('-')}>
         {row.map((templateName) => {
           const template = templates[templateName];
           return (
             <TemplateSite
+              key={templateName}
               name={templateName}
               index={index++} // eslint-disable-line no-plusplus
               thumb={template.thumb}


frontend/components/AddSite/TemplateSiteList/index.js
```diff
@@ -63,11 +63,12 @@ export class TemplateList extends React.Component {
     let index = 0;
 
     const templateGrid = templateRows.map(row => (
-      <div className={`flexbox-grid flexbox-grid-${MAX_CELLS_PER_ROW}`} key={row}>
+      <div className={`flexbox-grid flexbox-grid-${MAX_CELLS_PER_ROW}`} key={row.join('-')}>
```
el-mapache: I think this can just be the second argument map passes to its calback, the `index`. I think I wrote the `createRowsOf` method, but maybe we could just clean it up in this PR? I don't think it's necessary since we only have 4 templates and we are creating the rows with flexbox grid classes

jseppi: Using an index or other static counter as a `key` is bad practice with React because the `key` is how it determines if the child should be re-rerendered. [This article](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318) explains the reasoning in depth, but if the _contents_ of an array are mutated while the number of items do not, then the keys will remain the same (`1..N`) and React might not know to re-render the changed contents.

Our linter is also setup to throw an error if it detects an index counter used as a `key` (ref: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L316)

jseppi: I _think_ we're planning to add more templates, so we might still want the flexibility of `createRowsOf`. Unsure though.

el-mapache: I would argue that mutating the data is pretty bad practice too, but if the linter doesn't like it, thats good enough for me

jseppi: yeah, for sure :)

PR #1373, opened 2017-12-14
Content3

OpenGlobe:
content edits for discussion - changes to landing page


views/base.njk
```diff
@@ -65,7 +65,7 @@
 
                 {% if not isAuthenticated %}
                   <li>
-                    <a href="/content/examples/" rel="noopener noreferrer">Example sites</a>
+                    <a href="/content/examples/" rel="noopener noreferrer">Examples</a>
```
wslack: I'm not sure the noun is otherwise clear here....

views/home.njk
```diff
@@ -7,34 +7,36 @@
       <div class="usa-grid">
         <div class="usa-hero-callout usa-section-dark">
           <h1>
-            Federalist compliantly hosts federal government websites.
+            Launch a site in 30 minutes.
```
wslack: Per https://docs.google.com/document/d/1Sf5ie1eUqsiqQbZAjgpFPfnbr5ZfTXWSGqm4QktOsG4/edit# - I don't know if the launch speed is the key need for people. Federalist is *technically capable* of it but the compliance is more important for folks, I think.

views/home.njk
```diff
@@ -7,34 +7,36 @@
...
+          <p>Select one of our pre-built designs and launch a fully compliant website for your federal agency, program or office today.</p>
+          <p>
+          <p><a class="usa-button" href="https://docs.google.com/forms/d/1iB8aW7c9r1QH3s8XElQCrnXRGjAiPUYpWG1CMeEqGIo/viewform">Get started</a></p>
         </div>
       </div>
     </div>
     <div class="build-count">
       <div class="usa-grid text-center">
-        <p>{{ buildCount }} site updates deployed in the past week.</p>
+        <p>Number of site updates this week: {{ buildCount }}</p>
```
wslack: maybe say "number of site updates deployed by federalist" or something? Also, is "this week" clear - that seems to imply a time segment of Sunday-Saturday that resets each weekend.

views/home.njk
```diff
@@ -7,34 +7,36 @@
...
       </div>
     </div>
     <div class="usa-grid">
       <div class="usa-width-one-third home-feature">
         <div class="feature-text">
-          <h2><img src="/images/lock.svg"/> Secure and scalable</h2>
-          <p>Federalist websites offer the highest level of security and scalability. They are generated static sites hosted on FedRAMP certified cloud infrastructure.</p>
+          <h2><img src="/images/layers.svg"/> No experience necessary</h2>
```
wslack: I don't know if we can say that for Chris the Communicator. Most customers do need HTML/front end experience do accomplish their needs, unless their needs are simple.

views/home.njk
```diff
@@ -7,34 +7,36 @@
...
+          <p>Choose from one of our pre-built templates and add your content with our web-based editor. You don't have to be a programmer to get started.  </p>
```
wslack: is this misleading since its not "our" editor?

line47: I would agree with changing "our" in this instance

views/base.njk
OpenGlobe: @tram what do you think?  Portfolio perhaps or something else?  "Example sites" also seems to almost make it sound like they're not real.

views/home.njk
OpenGlobe: From that document it seems like ease of use and speed are key factors.  I think the compliance is a feature, but I don't know if it should be the primary focus of the lead messaging?

tram: Maybe split the difference? Something like "Launch a compliant site in 30 minutes."

wslack: I think _speed of updates_ is more important than _speed of launch_. The other stat might be that it only takes three weeks to do the paperwork and tech work to launch a site - the technical launch speed by itself is less important.

annahsebok: Made some edits here to enforce ease of us (in addition to adding compliance).

line47:
I'm happy to work through the merge conflict @wslack

wslack:
@line47 I don't recall - was it your orJames in discussion with Chris in this PR?

line47:
### Screenshot of the updated homepage content

localhost_1337_

wslack:
Not sure if there's a doc to say this in, but imo we probably shouldn't say "GitHub" on the homepage - that feels a little "endorse-y" to me - but I get the utility of it.

jseppi:
Closing in favor of #1582

PR #1378, opened 2017-12-15
Redirect user to sites list when they remove themselves from a site

el-mapache:
**GIF**:

fed-remove-myself

### DIFF
--- a/frontend/actions/siteActions.js
+++ b/frontend/actions/siteActions.js
@@ -19,6 +19,14 @@ const alertError = (error) => {
   alertActions.httpError(error.message);
 };
 
+const onUserRemoveFromSite = (site) => {
+  if (site) {
+    dispatchUserRemovedFromSiteAction(site);
+  }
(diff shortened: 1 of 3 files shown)


frontend/actions/siteActions.js
```diff
@@ -19,6 +19,14 @@ const alertError = (error) => {
   alertActions.httpError(error.message);
 };
 
+const onUserRemoveFromSite = (site) => {
+  if (site) {
+    dispatchUserRemovedFromSiteAction(site);
+  }
+
+  alertActions.alertSuccess('User successfully removed.');
```
jseppi: Well, I guess this would be cool if it changed the message based on if it was `me`. I think you could just pass it along from `removeUserFromSite`. Maybe in a separate PR?

el-mapache: Yeah agreed, I will put it in a new PR.

Issue #1242, opened 2017-10-26
Add a template to Federalist based on our guides styling


wslack:
**Definition of done**
- [ ] When we can replace a guides site like https://agile.18f.gov/ with the [Federalist-uswds-template](https://github.com/18F/federalist-uswds-template)

line47:
I have some thoughts here, the [USWDS-Jekyll](https://github.com/18F/uswds-jekyll) and the [guides template](https://github.com/18F/guides-template/) are very similar and I've been working to incorporate any functionality from the guides template into the uswds-jekyll. 

Question: Is there a particular guides site (URL) you have in mind to use at the template?

wslack:
The template would not have any guides content but I think I want to use the styling at https://eng-hiring.18f.gov/ or https://boise.18f.gov/

line47:
So basically this type of style layout (minus the top nav) https://standards.usa.gov/page-templates/docs/#section-heading-h2 

☝️ This is already part of USWDS-Jekyll

wslack:
Hmmmm. That makes sense, but I would love a version from the USWDS site that's a little more stripped down - no headers, no complex footer. Is there a version of the site stripped of that?

wslack:
The other, basic thing about this template is that anyone making a guide in the future would want it, so having the "guides" stuff in the footer is helpful from the getgo.

line47:
Yeah I think these are all easily achieved and would like to get some different versions of the theme up for previews to see what can be done. I'm a bit unsure how we can get the uswds-jekyll setup within federalist so I can build out a few preview URLs with these different variations.

wslack:
What happens when you just add the repo? (I don't know enough about it to see if Federalist can build the repo as is)

line47:
Is there a way to try to add this branch as a template option in a staging environment to see if it's possible? https://github.com/18f/federalist-uswds-template/tree/report-config @jseppi

wslack:
I think we can link to that build out from the branch table but I don't think the backend supports cloning a repo and then building out with a different master branch

line47:
The code in the templates.js file leads me to believe we can point it at a branch other than `master`
```
  report: {
    owner: '18f',
    repo: 'federalist-report-template',
    branch: 'master',
    example: 'https://federalist-report-template.18f.gov',
    title: 'Basic Report Template',
    description: 'A basic template to host a set of pages about a recent report or order.',
    thumb: '/images/federalist-report-template.jpg',
  },```

[...]

PR #1381, opened 2017-12-15
[Federalist application] - Set report template to use a sub branch of uswds-federalist template

line47:
Trying this out to set the branch to `report-config` from the https://github.com/18F/federalist-uswds-template/tree/report-config repo. 

More info - https://github.com/18F/federalist/issues/1242#issuecomment-352091592


wslack:
@line47 can we do this as a new template in the list instead of changing the existing listing?

line47:
Certainly - update the PR incoming

config/templates.js
```diff
@@ -1,13 +1,22 @@
 module.exports = {
   report: {
     owner: '18f',
-    repo: 'federalist-report-template',
+    repo: 'federalist-uswds-template',
```
jseppi: was this change intentional, @line47?

line47: Nope, thanks for catching 👍

line47:
🎉 🎉 🎉

Issue #1316, opened 2017-11-30
[Federalist application] - Add link tracking for outbound links


line47:
Use Google Tag Manager to track outbound links from the Federalist application
Things like links to (GitHub, Documentation, Previews of sites, live versions of sites, etc...)

@laurenancona has offered up to put in a PR around this

jseppi:
What's blocking this one?

line47:
I'm unsure how to use GTM to set this up and was going to pair with @laurenancona on it. If this is something that you know how to do @jseppi and we can pair on it that would be awesome!

jseppi:
Not a clue actually :)

laurenancona:
@line47 we can pair on config in the GTM UI if you'd like whenever this is merged to a server we can both see

laurenancona:
Config to add to GTM:
[ ] add GA Pageview tag for 18f account - Federalist profile
[ ] add GA Custom Event tag: Outbound Link Tracking

We'll use this schema for the[ Custom Event parameters](https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide)

| Category | Action | Label |
|:----|:----|:----|
 Exit | Outbound link | {{ URL }}

line47:
Set up a time on Thursday Dec, 21 to pair with @laurenancona to get this knocked out! 🎅

laurenancona:
One more PR (#1400) that should be merged at the same time as #1382 when pushing to prod and this should be 💯

PR #1382, opened 2017-12-15
adds tag manager snippet

laurenancona:
This is the first step toward #1316 - need the container added so we can take advantage of local testing/preview functionality in GTM.

This does not deploy any other code except the currently inert (empty) Tag Manager container.


line47:
@jseppi tagged you on this too just so you can be aware of the addition of GTM

Issue #1384, opened 2017-12-18
Increase test timeout from 1500ms to 2500 (?)


jseppi:
Seems like our tests individually have been taking longer in CircleCI than our specified timeout of 1500ms will allow for. This has resulted in various tests failing, even when logically correct, during CI builds. We should increase the timeout.

PR #1385, opened 2017-12-18
Increase server test timeout to 2500ms

jseppi:
ref #1384


jseppi:
Sure, though it looks to be somewhat random to me. We do have a few that are consistently slow, but the ones that have been failing this morning are not consistent with what I've seen before.

jseppi:
opened #1387

PR #1388, opened 2017-12-18
Adding alert styling for alerts

line47:
Some alerts did not have any styling. 
This adds some consistent styling to mostly informational alerts. 

### Before
screen shot 2017-12-18 at 1 23 32 pm

### After
screen shot 2017-12-18 at 1 33 10 pm

### DIFF
--- a/frontend/components/alertBanner.js
+++ b/frontend/components/alertBanner.js
@@ -1,35 +1,50 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 const propTypes = {
-  message: React.PropTypes.string,
-  status: React.PropTypes.string,
+  message: PropTypes.string,
+  status: PropTypes.string,
+  header: PropTypes.string,
+  children: PropTypes.node,
 };
 
 const defaultProps = {
   message: null,
   status: 'info',
+  children: null,
 };
 
-const AlertBanner = ({ message, status = 'info' }) => {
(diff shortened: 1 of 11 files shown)


el-mapache:
@line47 I think that we can probably make use of the `alertBanner` component for these updates. The banner doesn't include a header section but the component could be modified to accept and display one. Let me know what you think!

line47:
@el-mapache OK that sounds good.

How would I transform something like this to use that component?
```HTML
<div className="usa-alert usa-alert-info" role="alert">
  <div className="usa-alert-body">
    <h3 className="usa-alert-header">No branches were found for this repository.</h3>
    <p className="usa-alert-text">
    Often this is because the repository is private or has been deleted.
    </p>
  </div>
</div>
```

el-mapache:
The component currently looks like this:
```
    <div className="usa-grid">
      <div className="alert-container container">
        <div
          className={`usa-alert usa-alert-${status}`}
          role="alert"
        >
          <div className="usa-alert-body">
            <p className="usa-alert-text">
              { message }
            </p>
          </div>
        </div>
      </div>
    </div>
```
so it is already pretty similar to what you pasted in your comment above.

The current banner component would need to have an `<h>` tag added, and a new property added to pass in the header text.

It would be used like this:
`<AlertBanner status='info' header='great job!' message='that action is ok!' />`

PR #1391, opened 2017-12-18
Begins adding tables for user-actions

el-mapache:
* Add model for user-action
* add model for action-type
* Modifies user model
* begins writing migrations

Going down the road of using separate tables instead of enum fields, this gives us the flexibility to add additional metadata down the line, treats the enum values as data, normalizes the db, etc.

I don't think the foreign keys are specified properly in the models yet, I'll revisit once the migrations are written!

### DIFF
--- a/migrations/20171219144357-add-action-data.js
+++ b/migrations/20171219144357-add-action-data.js
@@ -1,0 +1,17 @@
+/* eslint-disable */
+const actionsExist = 'select count(*) from action_type;';
+
+module.exports.up = (db) =>
+  db.runSql(actionsExist)
+    .then((out) => {
+      if (out.rows.count !== 0) {
+        return db.runSql("insert into action_type (action) values ('add'), ('remove'), ('update')");
+      }
+    })
+    .catch((err) => {
+      console.log(err);
+    });
+
+module.exports.down = (db, callback) => {
+  callback();
+};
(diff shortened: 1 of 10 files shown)


test/api/unit/models/user_action.test.js
```diff
@@ -0,0 +1,70 @@
...
+      model.validate().then(() => {
+        expect(model).to.exist;
+        done();
+      }).catch(done);
+    });
+  });
+
+  describe('validations', () => {
+    it('requires userId, targetId, targetType, actionId, and createdAt', (done) => {
+      const requiredFields = ['userId', 'targetId', 'targetType', 'actionId'];
```
jseppi: missing `createdAt` and `updatedAt`?

migrations/20171219144357-add-action-data.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
el-mapache: no I was being lazy

test/api/unit/models/user_action.test.js
el-mapache: those fields are automatically added by the ORM, so I don't think they need to be tested

jseppi: that's fine -- though you have `createdAt` listed up in the test case name :)

el-mapache: ahh, my mistake!

Issue #1390, opened 2017-12-18
lint:diff is not properly working within the docker-compose environment


jseppi:
`docker-compose run app yarn lint:diff`

Expected:
Only files that have been changed relative to the `staging` branch should be linted.

Actual:
All files appear to be linted.

This is a problem because we only want to check that modified/new files get linted since we are progressively fixing up our codebase as we work instead of linting all files at once. Having the command work locally is crucial so that devs do not have to wait for a result from CircleCI to alert them to issues.

More context:
This used to work before changing to the docker-compose-based development setup. It appears that running `git diff-index --name-only $DIFF_TARGET_BRANCH` always returns all files when run in our `app` container.

jseppi:
Looks like this will get us the listing of files that we need (`git diff` instead of `git diff-index`):

```sh
git diff --name-only staging..$DIFF_TARGET_BRANCH
```

PR #1393, opened 2017-12-18
use git diff instead of git diff-index

jseppi:
in order to properly get list of changed files for our progressive linting strategy.

ref #1390.

### DIFF
--- a/scripts/lint-diff.sh
+++ b/scripts/lint-diff.sh
@@ -3,7 +3,7 @@
 DIFF_TARGET_BRANCH="${DIFF_TARGET_BRANCH:-staging}"
 echo "Linting diff against \"$DIFF_TARGET_BRANCH\" branch"
 
-CHANGED_FILES_LIST=`git diff-index --name-only $DIFF_TARGET_BRANCH | grep -E '.js$|.jsx$' | tr '\n' ' '`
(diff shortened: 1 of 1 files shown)


jseppi:
with 2 files intentionally edited to be linty

**before this change:**

screen shot 2017-12-18 at 4 55 48 pm

...listing of lint problems in every one of our javascript files


**after this change:**

screen shot 2017-12-18 at 4 54 44 pm

scripts/lint-diff.sh
```diff
@@ -3,7 +3,7 @@
 DIFF_TARGET_BRANCH="${DIFF_TARGET_BRANCH:-staging}"
 echo "Linting diff against \"$DIFF_TARGET_BRANCH\" branch"
 
-CHANGED_FILES_LIST=`git diff-index --name-only $DIFF_TARGET_BRANCH | grep -E '.js$|.jsx$' | tr '\n' ' '`
+CHANGED_FILES_LIST=`git diff --name-only $DIFF_TARGET_BRANCH -- '*.js' '*.jsx' | tr '\n' ' '`
```
jseppi: I also removed the piped call to `grep` since `git diff` takes a `path` arg (the `-- '*.js' '*.jsx'` portion of my change)

Issue #1392, opened 2017-12-18
Federalist site add feature fails to display error messages


wslack:
When I try to add a brand new repo in https://federalist.18f.gov/sites/new that's new, and I don't have admin privileges, Federalist fails but displays no error message. Clicking on the button appears to do nothing.

wslack:
Confirmed that I can still add sites when I have `write` permission and they are already in Federalist.

wslack:
I should have noted: I had this happen with https://github.com/gsa/emerging-technology-atlas

saracope:
I just had this happen to me also with Opstat. Clicking the Submit repository-based site does nothing, but I can see the error in the console. 

<img width="1025" alt="screen shot 2017-12-21 at 9 01 36 am" src="https://user-images.githubusercontent.com/2197515/34258858-e317a698-e62d-11e7-805c-b9dd4e17fd0e.png">

jseppi:
Sorry about that, Sarah. We have the fix implemented and will roll it out to production with our release tomorrow.

PR #1394, opened 2017-12-19
redirect to Site List page when adding a site fails

jseppi:
in order to show the alert message with the error from the POST request.
ref #1392

Bonus: cleaned up some component tests that had props missing, and renamed `RenderField` var to `RenderUrlField` to match the filename

### DIFF
--- a/test/frontend/actions/siteActionsTest.js
+++ b/test/frontend/actions/siteActionsTest.js
@@ -149,15 +149,22 @@ describe('siteActions', () => {
       });
     });
 
-    it('triggers an error when adding a site fails', () => {
+    it('triggers an error and triggers a router update to the site list page when adding a site fails', () => {
       const siteToAdd = {
         hey: 'you',
       };
-      addSite.withArgs(siteToAdd).returns(rejectedWithErrorPromise);
+
+      // addSite returns nothing when the POST request fails, so resolve to undefined
+      addSite.withArgs(siteToAdd).returns(Promise.resolve(undefined));
 
       const actual = fixture.addSite(siteToAdd);
 
-      return validateResultDispatchesHttpAlertError(actual, errorMessage);
+      return actual.then(() => {
+        expect(dispatchSiteAddedAction.called).to.be.false;
+        expect(updateRouterToSiteBuildsUri.called).to.be.false;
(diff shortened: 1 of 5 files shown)


jseppi:
This brings back existing behavior that was lost when we started redirecting a successful Add Site action to the Site Builds page for the new site.

It would be a better UX to display the error on the Add Site form itself, but we should handle that separately since this bug is currently in produciton.

test/frontend/actions/siteActionsTest.js
```diff
@@ -149,15 +149,22 @@ describe('siteActions', () => {
...
 
-    it('triggers an error when adding a site fails', () => {
+    it('triggers an error and triggers a router update to the site list page when adding a site fails', () => {
       const siteToAdd = {
         hey: 'you',
       };
-      addSite.withArgs(siteToAdd).returns(rejectedWithErrorPromise);
+
+      // addSite returns nothing when the POST request fails, so resolve to undefined
+      addSite.withArgs(siteToAdd).returns(Promise.resolve(undefined));
```
el-mapache: I didn't think you needed to pass undefined to Promise.resolve, does the test complain without an explicit value?

jseppi: you're right, will remove

jseppi:
want to give this another once-over, @el-mapache?

el-mapache:
@jseppi yes!

test/frontend/components/addSite/AddRepoSiteForm.test.jsx
```diff
@@ -21,6 +21,7 @@ describe('<AddRepoSiteForm />', () => {
   it('renders', () => {
     const props = {
       showAddNewSiteFields: false,
+      initialValues: { engine: 'jekyll' },
```
el-mapache: We might want to add a test that this prop is being supplied to the rendered component. It may be indicative of some missing coverage since the tests passed without this field?

jseppi: `initialValues` is passed along by ReduxForm. I made it a required prop for development ease. see: https://github.com/18F/federalist/blob/staging/frontend/components/AddSite/AddRepoSiteForm.jsx#L24-L28

PR #1396, opened 2017-12-20
Add migrations to support paranoid models

el-mapache:
* User and Site models will be soft deleted, allowing us to track
actions taken on them
* Updates test bootstrap file to comply with linter
* Adds a migration reset script to package.json

### DIFF
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -165,6 +165,8 @@ module.exports = (sequelize, DataTypes) => {
       afterValidate,
       validationFailed,
     },
+    timestamps: true,
+    paranoid: true,
   });
 
   Site.withUsers = id => Site.findById(id, { include: [sequelize.models.User] });
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -165,7 +165,6 @@ module.exports = (sequelize, DataTypes) => {
       afterValidate,
       validationFailed,
     },
-    timestamps: true,
     paranoid: true,
   });
 
--- a/test/api/bootstrap.test.js
+++ b/test/api/bootstrap.test.js
@@ -1,6 +1,6 @@
 Promise.props = require('promise-props');
 
-const AWSMocks = require('./support/aws-mocks'); // eslint-disable-line no-unused-vars
+require('./support/aws-mocks');
 const sequelize = require('../../api/models').sequelize;
 
 const models = sequelize.models;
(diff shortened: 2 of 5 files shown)


test/api/bootstrap.test.js
```diff
@@ -1,19 +1,29 @@
-Promise.props = require("promise-props")
+Promise.props = require('promise-props');
 
-const AWSMocks = require("./support/aws-mocks")
+const AWSMocks = require('./support/aws-mocks'); // eslint-disable-line no-unused-vars
```
jseppi: since the actual variable is never used, maybe we could just leave out the assignment and the eslint-disable:

```js
require('./support/aws-mocks');
```

api/models/site.js
```diff
@@ -165,6 +165,8 @@ module.exports = (sequelize, DataTypes) => {
       afterValidate,
       validationFailed,
     },
+    timestamps: true,
+    paranoid: true,
```
jseppi: interesting, didn't even know that this was part of sequelize. neat!

package.json
```diff
@@ -55,6 +55,7 @@
     "migrate:up": "node migrate.js up || true",
     "migrate:down": "node migrate.js down",
     "migrate:test": "node migrate.js up --dry-run",
+    "migrate:reset": "node migrate.js reset",
```
jseppi: helpful

api/models/site.js
```diff
@@ -165,6 +165,8 @@ module.exports = (sequelize, DataTypes) => {
       afterValidate,
       validationFailed,
     },
+    timestamps: true,
```
jseppi: I think the `timestamps` option defaults to `true`, right? Specifying it explicitly, while not really harming anything, makes it less clear that other models without it specified also have timestamps enabled.

el-mapache: Yeah, I made it explicit because I was seeing weird errors i which sequelize would complain that it couldn't find the field 'deletedAt' on the models; but those errors might have been mitigated by adding the `sequelize.sync` call to our test bootstrap file. I'll try removing the `timestamps` declaration.

test/api/bootstrap.test.js
el-mapache: Thats a better idea. We might actually be able to do that in the `models/index` file as well, since we require all the models just to inject them into the global scope.

api/models/site.js
el-mapache: Yes, its realllllly helpful to not have to write this logic ourselves!

Issue #1398, opened 2017-12-21
S3 file viewing doesn't show folders for "static" builds


wslack:
Example: https://federalist.18f.gov/sites/334/published/master121917

We thus run out of room for viewing all of the files on this screen.

<img width="1026" alt="screen shot 2017-12-21 at 10 37 44 am" src="https://user-images.githubusercontent.com/10016577/34262672-09fe3d32-e63b-11e7-8c01-1601ab5f4203.png">

jseppi:
duplicate of #1362?

PR #1402, opened 2017-12-21
Improve retrieval of and paginate published files

jseppi:
**work-in-progress**, but ready for preliminary approach review

To Do:
- [x] Implement review suggestions
- [x] Increase test coverage of `SitePublishedFilesTable.jsx`
- [x] Determine if `nav` is appropriately used to house `Next` and `Previous` buttons
- [x] Determine if re-factoring functions per CodeClimate check is appropriate/worthwhile

ref #1398 

Looks like this:

paged-files


jseppi:
hrm, CodeClimate doesn't like the length of some functions. I don't think reducing their lines will necessarily lead to improved readability though:

screen shot 2018-01-02 at 10 46 38 am

jseppi:
Need to improve coverage in published files table component:

screen shot 2018-01-02 at 10 49 55 am

api/services/S3SiteRemover.js
```diff
@@ -38,33 +38,22 @@ const deleteObjects = (objects) => {
...
-      if (err) {
-        reject(err);
-      } else {
-        const keys = data.Contents.map(object => object.Key);
-        resolve(keys);
-      }
-    });
-  });
+const getKeys = prefix =>
+  S3Helper.listObjects(prefix)
```
jseppi: Using the new helper here should mitigate https://github.com/18F/federalist/issues/1416

frontend/sass/_header.scss
```diff
@@ -1,60 +1,60 @@
 header {
   background-color: $header-bg-color;
-}
 
-$nav-padding: .45em;
+  $nav-padding: .45em;
```
jseppi: nested all these inside of `header` since I didn't want this styling to necessarily apply to the `nav` element in `SitePublishedFilesTable`

jseppi: this was when I thought I could actually page to random pages of results and would need a fuller paging implementation. perhaps best to revert this change and remove the `nav` element that holds the "Previous" and "Next" buttons from `SitePublishedFilesTable`. Though those buttons are semantically a form of navigation. hrm...

api/services/S3SiteRemover.js
```diff
@@ -10,25 +11,24 @@ const s3Client = new AWS.S3({
...
-
   Since AWS limits the number of objects that can be deleted at a time, this
   method deletes the objects 1000 at a time. It does so recursively so each
   group of 1000 is deleted one after the other instead of simultaneously. This
   prevents the delete requests from breaking AWS's rate limit.
 */
-const deleteObjects = (objects) => {
-  if (objects.length === 0) {
+const deleteObjects = (keys) => {
+  if (keys.length === 0) {
```
el-mapache: can be simplified to `if (!keys.length)`

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+
   componentDidMount() {
     const site = { id: this.props.params.id };
     const branch = this.props.params.name;
-    publishedFileActions.fetchPublishedFiles(site, branch);
+
+    const startAtKey = null; // start without a startAtKey
+    publishedFileActions.fetchPublishedFiles(site, branch, startAtKey);
+
+    autoBind(this, 'previousPage', 'nextPage', 'shouldDisablePreviousPage',
```
el-mapache: Is `autoBind` supposed to be called here? Generally binding like this happens in the constructor. Just wondering!

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const publishedFiles = nextProps.publishedFiles;
+
+    if (publishedFiles.data && !publishedFiles.isLoading) {
+      const files = publishedFiles.data.files || [];
+
+      // save the current page of files state
+      const filesByPage = this.state.filesByPage;
```
el-mapache: I would just write 
`const { filesByPage, currentPage } = this.state`

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+      this.setState({ filesByPage });
+
+      if (!publishedFiles.data.isTruncated) {
+        this.setState({ lastPage: currentPage });
+      }
+    }
+  }
+
+  shouldDisablePreviousPage() {
+    return this.state.currentPage === 0;
```
el-mapache: `return !this.state.currentPage`, unless we need to know if `currentPage` has been defined.

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+  }
+
+  previousPage() {
+    if (this.state.currentPage > 0) {
+      this.setState({ currentPage: this.state.currentPage - 1 });
```
el-mapache: should we clamp this at zero? Could it put the app in a weird state ( and does it matter )?

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+    }
   }
 
-  publishedFiles() {
+  shouldDisableNextPage() {
+    if (this.state.lastPage !== null) {
+      return this.state.currentPage === this.state.lastPage;
```
el-mapache: You should just be able to return line 62 here, `null` won't be equivalent to `0` and will return false.

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+    }
+    return false;
+  }
+
+  nextPage() {
+    const currentPage = this.state.currentPage;
+    const nextPage = currentPage + 1;
+
+    if (this.state.lastPage !== null && currentPage === this.state.lastPage) {
```
el-mapache: could we just call `this.shouldDisableNextPage` function here?

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+        </p>
+        <table className="usa-table-borderless">
+          <thead>
+            <tr>
+              <th>File</th>
+              <th>Actions</th>
+            </tr>
+          </thead>
+          <tbody>
+            { files.filter(f => !!f.name).map(this.renderBranchFileRow) }
```
el-mapache: What causes files to not have names? Just curious 🐈

frontend/sass/_header.scss
el-mapache: The buttons can definitely stay wrapped inside a nav, is it possible to not tie the `nav` styles directly to the tag? The styles controlling the buttons on the `SitePublishedFilesTable` should probably live in a class anyway.

api/services/S3SiteRemover.js
```diff
@@ -10,25 +11,24 @@ const s3Client = new AWS.S3({
...
     return Promise.resolve();
   }
 
-  const objectsToDeleteNow = objects.slice(0, 1000);
-  const objectsToDeleteLater = objects.slice(1000, objects.length);
+  const keysToDeleteNow = keys.slice(0, 1000);
```
el-mapache: Do we have a global constants file for the backend? Might be nice to use the S3_MAX_FILES constant here, so there is some context for why we choose 1000

api/services/S3Helper.js
```diff
@@ -0,0 +1,128 @@
...
+      // call recursively
+      listObjectsHelper(objects, newExtraParams, opts, callback);
+    } else {
+      // done !
+      callback(null, objects);
+    }
+  });
+}
+
+function resolveCallback(resolve, reject) {
```
el-mapache: Why this construct vs wrapping the listObjectHelper  function in a promise? I suppose this is easier to think about than a recursive promise.

api/services/S3Helper.js
```diff
@@ -0,0 +1,128 @@
...
+    const aggregationKey = opts.aggregationKey || 'Contents';
+    const totalMaxObjects = opts.totalMaxObjects;
+
+    const objects = currObjects ?
+      currObjects.concat(data[aggregationKey])
+      : data[aggregationKey];
+
+    // if results should be limited
+    if (totalMaxObjects) {
+      // We're ready to exit if either the response data are not truncated
```
el-mapache: I think T=this behavior can be broken out into a named function, which will hopefully make CC happy

api/services/S3Helper.js
```diff
@@ -0,0 +1,128 @@
...
+
+const S3_DEFAULT_MAX_KEYS = 1000;
+
+const s3Client = new AWS.S3({
+  accessKeyId: s3Config.accessKeyId,
+  secretAccessKey: s3Config.secretAccessKey,
+  region: s3Config.region,
+});
+
+function listObjectsHelper(currObjects, extraS3Params = {}, opts = {}, callback) {
```
el-mapache: you can also cheat the line count by just returning the callback on the same line, rather than returning undefined.

api/services/S3Helper.js
jseppi: yeah, that's the reason. I found the callback easier to think about with the recursion.

frontend/components/site/SitePublishedFilesTable.jsx
```diff
@@ -1,41 +1,141 @@
...
+          </tbody>
+        </table>
+        {this.shouldShowButtons() &&
+          <nav className={'pagination'} aria-label="Pagination">
```
el-mapache: This `nav` element could get moved into a named function as well.

api/services/S3SiteRemover.js
jseppi: good idea. the changes in this file were mostly just renaming variables, so i didn't think too much about them.

frontend/components/site/SitePublishedFilesTable.jsx
jseppi: you're right - bad copy-paste, thanks!

frontend/components/site/SitePublishedFilesTable.jsx
jseppi: Our redirect object constructs (like for redirecting from `some-site.gov/abc` to `some-site.gov/abc/` end up being "nameless."

frontend/sass/_header.scss
jseppi: Yeah, I was just going for minimal changes to the existing `nav` styling. That element was previously only used in our main site `header` element.

frontend/components/site/SitePublishedFilesTable.jsx
jseppi: 👍, should be clamped by the surrounding `if` statement

jseppi:
Alrighty, should be good for final review, @el-mapache.

jseppi:
If you want a big repo with lots of files to test the paging locally, you can fork https://github.com/jseppi/test-federalist-big

or here's the script i used to make all the files:

```sh
#! /bin/bash
for n in {1..2005}; do
    base64 /dev/urandom | head -c 100 > file$( printf %04d "$n" ).txt
done
```

PR #1414, opened 2017-12-26
Add front-end table, associations to user_action

el-mapache:
**Screenshot**:
screen shot 2017-12-27 at 11 13 13 am

### DIFF
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -132,6 +132,7 @@ module.exports = {
         userId: req.user.id,
         targetId: userId,
         targetType: 'user',
+        siteId: site.id,
       })
     )
     .then(() => sendJSON(site, res))
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -2,6 +2,7 @@ const authorizer = require('../authorizers/site');
 const S3SiteRemover = require('../services/S3SiteRemover');
 const SiteCreator = require('../services/SiteCreator');
 const SiteMembershipCreator = require('../services/SiteMembershipCreator');
+const UserActionCreator = require('../services/UserActionCreator');
 const siteSerializer = require('../serializers/site');
 const { User, Site, Build } = require('../models');
 const siteErrors = require('../responses/siteErrors');
@@ -124,10 +125,13 @@ module.exports = {
       return authorizer.removeUser(req.user, site);
     })
     .then(() =>
-      SiteMembershipCreator.revokeSiteMembership({
-        user: req.user,
-        site,
-        userId,
+      SiteMembershipCreator.revokeSiteMembership({ user: req.user, site, userId })
+    )
+    .then(() =>
+      UserActionCreator.addRemoveAction({
+        userId: req.user.id,
+        targetId: userId,
+        targetType: 'user',
       })
     )
     .then(() => sendJSON(site, res))
--- a/api/models/action-type.js
+++ b/api/models/action-type.js
@@ -1,6 +1,6 @@
 const associate = ({ ActionType, UserAction }) => {
-  ActionType.belongsTo(UserAction, {
-    foreignKey: 'actionTypeId',
+  ActionType.hasMany(UserAction, {
+    foreignKey: 'actionId',
   });
 };
 
@@ -16,6 +16,7 @@ module.exports = (sequelize, DataTypes) => {
     classMethods: {
       associate,
     },
+    timestamps: false,
   });
 
   return ActionType;
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -16,7 +16,7 @@ const associate = ({ User, Build, Site, UserAction }) => {
     as: 'actionTarget',
     foreignKey: 'targetId',
     unique: false,
-  })
+  });
 };
 
 function toJSON() {
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -1,3 +1,9 @@
+const protectedAttributes = [
+  'githubAccessToken',
+  'githubUserId',
+  'signedInAt',
+  'site_users__user_sites',
+];
 const associate = ({ User, Build, Site, UserAction }) => {
   User.hasMany(Build, {
     foreignKey: 'user',
@@ -20,25 +26,23 @@ const associate = ({ User, Build, Site, UserAction }) => {
 };
 
 function toJSON() {
-  const object = this.get({
+  const record = this.get({
     plain: true,
   });
-
-  delete object.githubAccessToken;
-  delete object.githubUserId;
-  delete object.signedInAt;
-  delete object.site_users__user_sites;
-
-  object.createdAt = object.createdAt.toISOString();
-  object.updatedAt = object.updatedAt.toISOString();
-
-  Object.keys(object).forEach((key) => {
-    if (object[key] === null) {
-      delete object[key];
+  /* eslint-disable no-param-reassign */
+  return Object.assign({}, Object.keys(record).reduce((out, attr) => {
+    if (protectedAttributes.indexOf(attr) !== -1) {
+      return out;
     }
-  });
 
-  return object;
+    out[attr] = record[attr];
+
+    return out;
+  }, {}), {
+    createdAt: record.createdAt.toISOString(),
+    updatedAt: record.updatedAt.toISOString(),
+  });
+  /* eslint-enable */
 }
 
 module.exports = (sequelize, DataTypes) => {
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -9,7 +9,14 @@ const associate = ({ User, Build, Site, UserAction }) => {
   });
   User.hasMany(UserAction, {
     foreignKey: 'userId',
+    as: 'userActions',
   });
+  User.belongsToMany(User, {
+    through: 'user_action',
+    as: 'actionTarget',
+    foreignKey: 'targetId',
+    unique: false,
+  })
 };
 
 function toJSON() {
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -29,20 +29,19 @@ function toJSON() {
   const record = this.get({
     plain: true,
   });
-  /* eslint-disable no-param-reassign */
+
   return Object.assign({}, Object.keys(record).reduce((out, attr) => {
     if (protectedAttributes.indexOf(attr) !== -1) {
       return out;
     }
 
-    out[attr] = record[attr];
+    out[attr] = record[attr]; // eslint-disable-line no-param-reassign
 
     return out;
   }, {}), {
     createdAt: record.createdAt.toISOString(),
     updatedAt: record.updatedAt.toISOString(),
   });
-  /* eslint-enable */
 }
 
 module.exports = (sequelize, DataTypes) => {
--- a/api/serializers/user.js
+++ b/api/serializers/user.js
@@ -4,7 +4,7 @@ const serializeObject = user => user.toJSON();
 const includedModels = [{
   model: UserAction,
   as: 'userActions',
-  attributes: ['id', 'targetType', 'createdAt'],
+  attributes: ['id', 'targetType', 'siteId', 'createdAt'],
   include: [{
     model: User,
     as: 'actionTarget',
--- a/api/serializers/user.js
+++ b/api/serializers/user.js
@@ -1,23 +1,33 @@
-const { Build, User, Site } = require('../models');
+const { User, UserAction, ActionType } = require('../models');
 
-const serializeObject = (user) => {
-  const json = user.toJSON();
-  json.sites = user.Sites.map(site => site.toJSON());
-  json.builds = user.Builds.map(build => build.toJSON());
-  delete json.Sites;
-  delete json.Builds;
-  return json;
-};
+const serializeObject = user => user.toJSON();
 
 const serialize = (serializable) => {
   if (serializable.length !== undefined) {
     const userIds = serializable.map(user => user.id);
-    const query = User.findAll({ where: { id: userIds }, include: [Site, Build] });
+    const query = User.findAll({ where: { id: userIds } });
 
     return query.then(users => users.map(serializeObject));
   }
-  const user = serializable;
-  const query = User.findById(user.id, { include: [Site, Build] });
+
+  const query = User.findById(serializable.id, {
+    include: [{
+      model: UserAction,
+      as: 'userActions',
+      attributes: ['targetType', 'createdAt'],
+      include: [
+        {
+        model: User,
+        as: 'actionTarget',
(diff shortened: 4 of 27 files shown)


api/serializers/user.js
```diff
@@ -1,23 +1,33 @@
...
-  delete json.Builds;
-  return json;
-};
+const serializeObject = user => user.toJSON();
 
 const serialize = (serializable) => {
   if (serializable.length !== undefined) {
     const userIds = serializable.map(user => user.id);
-    const query = User.findAll({ where: { id: userIds }, include: [Site, Build] });
+    const query = User.findAll({ where: { id: userIds } });
```
el-mapache: I removed the eager loading of `Site` and `Build` because we don't use them at all on the frontend

jseppi:
Does the table list actions just made by the current user, or all actions for all users associated with the site?

el-mapache:
only the current user.

el-mapache:
I can add 'Your' to the table legend, to make it clear. Ideally we get some feedback from users on how to best display this info.

wslack:
Why not have one table for the entire site (so everyone can see the log?)

Agreed on getting feedback.

el-mapache:
I feel like one table with the actions taken by every user is information overload, and I don't have a clean way to represent that in my mind at this point.

It might be useful to see a record of all actions, but shoehorning it in to the users view is not the best UX IMO

jseppi:
I agree that a table of all actions by all users of a site is likely the most useful for auditing, which I think is the original intent of the feature. I think the back-end code here could easily support this kind of table, right?

I also agree it shouldn't be shoehorned into the user table page, but rather on its own page.

line47:
I agree on making it's own page. I think adding a link to view the user actions log table would make sense. 

Something like this that would then take you to the table of user actions log 

screen shot 2018-01-02 at 3 08 42 pm

wslack:
I like that; could certainly be linked from the users page. I think we can leave a user's own actions on their own page and build the full auditable log as a separate PR; what do y'all think?

jseppi:
Can we move this discussion to a new issue so we don't 1) further cloud this PR and 2) don't lose the context when this PR is merged?

el-mapache:
I'll open an issue

frontend/components/site/UserActionsTable.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { dayAndDate } from '../../util/datetime';
+
+const propTypes = {
+  userActions: PropTypes.arrayOf(USER_ACTION),
+};
+const defaultProps = {
+  userActions: [],
+};
+
+const UserActionsTable = ({ userActions }) =>
```
jseppi: I think we'll want to render an empty message or a row that says "no actions logged yet" or something when the table is empty, since just the heading row looks a little weird:

screen shot 2018-01-02 at 2 59 45 pm

test/frontend/actions/siteActionsTest.js
```diff
@@ -281,10 +266,11 @@ describe('siteActions', () => {
   });
 
   describe('.removeUserFromSite', () => {
-    it('triggers an http success alert when user is removed', () => {
+    it('triggers an http success alert when user is removed, and refetches user data', () => {
```
jseppi: 👍

frontend/components/site/UserActionsTable.jsx
```diff
@@ -0,0 +1,43 @@
...
+  <table>
+    <caption><b>Action Log</b></caption>
```
jseppi: 👍 I think we should open a separate issue to use `<caption>` more broadly (in place of the typical `<h#>` elements we have been using), and style them consistently.

jseppi: Maybe remove the `<b>` for now and rely on a future caption-styling PR, though.

api/services/UserActionCreator.js
```diff
@@ -0,0 +1,23 @@
...
+      userId,
+      targetId,
+      targetType,
+      actionId: action.id,
+    })
+  );
+
+const addRemoveAction = buildUserAction('remove');
+const addUpdateAction = buildUserAction('update');
+const addCreateAction = buildUserAction('add');
```
jseppi: cool

api/serializers/user.js
```diff
@@ -1,23 +1,31 @@
...
-const serializeObject = (user) => {
-  const json = user.toJSON();
-  json.sites = user.Sites.map(site => site.toJSON());
-  json.builds = user.Builds.map(build => build.toJSON());
-  delete json.Sites;
-  delete json.Builds;
-  return json;
-};
+const serializeObject = user => user.toJSON();
+const includedModels = [{
```
jseppi: 👍

api/models/user.js
```diff
@@ -9,29 +15,34 @@ const associate = ({ User, Build, Site, UserAction }) => {
...
+    }
 
-  delete object.githubAccessToken;
-  delete object.githubUserId;
-  delete object.signedInAt;
-  delete object.site_users__user_sites;
-
-  object.createdAt = object.createdAt.toISOString();
-  object.updatedAt = object.updatedAt.toISOString();
+    out[attr] = record[attr];
```
jseppi: maybe move the eslint exception just to this specific line

```js
out[attr] = record[attr]; // eslint-disable-line no-param-reassign
```

api/models/user.js
```diff
@@ -9,29 +15,34 @@ const associate = ({ User, Build, Site, UserAction }) => {
...
   });
 };
 
 function toJSON() {
-  const object = this.get({
+  const record = this.get({
     plain: true,
   });
+  /* eslint-disable no-param-reassign */
+  return Object.assign({}, Object.keys(record).reduce((out, attr) => {
```
jseppi: This is cool. We should abstract to a util function and re-use in future code/updates.

api/controllers/site.js
```diff
@@ -124,10 +125,13 @@ module.exports = {
...
     })
     .then(() =>
-      SiteMembershipCreator.revokeSiteMembership({
-        user: req.user,
-        site,
-        userId,
+      SiteMembershipCreator.revokeSiteMembership({ user: req.user, site, userId })
+    )
+    .then(() =>
+      UserActionCreator.addRemoveAction({
```
jseppi: I wonder if in general putting the action creator code into the related service would be a cleaner approach than putting it in the controller code's promise chain. ie, for this one, put the call to `UserActionCreator.addRemoveAction` into `SiteMembershipCreator.revokeSiteMembership`. Thoughts?

jseppi: I don't know if this necessarily needs to be changed now, especially since this is the only use of `UserActionCreator` currently

api/models/action-type.js
```diff
@@ -1,6 +1,6 @@
 const associate = ({ ActionType, UserAction }) => {
-  ActionType.belongsTo(UserAction, {
-    foreignKey: 'actionTypeId',
+  ActionType.hasMany(UserAction, {
+    foreignKey: 'actionId',
```
jseppi: I think `actionTypeId` is clearer here, since it is the FK to the `ActionType` table, right?

screen shot 2018-01-02 at 3 49 39 pm

frontend/actions/siteActions.js
```diff
@@ -83,7 +83,9 @@ export default {
         });
     }
 
-    return onRemoveUser.then(onUserRemoveFromSite);
+    return onRemoveUser
+      .then(userActions.fetchUser)
+      .then(onUserRemoveFromSite);
```
jseppi: Hrm, so on the Users view, I clicked Remove User to remove a user from my site. The User Action table refreshed and showed the new action in it, but the table of users did not refresh (so the removed user was still listed). Maybe the site also needs to be refreshed at this point?

I should have made a gif, but here's a screenshot of the "bad" state showing both the user that I removed and the action that removed the user:

screen shot 2018-01-02 at 3 16 23 pm

frontend/util/datetime.js
```diff
@@ -24,10 +24,23 @@ export const duration = (startTime, endTime) => {
...
   return moment(date).fromNow();
-}
+};
+
+/**
+ * Return a human-readable day, months and year (i.e. Monday, Dec. 25th 2020)
+ * @param  {String | Date} date format "YYYY-DD-DDT00:00:00.000Z"
+ * @return {String}
+ */
+export const dayAndDate = (date) => {
```
jseppi: 👍

test/api/bootstrap.test.js
```diff
@@ -4,26 +4,24 @@ require('./support/aws-mocks');
...
 const models = sequelize.models;
-
+const ActionType = models.ActionType;
+const types = ['add', 'remove', 'update'];
 const cleanDatabase = () => {
   const promises = Object.keys(models).map((name) => {
     const model = models[name];
-    const promise = model.sync()
-      .then(() => model.destroy({ where: {} }));
+    // { force: true } removes soft deleted models and prevents uniqueness validation errors
```
jseppi: remove this line since it is commented out?

api/models/user.js
wslack: oooooooh

api/models/action-type.js
el-mapache: I don't know if its clearer, but IIRC it straight up doesn't work without this foreignKey. I think its less an issue of correct SQL and more appeasing sequelize's API.

test/api/bootstrap.test.js
el-mapache: I guess this is unclear, its literally a comment explaining why `{force: true }` is used.

frontend/components/site/UserActionsTable.jsx
el-mapache: I almost feel like we can just render nothing.

api/models/action-type.js
jseppi: oh, like the string `'actionTypeId'` doesn't work, while `'actionId'` does?

frontend/actions/siteActions.js
el-mapache: Innnnnteresting. I thought it was working properly when I opened the PR but I don't have a GIF to prove it! I'll check this out.

frontend/components/site/UserActionsTable.jsx
jseppi: I'm not opposed to that either

api/models/action-type.js
el-mapache: oh I see what you're saying, I guess we can go back and change it, I think we'll have to write another migration though.

test/api/bootstrap.test.js
jseppi: Oh, I gotcha. That was mostly my bad skimming then. Maybe slight change like:

```js
// Using `force: true` removes soft deleted models and prevents uniqueness validation errors

File: api/models/action-type.js (hunk 10 above)
Comment (el-mapache) [reply api/models/action-type.js]: actionTypeID is the automatic foreignKey prop that sequelize assigns under the hood. I think I set it to this value when I was trying to get these associations to behave properly. It seemed clearer than some of our current foreign key naming conventions.

File: api/controllers/site.js (hunk 9 above)
Comment (el-mapache) [reply api/controllers/site.js]: I think you are probably right and it is something to keep in mind...not sure if we should add this change to this PR. I can do a follow-up with the change, though!

File: frontend/components/site/UserActionsTable.jsx (hunk 4 above)
Comment (el-mapache) [reply frontend/components/site/UserActionsTable.jsx]: I agree!

File: api/controllers/site.js (hunk 9 above)
Comment (jseppi) [reply api/controllers/site.js]: follow-up works for me

**Comment by el-mapache** [comment]
@jseppi should be good to go at this point!

File: frontend/components/site/SiteUsers.jsx [hunk 14]
```diff
@@ -72,6 +80,9 @@ const SiteUsers = ({ site, user }) => {
           )}
         </tbody>
       </table>
+      <div className="offset-top">
+        {renderUserActions(user.userActions)}
```
jseppi: I think we should we add `userActions` to the `USER` propType in `propTypes.js`:

```js
export const USER = PropTypes.shape({
  id: PropTypes.number,
  email: PropTypes.string,
  username: PropTypes.string,
  userActions: PropTypes.arrayOf(USER_ACTION),
});
```

can do in separate PR though.

frontend/components/site/SiteUsers.jsx
```diff
@@ -2,12 +2,20 @@ import React from 'react';
...
 import siteActions from '../../actions/siteActions';
+import UserActionsTable from './UserActionsTable';
 
 import GitHubMark from '../GitHubMark';
 
 const isSiteOwner = (user, siteOwner) =>
   user.username.toLowerCase() === siteOwner.toLowerCase();
 
+const renderUserActions = (userActions) => {
+  if (!userActions.length) {
```
jseppi: I don't know what was going on, but I was getting `userActions` as `undefined` (and thus throwing an error here) when I first switched to this PR after working on something else. It might have been a blip from switching branches though.

Nonetheless, maybe a default value of `[]` should be set in `defaultProps`? Or, since this table is likely going to be moved in the future, maybe don't worry about it here for now.

jseppi: I think something must have been up with the state of my db, because looking at the user serializer, there's no way that should have been `undefined`

frontend/reducers/user.js
```diff
@@ -23,11 +23,7 @@ export default function user(state = initialState, action) {
       return {
         isLoading: false,
         data: {
-          id: action.user.id,
-          username: action.user.username,
-          email: action.user.email,
-          createdAt: action.user.createdAt,
-          updatedAt: action.user.updatedAt,
+          ...action.user,
```
jseppi: 👍

jseppi:
Oh, I just noticed this while clicking around in my local instance: does the User Actions table always show _all_ actions that the user made, regardless of which site is being looked at? 

I removed a user from Site A, but when I go to Site B, I still see the same action logged in the actions table, so it is not clear from _where_ that user was removed. Hrm...we should display and save that info (like User 1 removed User 2 from Site A), though that will require changes to the Action stuff.

Maybe in the meantime we should add some text above the action table that says "These are all of the user removal actions that you have made for all sites that you access to."

el-mapache:
Hmm seem like it does. I only have one site (I'm a bad QA person apparently!) I'll just add the site id to the fetch request

frontend/components/site/SiteUsers.jsx
el-mapache: I didn't know if it should be added since we pass some additional information to the user reducer that isn't enumerated in this propType declaration. Basically I didn't know if we just wanted to validate this subset for some reason

PR #1422, opened 2017-12-27
Pass build id to build `create` action

el-mapache:
* The `create` route handler is a little weird, as we only use to create builds that are being restarted, i.e. we only use it when there is an existing build to get an id from.

### DIFF
--- a/api/controllers/build.js
+++ b/api/controllers/build.js
@@ -41,13 +41,21 @@ module.exports = {
     .catch(res.error);
   },
 
-  create: (req, res) => {
+  /**
+   * Create a new build using data from an existing build
+   * Currently, the only way for a user to directly create a new build is
+   * the `restart build` interface in the site builds view.
+   *
+   * This method is named `restart` as it's
+   * not otherwise possible to create a build via the API.
+   */
+  restart: (req, res) => {
     let params;
 
     Build.findById(req.body.buildId, { include: [Site] })
     .then((build) => {
       if (!build) {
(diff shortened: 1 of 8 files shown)


api/controllers/build.js
```diff
@@ -42,14 +42,23 @@ module.exports = {
   },
 
   create: (req, res) => {
```
jseppi: should we rename this `restart` or `createCopy` or something to make it more obvious?

api/controllers/build.js
```diff
@@ -42,14 +42,23 @@ module.exports = {
...
-      user: req.user.id,
-      commitSha: req.body.commitSha,
-    };
+    let params;
 
-    authorizer.create(req.user, params)
+    Build.findById(req.body.buildId, { include: [Site] })
+    .then((build) => {
+      if (!build) {
+        throw 400;
```
jseppi: `404` might be more appropriate

api/controllers/build.js
jseppi: And/or even just a documentation comment would be good.

api/controllers/build.js
el-mapache: Yes definitely, I will change that!

api/controllers/build.js
el-mapache: I'll do both, naming it `create` doesn't necessarily fit, but I'd want the next developer to know why there isn't a `create` action.

PR #1431, opened 2018-01-02
Updating styles to fix repo name wrap

line47:
### Screenshot
screen shot 2018-01-02 at 12 42 35 pm

🍄  Bonus - Updated the links to `view repo` and `view website` to display inline-block to prevent the text from wrapping until very small screen sizes.

CC @wslack

### DIFF
--- a/frontend/sass/_header.scss
+++ b/frontend/sass/_header.scss
@@ -28,8 +28,8 @@ nav {
 
 nav ul {
   display: flex;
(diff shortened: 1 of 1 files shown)


wslack:
I think this change broke https://federalist-staging.18f.gov/sites

jseppi:
oof, didn't see that it affected that page. opened https://github.com/18F/federalist/issues/1433

PR #1432, opened 2018-01-02
Add create build from branch functionality

el-mapache:
* If the build hasn't been built at least once, a 403 will be thrown
* A future PR will get builds directly form github when an existing
branch cant be found
* We have to use commit shas, because github will complain without one

### DIFF
--- a/api/authorizers/build.js
+++ b/api/authorizers/build.js
@@ -1,26 +1,59 @@
-const { User, Site } = require('../models');
+const { User, Build } = require('../models');
 
-const authorize = (user, build) => (
-  User.findById(user.id, { include: [Site] })
-    .then((userModel) => {
-      const buildSiteId = build.site || build.Site.id;
-      const matchingUserSite = userModel.Sites.find(site => buildSiteId === site.id);
+const shaRegex = /^[a-f0-9]{40}/;
+const validateSha = maybeSha => shaRegex.test(maybeSha);
 
-      if (matchingUserSite) {
-        return Promise.resolve();
-      }
-      return Promise.reject(403);
-    }
-  )
-);
+const verifyBuild = (model) => {
+  const build = model.Builds[0];
+
+  if (!build) {
+    return Promise.reject(403);
+  }
+
+  return build.get({ plain: true });
+};
+
+const getBuildById = (user, params) => {
+  const { buildId, siteId } = params;
+
+  return User.findById(user.id, {
+    include: {
+      model: Build,
+      where: { site: siteId, id: buildId },
+    },
+  })
+  .then(verifyBuild);
+};
+
+const getBuildByBranch = (user, params) => {
+  const { siteId, sha, branch } = params;
+
+  if (!validateSha(sha)) {
+    throw 400;
+  }
+
+  return User.findById(user.id, {
+    include: {
+      model: Build,
+      where: {
+        branch,
+        site: siteId,
+      },
+      order: User.sequelize.literal('"createdAt" desc'),
+      limit: 1,
+    },
+  })
+  .then(verifyBuild)
+  .then(build => Object.assign({}, build, { commitSha: sha }));
+};
+
+const authorize = (user, params) => {
+  const finderFn = params.buildId ? getBuildById : getBuildByBranch;
+
+  return finderFn(user, params);
+};
 
 const findOne = (user, build) => authorize(user, build);
-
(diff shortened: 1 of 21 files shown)


api/authorizers/build.js
```diff
@@ -1,26 +1,59 @@
...
+  return build.get({ plain: true });
+};
+
+const getBuildById = (user, params) => {
+  const { buildId, siteId } = params;
+
+  return User.findById(user.id, {
+    include: {
+      model: Build,
+      where: { site: siteId, id: buildId },
```
jseppi: nice use of the sequelize api

api/authorizers/build.js
```diff
@@ -1,26 +1,59 @@
...
-  return authorize(user, params);
+
+  return User.findById(user.id, {
+    include: {
+      model: Build,
+      where: {
+        branch,
+        site: siteId,
+      },
+      order: User.sequelize.literal('"createdAt" desc'),
```
jseppi: would this work: `order: [['createdAt', 'DESC']],` ?

I'm not that familiar with the ordering API in sequelize, but it looks like that might work based on the first "example" at http://docs.sequelizejs.com/manual/tutorial/querying.html#ordering (`['title', 'DESC'],`)

api/authorizers/build.js
```diff
@@ -1,26 +1,59 @@
...
+  .then(build => Object.assign({}, build, { commitSha: sha }));
+};
+
+const authorize = (user, params) => {
+  const finderFn = params.buildId ? getBuildById : getBuildByBranch;
+
+  return finderFn(user, params);
 };
 
+const findOne = (user, build) => authorize(user, build);
```
jseppi: maybe change `build` here and on the next line also to `params` for readability

api/authorizers/build.js
```diff
@@ -1,26 +1,59 @@
...
-      }
-      return Promise.reject(403);
-    }
-  )
-);
+const { User, Build } = require('../models');
 
-const findOne = (user, build) => authorize(user, build);
+const shaRegex = /^[a-f0-9]{40}/;
+const validateSha = maybeSha => shaRegex.test(maybeSha);
```
jseppi: could we also/instead put this [validator](http://docs.sequelizejs.com/manual/tutorial/models-definition.html#validations) on the `Build` model's `commitSha` field? I think if we did it there we wouldn't have to explicitly put validations elsewhere, and wouldn't accidentally miss something in the future.

api/controllers/build.js
```diff
@@ -41,35 +41,19 @@ module.exports = {
...
-    .then((build) => {
-      if (!build) {
-        throw 404;
-      }
-
-      params = {
-        branch: build.get('branch'),
-        site: build.get('Site').get('id'),
+  create: (req, res) => {
+    authorizer.create(req.user, req.body)
```
jseppi: maybe some comments about what `req.body` could contain would be helpful

api/authorizers/build.js
el-mapache: Good point, totally forgot about that aspect of sequelize. Will move it there.

api/authorizers/build.js
el-mapache: yeah it should work, I just didnt like the tuple syntax, although I think they call it out as their preferred method 😬

jseppi: hehe, I don't feel strongly about either way. I personally like the tuple one better (barely), but it's not a big deal.

el-mapache:
@jseppi  I believe I've addressed your feedback!

jseppi:
When a branch doesn't have a build, and the action throws a `403` as you mentioned, is there any way we could show a better error to the user? As implemented, it will display "You are not authorized to perform that action" which could be confusing. I'm concerned that if this is deployed before the remaining build-from-branch functionality is implemented then we'll get some user confusion from that error message.

el-mapache:
We can show a better error, it just requires adding another branch to the authorizer's `verify` method. Since I restructured the finders to use `include` statements, we no longer have a clear separation between the case of a user not being associated with a site, or with the build not existing.

I think to get around this we'd have to get all builds for a site (rather than the single one we get now) and then filter by commit sha.

el-mapache:
@jseppi Check out the more verbose error I added in my latest commit!

el-mapache:
you got it!

PR #1434, opened 2018-01-02
Updating dashboard header

line47:
Moved the image inside of the heading tag to make consistent with the pages header.

**In addition**
Removed the redundant `Websites` header and made `Dashboard` an `H2` which follows the format of the sub pages.

I think we could even go one step further and remove the `Dashboard` heading as I feel that `Your websites` is enough to convey what is being shown on the page and the word Dashboard isn't necessary. 

# Before screenshot
screen shot 2018-01-02 at 2 55 07 pm

# After screenshot
screen shot 2018-01-02 at 2 53 07 pm

CC @wslack

### DIFF
--- a/frontend/components/siteList/siteList.js
+++ b/frontend/components/siteList/siteList.js
@@ -62,7 +62,10 @@ export const SiteList = ({ storeState }) =>
     <div className="usa-grid dashboard header">
       <div className="usa-width-two-thirds">
         <div className="header-title">
-          <h1><img className="header-icon" src="/images/websites.svg" alt="Websites icon" /> Your websites</h1>
+          <h1>
+            <img className="header-icon" src="/images/websites.svg" alt="Websites icon" />
+            Your websites
+          </h1>
           <h2>Dashboard</h2>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


wslack:
I think we should remove "your websites" or replace "dashboard" with "your websites"

wslack:
100% agree that both aren't needed; good point

line47:
@wslack - I'm thinking of going with `Your websites` like the screesnshot below. 
Let me know if that's good.

screen shot 2018-01-02 at 4 07 03 pm

wslack:
@line47 I would suggest moving the text down - that way all pages have the same title below the icon, which seems more consistant.

This screenshot puts the page title where the "repo name" typically goes.

line47:
What text would go next to the icon ? 

I think it looks strange to have an icon then a new line for the page name, thoughts?

screen shot 2018-01-02 at 4 47 03 pm

wslack:
This is my thought after fiddling for a sec:
screen shot 2018-01-02 at 4 52 45 pm

wslack:
Which, on second thought, can go next to the icon. I would want the text and the icon to have the same height, though.

jseppi:
Is this still a WIP or ready for final review?

line47:
This is ready for review 👍

Issue #1328, opened 2017-12-04
Use monospace font for site config textareas


jseppi:
It will be easier to read and make it a little more obvious that those textareas contain "code"

line47:
Related to settings issue https://github.com/18F/federalist/issues/1324

PR #1436, opened 2018-01-02
Update settings page to fix left alignment make more accessible

line47:
# Updates
- Fixes the left alignment of the settings page content Currently there is a gap that is caused by the `usa-grid` classes that have been removed.
- Adds labels to the text areas for configurations. Currently there is not label for any of the advance settings text boxes and that is an accessibility issue. 
- Groups the basic and advanced settings into a single well instead of multiple .This helps with giving the save buttons some context on what you are saving and clean up the UI a bit
- Sentence cased the headings, labels, buttons, and links as necessary. Did not address page title casing (that's for a diff PR)
- Added `fieldsets` and `legends` for greater accessibility.
- Removed the `usa-grid` and `usa-width-one-whole` divs as the content of the page is already inside a `usa-grid` div. This clean up the code a bit.
- Updated the danger zone to use the proper alert markup for headings and content.
- Updated the config textboxes in the advanced settings to use monospace font, per issue https://github.com/18F/federalist/issues/1328

## Screenshot
localhost_1337_sites_1_settings

CC @wslack

### DIFF
--- a/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
+++ b/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
@@ -50,12 +50,13 @@ export const AdvancedSiteSettings = ({
 
       {/* CUSTOM CONFIG */}
       <fieldset>
-        <legend>Site configuration</legend>
+        <legend>Live site</legend>
         <p className="well-text">
           Add additional configuration in yaml to be added to your
           {' '}
           <code>_config.yml</code> file when we build your site&apos;s primary branch.
         </p>
+        <label htmlFor="config">Site configuration</label>
         <Field
           component="textarea"
           name="config"
@@ -65,12 +66,13 @@ export const AdvancedSiteSettings = ({
 
       {/* DEMO CONFIG */}
       <fieldset>
-        <legend>Demo configuration</legend>
+        <legend>Demo site</legend>
         <p className="well-text">
           Add additional configuration in yaml to be added to your
           {' '}
           <code>_config.yml</code> file when we build your site&apos;s demo branch.
         </p>
+        <label htmlFor="demoConfig">Demo configuration</label>
         <Field
           component="textarea"
           name="demoConfig"
@@ -80,12 +82,13 @@ export const AdvancedSiteSettings = ({
 
       {/* PREVIEW CONFIG */}
       <fieldset>
-        <legend>Preview configuration</legend>
+        <legend>Preview site</legend>
         <p className="well-text">
           Add additional configuration in yaml to be added to your
           {' '}
           <code>_config.yml</code> file when we build a preview branch for your site.
         </p>
+        <label htmlFor="previewConfig">Preview configuration</label>
         <Field
           component="textarea"
           name="previewConfig"
--- a/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
+++ b/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
@@ -47,7 +47,8 @@ export const AdvancedSiteSettings = ({
           />
         }
       />
-
+    </div>
+    <div className="well">
       {/* CUSTOM CONFIG */}
       <fieldset>
         <legend>Live site</legend>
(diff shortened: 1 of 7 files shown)


frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
```diff
@@ -34,105 +34,90 @@ export const AdvancedSiteSettings = ({
...
-          Save Advanced Settings
-        </button>
-      </div>
-    </div>
 
-    <div className="usa-grid">
-      <h3>Danger Zone</h3>
-      <div className="usa-alert usa-alert-delete" role="alert">
-        Delete this site from Federalist?
+    <div className="usa-alert usa-alert-error usa-alert-delete" role="alert">
```
jseppi: hrm, its role isn't really that of an alert though. Like this section isn't shown in response to a user action that triggered an error.

Using `usa-alert-*` classes was more to get the styling than for semantics.

frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
```diff
@@ -34,105 +34,90 @@ export const AdvancedSiteSettings = ({
...
-          <p className="well-text">
-            Add additional configuration in yaml to be added to your
-            {' '}
-            <code>_config.yml</code> file when we build a preview branch for your site.
-          </p>
-          <Field
-            component="textarea"
-            name="previewConfig"
+    <div className="well">
+      <label htmlFor="engine">Static site engine</label>
```
jseppi: What do you think about increasing the font weight of the form element `label`s? I personally like them a little heavier.

frontend/sass/_forms.scss
```diff
@@ -51,3 +57,12 @@ select {
     display: inline;
   }
 }
+
+.form-control-mono {
+  font-family: monospace,monospace;
```
jseppi: Does `monospace` need to be listed twice? (honest question, since it wouldn't surprise me if that was some css/browser-specific thing)

frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
line47: I'm totally OK with adding that back in. I was leaning on the USWDS for guidance.

wslack:
Thanks for this Scott!

We deliberately split those sections into separate boxes because of user confusion that the live site and demo site settings were separate and independent. I agree the padding should be fixed, but I'm not sure merging the boxes is helpful given the past user feedback.

Similarly, the left alignment was on purpose - that shows via offset what is in "basic" vs "advanced." I don't know if we should have those changes in scope w/o supporting user feedback.

The "danger zone" change is good.

frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
line47: Yeah that makes sense, we need a `danger-well` ☠️  rather than an alert

jseppi: Good idea!

wslack: Good point. I'm not opposed to using the alert styling if its cleaner but we shouldn't confuse folks with a "something is wrong" symbol.

frontend/sass/_forms.scss
line47: Nope, good catch... thanks @jseppi

line47:
# Updated screenshot

localhost_1337_sites_1_settings

wslack:
Thanks Scott. Should the "reset" and "save settings" buttons be left aligned with the blue boxes? That's how it is currently.

Also, thinking more since yesterday, there are probably other ways to show what lives in each section besides left-align and I'd be happy to consider options if you want.

line47:
Yeah let's left align them for now to get this PR wrapped up. 

I agree with you that the indentation to show each section feels a bit clunky and you have some things in boxes yet the buttons are not. Let's revisit in another issue.

wslack:
@line47 did you open the other issue or would you like someone else to? I can't tell exactly what you meant.

line47:
Added the issue - feel free to add to/edit https://github.com/18F/federalist/issues/1442

PR #1441, opened 2018-01-04
Removing second header on dashboard page and setting header icon height

line47:
### Screenshots

screen shot 2018-01-04 at 11 35 27 am

screen shot 2018-01-04 at 11 35 36 am

CC @wslack


line47:
Sure thing this final commit was missing from that last related PR.

PR #1450, opened 2018-01-05
add a note with an e-mail instead of a button

wslack:


wslack:
@jseppi should we include this?

jseppi:
Ok with me!

PR #1457, opened 2018-01-05
[WIP] remove icon from menu

wslack:


wslack:
I don't think there are test implications to this, but not 100% sure.

wslack:
(update: after looking at this, we will actually re-add the icons)

jseppi:
Superseded by #1459

PR #1466, opened 2018-01-08
Use site authorizer in build log controller

el-mapache:
* Previously, we used the build authorizer, which was bsaically the same
as the site authorizer code. The site authorizer is acceptable here
since we only need to check if the user is associated with the site before
displaying logs

I added a test checking explicitly that the user was not associated with the build logs being retrieved.


el-mapache:
@jseppi I believe it is still used in the `build` controller, with the `findOne` method.

jseppi:
Ah, yup, you're right. My search for it was not exhaustive enough.

Issue #1448, opened 2018-01-05
[Federalist application] - Consider updating time last published text


line47:
Consider updating time last published text on site list to be more succinct.

- Currently reads `This site was last published at January 5th 2018, 12:34:35 pm.`
- Shorten to read `Last published on January 5th 2018, 12:34:35 pm.`

### Screenshot example 

![screen shot 2018-01-05 at 12 46 03 pm](https://user-images.githubusercontent.com/1449852/34621221-4c260974-f216-11e7-9592-8fe16d46823b.png)

jseppi:
hrm, I wonder what timezone those are in. We specify it! And, if it is UTC or something, maybe standardize them all to Eastern.

jseppi:
Done by #1475

Issue #1301, opened 2017-11-21
Format dates more nicely in log tables


jseppi:
Something slightly more readable.

Perhaps:
`2017-11-13T21:50:25.629Z` -> `2017-11-13 21:50:25 UTC`

<img width="750" alt="screen shot 2017-11-21 at 10 53 43 am" src="https://user-images.githubusercontent.com/697848/33085468-56142b30-ceaa-11e7-8258-d03195756eac.png">

wslack:
@el-mapache interested in this one?

jseppi:
@line47 and/or @OpenGlobe: Any suggestions for formatting these datetimes? A thought I had is that using US Eastern time for them would be nicer than UTC.

So, for example,
Changing `2017-11-13T21:50:25.629Z` to `2017-11-13 4:50:25 PM (ET)` (or something very similar to that)

el-mapache:
In the table currently up there, I think I have Day month and year written out in english. Do we want a timestamp or a friendly date?

jseppi:
@el-mapache: you mean on the User Actions table, right? Perhaps we should switch the Build Logs table to be formatted like the User Actions table so that we're consistent across the two tables.

The Build Logs table just has the ugly default JavaScript datetime-to-string datetimes:

<img width="818" alt="screen shot 2018-01-08 at 2 40 15 pm" src="https://user-images.githubusercontent.com/697848/34691344-ef8774be-f481-11e7-93e5-6374002345e0.png">

el-mapache:
Yeah I'm all for standardizing! Almost any other format is better than that one 😄

el-mapache:
But I also don't really have strong feelings about which one we should use, whatever is friendliest to users

wslack:
4:50 PM 11-13-2017 (ET)

I think this format might be better, unless the seconds ever matter.

line47:
We're using this on the your websites listing 

`January 9th 2018, 11:07:36 am`

jseppi:
Update: the times we display are user localized currently, except for those in the build log table, which are utc. I think that handling is good as is.

jseppi:
My PR will be finished soon and we can iterate on what I’ve implemented there. I’ll add screenshots of the various date(time)s we show.

jseppi:
Done by #1475

Issue #1160, opened 2017-09-14
Consider replacing moment.js in our front end bundle


jseppi:
`moment.js` is ~126 Kb of our bundle, which is nearly the same size as all of our custom code.

We barely use it, so I think we could get rid of in and instead use maybe https://date-fns.org/ and some other smaller humanize lib.

el-mapache:
Might be worth investigating, but I believe we can just include the functions we need now from moment, rather than bringing in the entire package.

jseppi:
Looks like moment still has open this issue from 2015 about making it more modularized: https://github.com/moment/moment/issues/2373

I couldn't find any reference to just bringing in required functions.

el-mapache:
I see, I noticed in their source that they were exporting each function, but I guess they haven't actually exposed that to the user when installing via npm

wslack:
What's the defnition of done and level of effort here @jseppi?

jseppi:
AC:
- moment is no longer in the front end bundle, but datetime formatting still works as desired

LOE:
- already done in #1475 :)

wslack:
LOL. Got it.

jseppi:
Done by #1475

PR #1475, opened 2018-01-09
Various datetime formatting; switch from moment to date-fns in front end bundle

jseppi:
To Do:
- [x] update the user actions table timestamps with the log-like timestamp formatting

Various datetime issues/features, including:

- #1301 - improve timestamp formatting in build log table
- #1448 - update site last published text
- #1160 - replace moment with date-fns in the front end (still used in back end though)

Summary of changes with screenshots:

### Sites List (#1448):

Changed to `Last published on January 9th 2018, 3:20:26 pm.`

screen shot 2018-01-09 at 12 22 47 pm

### Builds Logs Table (#1301)

Updated formatting to be a little more readable, ensure times are UTC. Added "(UTC)" to the column header.

This is the only place that has UTC times. I think they are nice to use here so that everyone looking at/debugging a build has the times in the same "zone," and they align with the timestamps shown in the actual logs from the build container (which are in UTC as well). 

screen shot 2018-01-09 at 12 24 18 pm

### User Actions Table

Updated to use UTC timestamps:

screen shot 2018-01-09 at 5 21 29 pm

### DIFF
--- a/frontend/util/datetime.js
+++ b/frontend/util/datetime.js
@@ -52,19 +52,6 @@ export const timeFrom = (date) => {
 };
 
 /**
- * Return a human-readable day, months and year (i.e. Monday, Dec. 25th 2020)
- * @param  {String | Date} date format "YYYY-DD-DDT00:00:00.000Z"
- * @return {String}
- */
-export const dayAndDate = (date) => {
-  if (!date) {
-    return NO_TIME;
-  }
-
-  return format(date, 'dddd, MMM Do YYYY');
-};
-
-/**
  * Return a human-readable day, months and year (i.e. December 25th 2020 5:22 PM)
  * @param  {String | Date} date format "YYYY-DD-DDT00:00:00.000Z"
  * @return {String}
--- a/frontend/util/datetime.js
+++ b/frontend/util/datetime.js
@@ -1,6 +1,25 @@
-import moment from 'moment';
+import { format, distanceInWords, distanceInWordsStrict } from 'date-fns';
+
 
 const NO_TIME = '-';
+
+/**
+ * Helper to get the given date or `now` in UTC.
+ * @param {String | Date} dateString  date to convert to UTC. Defaults to now.
+ * @return {Date}                     date in UTC
+ */
+function getUTCDate(dateString = Date.now()) {
+  const date = new Date(dateString);
+
+  return new Date(
+    date.getUTCFullYear(),
+    date.getUTCMonth(),
+    date.getUTCDate(),
+    date.getUTCHours(),
+    date.getUTCMinutes(),
+    date.getUTCSeconds()
+  );
+}
 
 
 /**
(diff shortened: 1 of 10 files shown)


jseppi:
Production bundle savings of `date-fns` vs `moment` : `6.42kb` vs `16.63kb` gzipped

screen shot 2018-01-09 at 9 57 44 am

screen shot 2018-01-09 at 9 57 54 am

el-mapache:
What is the api like for date-fns?

jseppi:
It has comparable methods, but is more "functional." Docs are at https://date-fns.org/v1.29.0/docs/format

jseppi:
I left the User Actions table timestamp as it is, but it is a little different from the two others (the last published datetime and the timestamp in the build logs table):

screen shot 2018-01-09 at 12 29 05 pm

Should it be changed to be more like one of the others? cc @line47 @wslack

frontend/util/datetime.js
```diff
@@ -1,7 +1,26 @@
...
 
+/**
+ * Helper to get the given date or `now` in UTC.
+ * @param {String | Date} dateString  date to convert to UTC. Defaults to now.
+ * @return {Date}                     date in UTC
+ */
+function getUTCDate(dateString = Date.now()) {
+  const date = new Date(dateString);
+
+  return new Date(
```
el-mapache: I wonder if there is a formatting helper so we don't have to manually call all these functions (event if it is encapsulated in this function 😄 )

package.json
```diff
@@ -82,6 +82,7 @@
     "concurrently": "^3.4.0",
     "cookie": "^0.3.1",
     "css-loader": "^0.28.4",
+    "date-fns": "^1.29.0",
```
el-mapache: should we also remove moment here?

jseppi: moment is still used on the back-end, but it is now out of the front-end.

frontend/util/datetime.js
jseppi: There is not one that I found. This was a recommended solution from an issue comment thread.

line47:
I feel like in the user actions log and last published dates should be more friendly and follow this format. `January 9th 2018, 3:20:26 pm.` Are these 2 friendly dates based on my local time or no @jseppi ?

While the builds table should be in UTC format.

jseppi:
Yes, currently those two dates (last published and the log of actions) are shown in the viewer's timezone. Unless I'm really misunderstanding something :P

I think the user action one should be more "timestamp-like" than "friendly-like" since it is basically an audit log. The timestamp format is also little more compact, which is nice for a potentially long table. I'll defer to others if anyone feels more strongly about it though.

jseppi:
Here's the full visualization of GZipped module sizes in a production build (notice no `moment` in there):

screen shot 2018-01-09 at 1 27 36 pm

wslack:
@jseppi I feel that the user actions log should certainly go to the specific minute or second.

jseppi:
@wslack: cool, will update the user actions table timestamps with the log-like timestamp formatting.

jseppi:
ready for re-review, @el-mapache and @wslack

test/frontend/components/site/UserActionsTable.test.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import UserActionsTable from '../../../../frontend/components/site/UserActionsTable';
+
+describe('<UserActionsTable/>', () => {
```
jseppi: new test file

jseppi:
Also when I made the new User Actions timestamp screenshot, I noticed the time is 00:00:00. It looks like this is because User Actions only capture the date, and not the datetime, when created. Opened another issue to track that: https://github.com/18F/federalist/issues/1479

PR #1476, opened 2018-01-09
Add more robust UserAction support on front-end

el-mapache:
* Adds reducer, action creators and tests

Screenshot of the new-ish table:

screen shot 2018-01-10 at 11 49 08 am

### DIFF
--- a/api/models/user-action.js
+++ b/api/models/user-action.js
@@ -2,6 +2,7 @@ const validTargetTypes = [['site', 'user']];
 
 const associate = ({ User, UserAction, ActionType, Site }) => {
   UserAction.belongsTo(User, {
+    as: 'initiator',
     foreignKey: 'userId',
   });
   UserAction.belongsTo(User, {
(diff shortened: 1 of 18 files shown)


jseppi:
screenshot of prop validation error:

screen shot 2018-01-10 at 11 12 23 am

el-mapache:
This error is actually already present in our current app, and I'm not really sure how to fix it without defining a default null object schema for the user action on the front end :/

el-mapache:
It's also not marked as required, as far as I can tell

jseppi:
You're right. This is super weird. Perhaps some issue between `connect` and React?

Found this, which is kind of similar: https://github.com/reactjs/react-redux/issues/6

el-mapache:
I see a lot of issues where libraries are throwing this error but haven't figured out the root cause yet

el-mapache:
this feels like a super goofy workaround but I guess it'll have to do

jseppi:
Well, let's work on it later, because like you noted, I don't see anything wrong in the propTypes declarations for the components here.

jseppi:
Filed https://github.com/18F/federalist/issues/1482

PR #1477, opened 2018-01-09
Only add tracking scripts when in production app env

jseppi:
Also do some test refactoring for sharing common things between the main app and content pages

To Do:
- [x] finish test refactoring


jseppi:
code climate's total coverage check appears stuck. going to try triggering another build on circle to re-report it.

PR #1492, opened 2018-01-10
Adds ability to build from branch on github

el-mapache:
Logic works but is a work-in-progress

### DIFF
--- a/api/authorizers/build.js
+++ b/api/authorizers/build.js
@@ -1,36 +1,55 @@
-const { User, Build } = require('../models');
+const { User, Build, Site } = require('../models');
 const buildErrors = require('../responses/buildErrors');
+const Github = require('../services/Github');
 
-const verifyBuild = (model) => {
-  const build = model && model.Builds && model.Builds[0];
+const rejectBuild = () =>
+  Promise.reject({
+    status: 403,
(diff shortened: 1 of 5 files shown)


el-mapache:
I have to fix/add a test but this is ready to be reviewed!

api/authorizers/build.js
```diff
@@ -1,47 +1,76 @@
...
+
+    if (build) {
+      return build;
+    }
+
+    return rejectBuild();
+  });
 };
 
 const getBuildByBranch = (user, params) => {
```
jseppi: Kind of a bigger picture thought: I think the original intent for our `authorizer`s was to just check that a user had access to a specific resource. Seems like we've been leaking a bit more than that (like fetching objects) into the authorizers. We might want to have an architectural review of the approach. I think I'll file a separate `debt` issue for that discussion/evaluation.

jseppi: Filed https://github.com/18F/federalist/issues/1497

el-mapache: I agree, for me it starts to get tricky when the associations are eagerly loaded, it's not clear which resource is the one you don't have access to (or doesn't exist, in this case)

Maybe it's clearer in the long run to just perform the additional DB lookups?

PR #1502, opened 2018-01-12
Remove non federalist users from site users table

el-mapache:
* Adds scope to user model to filter out those who do not have a
federalist account, or are not members of the federalist-users org

### DIFF
--- a/api/models/user.js
+++ b/api/models/user.js
@@ -44,6 +44,24 @@ function toJSON() {
   });
 }
 
+const tableOptions = {
+  tableName: 'user',
+  classMethods: {
+    associate,
+  },
+  instanceMethods: {
+    toJSON,
+  },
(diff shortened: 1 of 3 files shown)


api/models/user.js
```diff
@@ -44,6 +44,24 @@ function toJSON() {
...
+const tableOptions = {
+  tableName: 'user',
+  classMethods: {
+    associate,
+  },
+  instanceMethods: {
+    toJSON,
+  },
+  paranoid: true,
+  scopes: {
```
jseppi: whoooooaaa! I did not know about `scopes` in Sequelize. That looks super useful!

el-mapache: yeah! the only issue with them is that support for includes and nested includes is not great.

PR #1506, opened 2018-01-16
Adds validator to branch name field on AddSite form

el-mapache:
SCREENSHOT:

screen shot 2018-01-16 at 9 49 43 am

### DIFF
--- a/frontend/components/Fields/BranchField.js
+++ b/frontend/components/Fields/BranchField.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Field } from 'redux-form';
+
+const validBranchRegex = /^[^\/|\.]+(?!.*\.{2}|.*\/{2}|^@$|.*@\{)+(?!([\.lock|[\.\/]])).|^$/;
+
+const invalidPrefix = /^[^\/|\.]+/;
+const invalidChars = /(?!.*\.{2}|.*\/{2}|^@$|.*@\{)/;
+const invalidSuffix = /[^\.lock|\.|\/]$/;
+
+//new RegExp('^[^\/|.](?!.*\.{2}|.*\/{2}|^@$|.*@\{)[^\x00-\x1F\x7F\s<>~^:?*[\\]+(?<!\.lock|\.|\/)$');
+
+export const validateBranchName = (value) => {
+  if (value && value.length && !validBranchRegex.test(value)) {
+    return 'Branch name contains invalid characters.';
+  }
+  return undefined;
+};
+
+const BranchField = ({ ...props }) =>
+  <Field
+    validate={[validateBranchName]}
+    {...props}
+  />;
+
+export default BranchField;
--- a/frontend/components/Fields/BranchField.js
+++ b/frontend/components/Fields/BranchField.js
@@ -1,12 +1,10 @@
 import React from 'react';
 import { Field } from 'redux-form';
-import { validBranchName } from '../../util/validators';
+import validBranchName from '../../util/validators';
 import InputWithErrorField from './InputWithErrorField';
 
-const validate = string => validBranchName.test(string);
-
 export const validateBranchName = (value) => {
-  if (value && value.length && !validate(value)) {
+  if (value && value.length && !validBranchName(value)) {
     return 'Branch name contains invalid characters.';
   }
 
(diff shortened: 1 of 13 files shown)


el-mapache:
Do we want to link to some kind of branch name rules? They are pretty complicated and an error message can't really capture all the rules. 

I guess to me this feels more like a check that the user didn't accidentally mistype something (although it should also stop any kind of code from getting inserted into a db record)

test/frontend/components/Fields/BranchField.test.jsx
```diff
@@ -0,0 +1,30 @@
...
+import { validateBranchName } from '../../../../frontend/components/Fields/BranchField';
+
+describe('<BranchField />', () => {
+  it('validates', () => {
+    const msg = 'Branch name contains invalid characters.';
+
+    expect(validateBranchName()).to.be.undefined;
+    expect(validateBranchName('')).to.be.undefined;
+    expect(validateBranchName('branch')).to.be.undefined;
+    expect(validateBranchName('branch/1')).to.be.undefined;
```
jseppi: while this is a valid git branch name, I think we would want to disallow `/` because it would lead to weird pathing behavior when branches are published to S3 (since those paths look like `.../preview/<repo-owner>/<repo-name>/<branch-name>/...`

frontend/components/Fields/BranchField.js
```diff
@@ -0,0 +1,29 @@
...
+  if (value && value.length && !!validate(value)) {
+    return 'Branch name contains invalid characters.';
+  }
+
+  return undefined;
+};
+
+const BranchField = ({ ...props }) =>
+  <Field
+    component={RenderUrlField}
```
jseppi: I think this is the wrong component to use here since `RenderUrlField` adds `type="url"` to the underlying `input` element.

My intention (though not specified anywhere) was that we'd have different `RenderXYZField` components that could be shared among similar custom Field components. Like `RenderUrlField` is used in both `HttpsUrlField` and `GitHubRepoUrlField`.

So, maybe you could either make a new `RenderTextField` component that has `type="text"` and other underlying attributes set appropriately, OR further abstract `RenderUrlField` to something like `RenderInputField` and move the `type` and `autoComplete` up a level.

Hopefully this all makes sense. I'm not that well-versed in redux-form and it has a somewhat bizarre API.

jseppi:
I agree it is more like a check they didn't type something incorrectly. Though we should probably have similar validation on the back-end as well so that malicious parties can't just spoof the form and get around the front-end validations (separate PR is fine for that).

test/frontend/components/Fields/BranchField.test.jsx
el-mapache: What is the current behavior if the user adds a repo that has a branch named using that pattern? If the example on line 12 is a valid branch name it's possible a user will add a repo and try to build a branch named something similar right?

I guess my question is: is this actually a problem with the validator

frontend/components/Fields/BranchField.js
el-mapache: I think the best strategy is actually to rename this to `FieldWithError` or something similarly descriptive, since that is what it's actual behavior seems to be.

The `type` of the field can just be controlled with props, and since we need to wrap these fields in an additional component to handle validations, I don't think we need to necessarily have multiple components that differ only by their `type`

el-mapache:
The redux-form API is not the best, I agree 😞

frontend/components/Fields/BranchField.js
jseppi: Yeah, that sounds good to me!

test/frontend/components/Fields/BranchField.test.jsx
jseppi: I think everything is allowed currently, as long as the site creation process succeeds. We run into problems later, like the inability to create viewable branch preview links (for which https://github.com/18F/federalist/issues/1018 was implemented).

It's not a problem with the validator per se, but could help prevent other later problems if the allowed branch name characters are further restricted.

el-mapache: I suppose my ultimate point is that we need to address it more fully in a follow up PR, because in the case described here, we should probably have an additional message letting a user know why we can't build a site using a branch named in this way

jseppi:
Created https://github.com/18F/federalist/issues/1509 for the back end portion of this

frontend/util/validators.js
```diff
@@ -0,0 +1,3 @@
+const validBranchName = /^[a-zA-Z0-9._-]+$/;
+
+export { validBranchName };
```
jseppi: Might want to just go ahead and make the exported value a function instead of a regex since it is used that way in both places. ie, `validBranchName = s => validBranchNameRegex.test(s)`

PR #1510, opened 2018-01-17
Fix users being unable to rebuild any build

el-mapache:
* Add a new build error to differentiate between a build not being found
and a branch not being found on github

* Adds build resolver to move finding branches out of an authorizer

* Properly use siteAuthorizer and return 403s when user doesnt have
access to a site

* A build doesn't require authorization provided a user has access to a
site, so remove those checks


api/services/BuildResolver.js
```diff
@@ -0,0 +1,73 @@
...
+    attributes: ['id', 'owner', 'repository'],
+    include: {
+      model: Build,
+      where: { branch },
+      required: false,
+    },
+  })
+  .then((model) => {
+    const site = model;
+    const build = site.Builds && site.Builds[0];
```
jseppi: Is the first `Build` guaranteed to be the one we want?

api/services/BuildResolver.js
```diff
@@ -0,0 +1,73 @@
...
+const buildErrors = require('../responses/buildErrors');
+const GitHub = require('../services/GitHub');
+
+const rejectBuild = msg =>
+  Promise.reject({
+    status: 404,
+    message: msg,
+  });
+
+const getBranch = ({ user, site, branch }) =>
```
jseppi: maybe rename to `getBranchFromGitHub` for a little extra clarity, especially where this is used down below

api/services/BuildResolver.js
jseppi: Like maybe in the `include` config they should be ordered descending by `createdAt` to make sure we're getting the latest one at the `[0]` element?

api/services/BuildResolver.js
```diff
@@ -0,0 +1,73 @@
...
+      required: false,
+    },
+  })
+  .then((model) => {
+    const site = model;
+    const build = site.Builds && site.Builds[0];
+
+    // The branch we want to create a new build from has been built via federalist before
+    if (build) {
+      return Object.assign({}, build.toJSON(), { commitSha: sha });
```
jseppi: the `build` should/can already have a `commitSha`, right? Do we really want to overwrite it here if it does?

api/services/BuildResolver.js
jseppi: You might also be able to add `limit: 1` to the the `include` config for a slightly speedier database query.

api/services/BuildResolver.js
el-mapache: It isn't guaranteed to have one, no. I have a few builds on staging at least that were completed successfully but don't have SHAs for whatever reason. We can not overwrite it if present but we do need to account for it.

If this isn't an issue in the prod database I'm happy to remove it.

api/services/BuildResolver.js
el-mapache: No you're right, when I shuffled things around from relying on the user, I didn't account for the fact that we could now get more than one build back.

api/services/BuildResolver.js
jseppi: It is probably the case that some builds in prod are missing it as well since that field was added in a migration in 2017.

Maybe just add it if it isn't there already?

```js
const serializedBuild = build.toJSON();
serializedBuild.commitSha = serializedBuild.commitSha || sha;
return serializedBuild;
```

el-mapache:
thank _you_ for reviewing :bowtie:

PR #1513, opened 2018-01-17
Adds model validation to build

el-mapache:
* Invalid branch names will no longer persist to build records

### DIFF
--- a/test/api/unit/models/build.test.js
+++ b/test/api/unit/models/build.test.js
@@ -1,4 +1,5 @@
 const { expect } = require('chai');
+const { stub } = require('sinon');
 const SQS = require('../../../../api/services/SQS');
 const factory = require('../../support/factory');
 const { Build, Site } = require('../../../../api/models');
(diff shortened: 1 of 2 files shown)


test/api/unit/models/build.test.js
```diff
@@ -1,160 +1,183 @@
...
-      const sitePromise = factory.site()
+        expect(build.token).to.equal('123abc');
+        done();
+      })
+      .catch(done);
+    });
+  });
+
+  describe('after create hook', () => {
+    // TODO: This doesnt test anything
```
jseppi: ha, weird o_O

Looks like `SQS.sendBuildMessage` should be called with the newly created `buildId`. Want to add a quick expectation to check that?

test/api/unit/models/build.test.js
```diff
@@ -1,160 +1,183 @@
...
+    it('should send a build new build message', (done) => {
+      const oldSendMessage = SQS.sqsClient.sendMessage;
+
+      SQS.sqsClient.sendMessage = () => {
+        SQS.sqsClient.sendMessage = oldSendMessage;
```
jseppi: this behavior of restoring the old method should probably take place in `before()` and `after()` blocks

jseppi: I know this wasn't part of this PR's change, but looks easy enough to fix this up and make it meaningful if you have time.

el-mapache:
Added the expectation!

test/api/unit/models/build.test.js
```diff
@@ -1,160 +1,194 @@
...
+
+    beforeEach(() => {
+      sendMessageStub = stub(SQS, 'sendBuildMessage');
+    });
+
+    afterEach(() => {
+      sendMessageStub.restore();
+    });
+
+    it('should send a build new build message', (done) => {
```
jseppi: 👍 thanks!

PR #1514, opened 2018-01-17
Validate branch names on the settings page

el-mapache:
Screenshot:

screen shot 2018-01-17 at 3 06 48 pm

### DIFF
--- a/frontend/components/site/SiteSettings/BasicSiteSettings.jsx
+++ b/frontend/components/site/SiteSettings/BasicSiteSettings.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Field, reduxForm } from 'redux-form';
+import { reduxForm } from 'redux-form';
 
 import HttpsUrlField from '../../Fields/HttpsUrlField';
+import BranchField from '../../Fields/BranchField';
 
 const propTypes = {
   // initialValues is what the initial form values are based on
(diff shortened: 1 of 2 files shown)


frontend/components/site/SiteSettings/BasicSiteSettings.jsx
```diff
@@ -45,15 +46,13 @@ export const BasicSiteSettings = ({
...
-          Branch name:
-        </label>
-        <Field
-          component="input"
+        <BranchField
+          label="Branch name:"
           type="text"
           id="defaultBranchInput"
           name="defaultBranch"
           className="form-control"
```
jseppi: What do you think about either setting a blank placeholder on the demoBranch field or adding a non-blank one here?

el-mapache: My feeling here is that its never going to be blank in the above example, I can't really think of any non-broken page state where the site isnt going to have a primary branch. We can duplicate the demoBranch value if it seems safer.

jseppi: Yeah, that's a fair point. I don't feel strongly about it.

el-mapache: I ended up changing, it is very simple, like you said, and it doesn't hurt anything 😄

wslack:
woot!

PR #1522, opened 2018-01-19
Find ups by instance_name

jseppi:
Fixes for case where the user-provided service credentials we need are not the first in the array. Rather than grabbing `[0]`, now we look for the one with an `instance_name` of `federalist-staging-env`

### DIFF
--- a/scripts/update-local-config.js
+++ b/scripts/update-local-config.js
@@ -4,7 +4,9 @@ const fs = require('fs');
 const app = 'federalist-staging';
 const space = 'staging';
 const org = 'gsa-18f-federalist';
+const credsUpsName = 'federalist-staging-env';
 const localConfigPath = './config/local-from-staging.js';
+
 
 function showUsage({ alreadyLoggedIn = true } = {}) {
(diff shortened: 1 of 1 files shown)


scripts/update-local-config.js
```diff
@@ -23,9 +25,15 @@ function showUsage({ alreadyLoggedIn = true } = {}) {
...
 }
 
-function updatedLocalConfig(vcapJson) {
+function getUpsCreds(vcapJson) {
+  return vcapJson.VCAP_SERVICES['user-provided']
+    .find(el => el.instance_name === credsUpsName)
+    .credentials;
+}
+
+function updateLocalConfig(vcapJson) {
```
el-mapache: Is it safe to assume that the credentials object will also always be at this index?

jseppi: Yeah, all of the entries in the `user-provided` array will have a `credentials` property.

I didn't put much error handling in this code at all, because I figured if it broke, it wasn't a big deal since it just impacts a local dev environment, and probably indicates something else weird going on.

PR #1527, opened 2018-01-19
Deploy to prod after 1/19/18 review

jseppi:
Should we wait until Monday, @wslack?


wslack:
I don't think we need to wait! There's time today to troubleshoot if anything goes awry.

-Will

PR #1528, opened 2018-01-19
Add branch validations to site

el-mapache:


api/models/build.js
```diff
@@ -1,6 +1,7 @@
 const crypto = require('crypto');
 const URLSafeBase64 = require('urlsafe-base64');
 const SQS = require('../services/SQS');
+const branchRegex = /^[a-zA-Z0-9._-]+$/;
```
jseppi: could we centralize this somewhere so that it is not repeated here and in `site.js`?

jseppi: maybe even just by exporting the one defined in `models/site.js` and importing it here.

el-mapache: yeah I actually tried exporting from the front end validations, but importing into node causes issues since it doesn't recognize the `exports` keyword. Should we just export from the `site` model as you suggest, or make a new folder/file where we can keep our regexes?

jseppi: Oh right, it won't be that easy to export from `site.js`.

Totally happy with a new file to keep validators, like we have for the front end. Would be cool to eventually share that code between front and back ends, but that can be done later.

el-mapache:
Ok, for now I'll add a separate file to import these from!

PR #1534, opened 2018-01-23
Adds error msg if org hasn't authorized federalist

el-mapache:
* When a user attempts to add a site from an exisiting repo,
differentiate between the user not having admin access to that repo
and federalist access not approved for that repo's organization

### DIFF
--- a/api/services/SiteCreator.js
+++ b/api/services/SiteCreator.js
@@ -48,7 +48,11 @@ function ownerIsFederalistUser(owner) {
 function checkGithubOrg({ user, owner }) {
   return ownerIsFederalistUser(owner)
   .then((model) => {
-    if (model) return Promise.resolve();
+    if (model) {
+      // Owner of the repo is a user, not an org.
+      // They exist in our DB, drop down to the next promise and bypass
+      return Promise.resolve(true);
+    }
 
     return GitHub.checkOrganizations(user, owner);
   })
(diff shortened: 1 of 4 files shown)


api/services/SiteCreator.js
```diff
@@ -38,6 +38,34 @@ function paramsForNewBuild({ user, site, template }) {
...
+      // They exist in our DB, drop down to the next promise and bypass
+      return Promise.resolve(true);
+    }
+
+    return GitHub.checkOrganizations(user, owner);
+  })
+  .then((existingOrg) => {
+    if (!existingOrg) {
+      throw {
+        message: `Organization '${owner}'' hasn't approved access for federalist. Ask an owner to authorize it`,
```
jseppi: Few editorial suggestions:
- accidental double single quote: `'${owner}''`
- capitalize `Federalist`
- add `.` after `it` in second sentence

api/services/SiteCreator.js
```diff
@@ -38,6 +38,34 @@ function paramsForNewBuild({ user, site, template }) {
...
+    if (model) {
+      // Owner of the repo is a user, not an org.
+      // They exist in our DB, drop down to the next promise and bypass
+      return Promise.resolve(true);
+    }
+
+    return GitHub.checkOrganizations(user, owner);
+  })
+  .then((existingOrg) => {
+    if (!existingOrg) {
```
jseppi: This is essentially a boolean value, right? Maybe call it something like `hasOrgAccess` or something for future clarity.

PR #1537, opened 2018-01-23
simplify issue template

wslack:


ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,15 @@
...
--------
 _After evaluating, edit this part:_
 
-### Level of effort
+##### Level of effort - (low/medium/high)
 
-TODO: e.g., low/medium/high
-
-### Implementation outline (if higher than a "low" effort):
+### Implementation outline (if higher than "low" effort):
```
jseppi: should this be level 4 as well? ie `####`

wslack: My mistake! yes, will fix

ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,15 @@
 #### Description of feature or bug
 
-TODO
+<TODO>
```
jseppi: putting the angle brackets around these makes them disappear when the markdown is rendered since they're interpreted as a DOM element. Is that what you were going for?

wslack: If they show up in the "new issue" screen but not in the final published screen, that seems idea.

wslack: that way they don't take up space unless filled in

Issue #1386, opened 2017-12-18
When creating new site from a template, always use `master` for the destination repo branch


line47:
When adding a new site always create `master` branch no matter what the branch name is of the site you are creating from. For example, if I am creating a site from a branch named `awesome-branch` it will create my new site under the branch name `master` and not `awesome-branch`.

jseppi:
This is for when creating a new site from a template only, right?

line47:
Yep that's correct 👍

jseppi:
I think this should only require a change to the build agent https://github.com/18f/federalist-garden-build

jseppi:
The template repo can have any number of branches. The branch from it to use will be specified in the template config, as it currently is.

The change is that when a template repo is cloned and a new destination repo is made from that template, we will always call the destination branch on the new repo `master`. Currently it uses the same branch name as specified in the template config.

jseppi:
Basically this is the change:

Current behavior:
`18f/template-repo/whatever-branch` -> `new-owner/repo-name/whatever-branch`

Suggested behavior:
`18f/template-repo/whatever-branch` -> `new-owner/repo-name/master`

wslack:
I get that. My point is that our template repos will not be able to have a `master` branch since that would prevent the nameswap during cloning.

jseppi:
No, that's not the case. There is no swapping going on. The template repos are not modified, only cloned. The destination repo can have a branch of any name pushed to it.

jseppi:
I think we'd have to essentially just change this line https://github.com/18F/federalist-garden-build/blob/staging/tasks/clone.py#L84

to 

```py
ctx.run(f'git push {remote_name} {branch}:master')
```

wslack:
I'm so confused. If you clone a repo with three branches: `master`, `new-feature`, and `legacy` - and then you choose to rename `new-feature` to `master`, doesn't that cause a problem?

jseppi:
There is no renaming. The way creating a new site from a template works is: 

1. create a new repository in GitHub called`target-owner/target-name`
1. clone the source template repo to the build container filesystem
1. `git push` the specified `whatever-branch` from the cloned template repo to the target repo in GitHub (`target-owner/target-name`).

We currently push the same branch name that was cloned (`whatever-branch` in this example). We can easily change that to push that same branch (`whatever-branch` but with a different branch name on the target repo (`master`).

wslack:
[...]

PR #1544, opened 2018-01-24
Adding redirect template

line47:
Live preview of redirect branch - https://federalist-proxy.app.cloud.gov/preview/18f/federalist-uswds-template/redirect-config/

### Screenshot

screen shot 2018-01-24 at 5 18 27 pm


config/templates.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     repo: 'federalist-report-template',
     branch: 'master',
     example: 'https://federalist-report-template.18f.gov',
-    title: 'Basic Report Template',
+    title: 'Basic report template',
```
line47: Also sentence cased the template titles

wslack:
should we change the grid to a 3 max to avoid the orphan effect?

line47:
I was thinking something similar @wslack. 
I have also been wondering about stacking the template offerings like the rough screenshot below.

### Stacked template list
screen shot 2018-01-24 at 8 20 52 pm

line47:
Hey @jseppi one thing I noticed when I was creating a new site locally and using the `redirect-config` is was not creating the new site on a `master` branch but rather keeping the `redirect-config` branch name. I thought this was fixed with issue https://github.com/18F/federalist/issues/1386

jseppi:
Huh, yeah, that seems to be a bug in the web app. I just confirmed and will open an issue

jseppi:
Opened https://github.com/18F/federalist/issues/1547

jseppi:
Using a non-master branch for a template is blocked by the issue described in #1547.

jseppi:
This should be unblocked now (thanks, @el-mapache).
@line47 - want to give it a try on your local instance after you merge `staging` into this PR?

line47:
I confirmed that using the redirect template works locally 👍 

**Screenshots**
screen shot 2018-01-31 at 10 26 00 am
screen shot 2018-01-31 at 10 26 09 am

line47:
We still need to address how adding a 5th template creates an orphan on a new line
@ericronne have thoughts on that? (see screenshots earlier in the thread)


CC @wslack

ericronne:
Let me take a look @line47  …

ericronne:
This might work better for now …
templates-tweak

ericronne:
Note two text changes, recommending …
- Remove "template" from the template titles (since we already have that in the section title and the button)
- Change label to "View sample" instead of "View template"

line47:
I like these ideas @ericronne 👍

wslack:
This dramatically raises the vertical size of the page; we'd likely need something at the top to make clear of the "add your own repo" option (or flip those sections).

ericronne:
@wslack We could flop the order and fold the git form into an accordion (and shrink the template thumbs, for overall "digestibility"), so both options are clearly presented to users …

templates-tweak-v2

jseppi:
That looks great

wslack:
I like the recent screenshot, but I wonder if we can make the thumbs a little larger, give them a border, and provider a little more prose for each ( @davidewarren can probably help with the prose).

I also think the "add your own" call to action is small enough that we shouldn't make folks unfurl it.

line47:
I agree in showing the full "add your own GitHub repository" form and think it could be even simplified a bit more.

I would hesitate in making the thumbs larger if we want to try and control the length of this page also we have a preview link right there where they can see the a live demo.

I think adding a bit more descriptive language to each template would be great too!

wslack:
I would be happy to make the thumbs smaller if we also remove the gray box - I just didn't like the vertical space being wasted.

jseppi:
Are we considering this blocked by the redesign of the page? If so, let's create an issue for that design implementation and mark it as a dependency to this one.

jseppi:
Sorry, I didn't just mean the actual design work, but also the implementation of that design. This seems to have drifted fairly far from just adding a new template, so it might be good to track that work separately.

ericronne:
i defer to your wisdom 🙏 

revision...

templates-tweak-v3

jseppi:
Filed https://github.com/18F/federalist/issues/1585

line47:
Just a heads up, I'm going to open a new PR for the UI work on the add new site page

PR #1556, opened 2018-01-26
Always specify branch 'master' when creating a new site from a template

el-mapache:


api/services/TemplateResolver.js
```diff
@@ -0,0 +1,15 @@
+const templates = require('../../config/templates');
```
el-mapache: This is kind of overkill but it makes testing easier, and it doesn't rely on us having templates in the`config/templates` file that match our edge cases.

wslack:
Why can't we clone the repo and then trigger the build from `master` in the cloned repo? That seems simpler.

el-mapache:
@wslack I think the point is that we don't want `master` as the base branch, we want whatever branch is specified in the template configs. I'm pretty sure the only change I need to make here is on line 28 of `SiteCreator`, using template.branch instead of `site.defaultBranch`

wslack:
Ohh I get it (I think) sounds like the issue is that we were pulling from the wrong field

So now it will be:

Source Template X specifies branch `Y`

Cloned Template Z will have branch `master` which is a clone of `Y`

jseppi:
Looks like a little more needs to happen. I've been testing locally by changing my local `templates.js`'s `uswds` config to reference that repo's `report-config` branch:

```js
  // ...
  uswds: {
    owner: '18f',
    repo: 'federalist-uswds-template',
    branch: 'report-config', // <--- changed this to use non-master branch
    example: 'https://federalist-uswds-template.18f.gov',
    title: 'REPORT CONFIG',
    description: 'A flexible template using the U.S. Web Design Standards.',
    thumb: '/images/federalist-uswds-template.jpg',
  },
```

Then I use the Add Site page to create a new site from that template. The site is created, and its default branch is properly set to`master` in the settings page. However, the first build still says it is happening for the `report-config` branch:

screen shot 2018-01-29 at 4 01 22 pm

The build is scheduled and the build goes through, but the environment variable params (`SITE_PREFIX` and `BASEURL`) sent to the builder and then to the build container make the build container think it is a `preview` site, so it gets published to the wrong S3 path (ie `/preview/jseppi/test-report-config/test-report-config` instead of `/site/jseppi/test-report-config` :

```sh
BASEURL: /preview/jseppi/test-report-config/report-config
BRANCH: report-config
BUCKET: <omitted>
CACHE_CONTROL: max-age=60
CONFIG: <nil>
FEDERALIST_BUILDER_CALLBACK: <omitted>
GENERATOR: jekyll
GITHUB_TOKEN: <omitted>
LOG_CALLBACK: <omitted>
OWNER: jseppi
REPOSITORY: test-report-config
SITE_PREFIX: preview/jseppi/test-report-config/report-config
SKIP_LOGGING: true
SOURCE_OWNER: 18f
SOURCE_REPO: federalist-uswds-template
```

el-mapache:
So, this is going to require more substantial changes?

Currently we kick off a build after creating the record in the DB, using the that data from the record.

But, what we actually want to do is send a message to SQS using the branch specified in the template branch name, not the one in the DB record (which should just be `master`).

Is this right?

wslack:
Is there a benefit to triggering the build from the source repo instead of waiting to run it from the cloned repo?

el-mapache:
I think one issue is that if something goes wrong with the clone we'd have an orphan build/site

jseppi:
The latter would probably require more changes. 

We don't clone repos via GitHub. Rather, an empty destination repo with the is created by the web application using the info the user provided (owner and repo name), then the build container clones the source (template) repo and pushes the configured branch from that repo to the `master` branch of the destination repo.

el-mapache:
To be clear, I was just using clone to indicate that we were making a copy of repo, not that we were calling `git clone`, I should have been more specific.

wslack:
Oh, that's me not understanding how the clone operation works. NVM.

el-mapache:
Yeah my thinking was if something goes wrong and we don't have an empty repository to put the template repo contents into.

el-mapache:
@jseppi I think this time the code should **really** be good, here is a sample of the output from the builder on staging:

```
SOURCE_REPO=federalist-modern-team-template
BASEURL=/site/el-mapache/so-great
BUCKET=<private!>
GENERATOR=jekyll
BRANCH=simplfy-building-team
```

el-mapache:
thanks for your reviews and for talking it through! 🎉

PR #1557, opened 2018-01-29
Remove StoreState, use connect HoC in SiteContainer children

el-mapache:
* Refactoring components + tests to use connect middleware to get data
deps
* For now, still passing the site down, as each component would need to
implement logic to derive current site from list of sites, or import
some helper code
* May be worth investigating moving pieces of router state to redux and
creating an additional reducer to get the current site


el-mapache:
The `SiteContainer` component is still manually passing the site down to its child component for now; changing that will probably require more code changes that I think are outside the scope of this PR.

jseppi:
I like this idea:

> May be worth investigating moving pieces of router state to redux and
creating an additional reducer to get the current site

Want to file an issue about it?

davidewarren:
Discussed today in standup - merging this in now so that we have more time to test in staging

jseppi:
Filed #1587 and #1588 for outstanding items

PR #1572, opened 2018-01-30
Adding plus icon to add website button. Removing unused images

line47:
### Screenshot
screen shot 2018-01-30 at 2 34 57 pm

CC @ericronne


ericronne:
I updated the svg with a shorter one. Can you see how that looks? I don't know how to adjust svgs, but ideally it would be centered vertically in the button. thx!

line47:
**Original**
screen shot 2018-01-30 at 2 34 57 pm


**Updated**
screen shot 2018-01-30 at 3 37 56 pm

line47:
Success! I have made Code Climate Happy! 👨‍💻

frontend/components/siteList/siteList.js
```diff
@@ -48,9 +48,33 @@ const addWebsiteButton = () => (
   <Link
     to="/sites/new"
     role="button"
-    className="usa-button"
+    className="usa-button button-add-website"
     alt="Add a new website"
   >
+    <svg width="12px" height="12px" viewBox="0 0 12 12" version="1.1">
```
jseppi: any way this could use the `icon-plus.svg` (like via an `img` tag or `background-img` rule) instead of duplicating its content here?

I always forget the best ways to use SVGs...I know it kinda depends on the situation.

If neither of those options works, I think we can actually use `import` to load the file. Let me know and I can help out doing it that way.

line47: Yeah I'd rather use it as a bg image - I'll see if I can work that out 👍

line47:
@ericronne if you want to tinker with the placement a bit it's here https://github.com/18F/federalist/blob/sw-visual-mods/frontend/sass/_actions.scss#L25

PR #1573, opened 2018-01-30
Use AlertBanner component over hand-coded alerts in component

el-mapache:
There were a few remaining components that manually built alert banners out of divs. This PR replaces those with `AlertBanner` components and updates the appropriate specs.

### DIFF
--- a/frontend/components/alertBanner.js
+++ b/frontend/components/alertBanner.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const propTypes = {
-  message: PropTypes.string,
+  message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   status: PropTypes.string,
   header: PropTypes.string,
   children: PropTypes.node,
--- a/frontend/components/alertBanner.js
+++ b/frontend/components/alertBanner.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const propTypes = {
-  message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  message: PropTypes.node,
   status: PropTypes.string,
   header: PropTypes.string,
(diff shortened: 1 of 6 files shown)


frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
```diff
@@ -116,19 +117,15 @@ export const AdvancedSiteSettings = ({
...
-        <p className="usa-alert-text">Delete this site from Federalist?</p>
-        <button
-          className="usa-button usa-button-red"
-          onClick={onDelete}
-        >
-          Delete
-        </button>
-      </div>
-    </div>
+    <AlertBanner
```
jseppi: I think should make a boolean prop on `AlertBanner` for whether or not we want `role="alert"` to render.

For this "Danger zone" usage, for instance, I think we don't want that `role` present since it's not an alert being displayed in response to user action.

Here's some more reading the alert role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

el-mapache: Seems reasonable! I think having it default to true (and be opt-out) is fine since the vast majority of our alerts are ones that we want people to know about when they happen

jseppi: agreed 👍

frontend/components/alertBanner.js
```diff
@@ -26,15 +40,18 @@ const renderHeader = (text) => {
...
+const AlertBanner = ({ children, header, message, status, alertRole }) => {
   if (!message) {
     return null;
   }
 
   return (
     <div
       className={`usa-alert usa-alert-${status}`}
-      role="alert"
+      {...addRole(alertRole)}
```
jseppi: cool!

PR #1575, opened 2018-01-30
Changing add website button to return to top link

line47:
### Screenshot
screen shot 2018-01-30 at 4 22 56 pm

### DIFF
--- a/frontend/components/siteList/siteList.js
+++ b/frontend/components/siteList/siteList.js
@@ -72,13 +72,8 @@ export const SiteList = ({ storeState }) =>
     </div>
 
     <AlertBanner {...storeState.alert} />
-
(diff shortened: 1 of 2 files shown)


frontend/components/siteList/siteList.js
```diff
@@ -74,7 +74,11 @@ export const SiteList = ({ storeState }) =>
     <AlertBanner {...storeState.alert} />
 
     {getSites(storeState.sites)}
-    {addWebsiteButton()}
+    <Link
```
jseppi: When I tried this locally, it didn't end up moving my scroll position up to the top.

Using a regular ole `a` tag did work, however: 

```html
<a href="#top">Return to top</a>
```

PR #1580, opened 2018-02-01
prose edits

wslack:


wslack:
I have to escape "this?"

frontend/components/site/SiteSettings/CopyRepoForm.jsx
```diff
@@ -23,18 +23,19 @@ class CopyRepoForm extends React.Component {
   renderDisclaimer() {
     return (
       <p>
-        This feature will allow you to copy this site to a new GitHub repository.
-        You must be an admin user of this site.
+        This feature will allow you to copy this site's code to a new GitHub repository.
```
jseppi: `'` in `site's` should be `&apos;`: `site&apos;s`

wslack: ohhh that makes more sense

PR #1596, opened 2018-02-06
Update templates.js

davidewarren:
fix template wording


ericronne:
multi-page sb hyphenated, i think

davidewarren:
@ericronne thanks! you are right. It should have a hyphen

PR #1598, opened 2018-02-07
Use react-svg-loader to load svg icons as components

jseppi:
This PR changes our SVG approach from using `background-image` swaps for active styling to directly embedding `<svg>` elements as React components, which can then be styled with CSS.

- To handle using webpack loading only our icon SVGs as React components, I moved them all into a subdirectory called `public/assets/images/icons/`, and adjusted our webpack config accordingly.
- Stripped out the inline styling rules (`fill` and `stroke` primarily) from all the icon SVG files since we should instead control those options with CSS.
- Simplified our GitHubLink/GitHubRepoLink/GitHubURLProvider code to just a single component, `GitHubLink`. It always includes the GitHub icon because in practice we were always adding that icon.

To Do:
- [x] Make sure all icons that are in use have been ported to new method.
  - [x] "View site" arrow icon
  - [x] Others?
- [x] Figure out how to get `mocha` to play well with the `import`ed SVGs.


jseppi:
Unfortunately I hit the time I had allotted to work on this before getting the tests to work, so I wanted to park it until we could discuss (cc @wslack and @el-mapache).

jseppi:
Other options that might fix the tests:

- Directly include the SVG icons in a JS file instead of `import`ing them. This would make it harder for modifications to the SVGs themselves.
- Perhaps use `mocha-webpack` or `jest` for running tests. That's probably a pretty hefty lift.

el-mapache:
@jseppi I think you can just use proxyquire to stub out the component with an empty div, e.g. `stub = () => <div></div>`. The test will still try to find a component named `IconGithub` but will accept the empty div that the test supplies.

As an aside, maybe its time to just make a single component that includes both the github icon and child elements? I think we have standardized on always having the icon accompanied by text, and currently we have to update every single component with this new code (and we'll have to keep doing it if we change import methods or something again)

jseppi:
Oh, `proxyquire` is a good idea! Thanks @el-mapache!

I agree with your suggestion for the GitHub link component, too.

jseppi:
All ready for review, @el-mapache. This touched a lot of stuff, so it would be great if you could exercise it locally a bunch.

jseppi:
I'll work on figuring out that drop in coverage too

el-mapache:
Taking a peek now!

jseppi:
Bleh, my guess for why coverage declined (by 0.1%) was because the overall amount of code declined due to the GitHubLink refactoring. All the new code here in this PR is 100% covered though.

jseppi:
Thanks -- will check that out to see if I can improve!

el-mapache:
If you can just force it through thats fine, I'm pretty unconcerned with a 0.1% drop. We will probably have to go through and have a PR with backfill tests to avoid this in the future, though.

jseppi:
I tried the `crispedges`, but it actually looked worse to me (on Chrome, at least).

I definitely see the small cutoff edges on the icons, and now I can't stop seeing them. I'm not totally sure how to best fix. It seems to be something to do with the `viewBox` on those icons. Maybe @ericronne and I can work on it (in a separate issue/PR). What do you think, @el-mapache?

el-mapache:
Interesting, I thought crispedges looked better than the current rendering on Chrome!

Separate issue is fine, the cutoff doesn't affect the functionality at all. It was just bugging me.  And now, I suppose, it's bugging you 😀

jseppi:
I added a minor test for the UserActionsTable to bring up the coverage.

PR #1611, opened 2018-02-09
update error message for edge case

wslack:


wslack:
I predict this fails code climate

wslack:
I miscoded something and will need help Monday.

wslack:
I figured it out! 🎉

wslack:
Oh I do like those more as well.

wslack:
@jseppi I think you had a typo in there; check my fixes?

api/services/SiteCreator.js
```diff
@@ -70,8 +70,8 @@ function checkGithubOrg({ user, owner }) {
     if (!federalistAuthorizedOrg) {
       throw {
         message: `Federalist can't confirm org permissions for '${owner}'. ` +
-          `Either '${owner}' hasn't approved access for Federalist or you aren't an org member. ` +
-          'Ensure you are an org member and ask an org owner to authorize Federalist for the organization.',
+        `Either '${owner}' hasn't approved access for Federalist or you aren't an org member. ` +
+        `Ensure you are an org member and ask an org owner to authorize Federalist for the organization.`,
```
jseppi: I think this will fail linting. Backticks are only used when there is an interpolated value (demarcated with `${...}`) in the string .

wslack: Ugh. That seems like weird styling but ok.

jseppi: It's not only styling, but kind of a check to make sure you are using the correct string delimiter type as appropriate. I do get that it can look kinda weird when they're right next to each other.

PR #1616, opened 2018-02-12
Restore call to fetchPublishedBranches

el-mapache:
* Now the published branches will populate correctly for sites


test/frontend/components/site/SitePublishedBranchesTable.test.jsx
```diff
@@ -46,4 +59,14 @@ describe('<SitePublishedBranchesTable/>', () => {
       'Please wait for build to complete or check logs for error message.'
     );
   });
+
+  it('fetches published branches on mount', () => {
```
jseppi: 👍

Issue #1330, opened 2017-12-04
Add YAML validation to site configuration textareas


jseppi:
Adding validation would elimination a whole class of potential errors during build, and would shorten the feedback loop for site builds (ie, you wouldn't have to wait until the jekyll step to see that you screwed up your YAML).

wslack:
Moving to backlog subject to further user feedback. 

Users haven't set text config w/o the skill set to address issues, and while this would be helpful, I'm unsure of the impact.

PR #1631, opened 2018-02-15
add YAML validation for site configs

jseppi:
Adds server-side validation to make sure submitted `config`, `demoConfig`, and `previewConfig` values are valid YAML.

Client-side would be nice, too, but I figured including `js-yaml` in the client-side bundle was a little too much just for this one thing.

ref #1330

### DIFF
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -1,7 +1,7 @@
 const validator = require('validator');
 const config = require('../../config');
 
-const { branchRegex } = require('../utils/validators');
+const { branchRegex, isValidYaml } = require('../utils/validators');
 
 const afterValidate = (site) => {
   if (site.defaultBranch === site.demoBranch) {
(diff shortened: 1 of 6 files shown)


api/models/site.js
```diff
@@ -154,9 +157,15 @@ module.exports = (sequelize, DataTypes) => {
...
     previewConfig: {
       type: DataTypes.STRING,
+      validate: {
+        isValidYaml,
+      },
     },
     demoConfig: {
       type: DataTypes.STRING,
+      validate: {
+        isValidYaml,
```
jseppi: empty strings will validate successfully, so no need to add an explicit `isEmpty` check

test/api/requests/site.test.js
```diff
@@ -345,21 +345,21 @@ describe('Site API', () => {
...
-            engine: 'jekyll',
-          })
-          .set('Cookie', cookie)
-          .expect(400)).then((response) => {
-            validateAgainstJSONSchema('POST', '/site', 400, response.body);
-            expect(response.body.message).to.equal('You do not have admin access to this repository');
-            done();
-          })
-          .catch(done);
+        .post('/v0/site')
```
jseppi: just formatting in this section

PR #1633, opened 2018-02-16
Removes unused error prop

el-mapache:
* Error not being provided by a reducer, so it shouldn't be ncessary.
Additionally, we don't surface errors on the AddSite page

* Also fixes key prop warning in TemplateList component, and removes
uneeded grid code

### DIFF
--- a/frontend/components/AddSite/TemplateSiteList/index.js
+++ b/frontend/components/AddSite/TemplateSiteList/index.js
@@ -10,30 +10,7 @@ const propTypes = {
   defaultOwner: PropTypes.string.isRequired,
 };
 
-const MAX_CELLS_PER_ROW = 4;
-
-/**
- * Create a two-dimensional array of values that represent one or more rows
- * of values, with `perRow` values per row.
- *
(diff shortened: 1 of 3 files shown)


frontend/components/AddSite/TemplateSiteList/index.js
```diff
@@ -53,35 +30,24 @@ export class TemplateList extends React.Component {
...
-      </div>
-    ));
+    const templateGrid = Object.keys(templates).map((templateName, index) => {
+      const template = templates[templateName];
+
+      return (
+        <div className="federalist-template-list" key={templateName}>
+          <TemplateSite
+            name={template.title}
+            index={index} // eslint-disable-line no-plusplus
```
jseppi: can get rid of this `// eslint` comment as well

el-mapache: oh yeah duh

el-mapache:
@wslack I don't think this PR is causing any issues.

Issue #1482, opened 2018-01-10
Resolve bizarre prop type validation issue in Users view


jseppi:
- Level of Effort: ?
- Implementation Sketch (if higher than a "low" effort): ?
- Definition of Done: the React prop type validation error message does not show up in the console any longer. This error is only visible in non-prod builds, btw.

For more context, see https://github.com/18F/federalist/pull/1476#issuecomment-356671202

el-mapache:
One thing I think we could do is just remove `connect` entirely and subscribe to the store directly in the component, saving the actions to the component's state.

This means we lose the value of the propTypes type checking, but our tests basically already outline what we expect to get back from the API, so its not a huge deal.

PR #1634, opened 2018-02-20
Fix duplicate PropTypes.shape wrapping in USER_ACTION propType definition

jseppi:
ref #1482

The bizarre warning was actually caused by the `actionTarget` and `initiator` props
being essentially doubled-wrapped in `PropTypes.shape`, since the underlying `USER`
propType was already a `PropType.shape`.

### DIFF
--- a/frontend/components/site/UserActionsTable.jsx
+++ b/frontend/components/site/UserActionsTable.jsx
@@ -5,10 +5,6 @@ import { USER_ACTION } from '../../propTypes';
 import { timestampUTC } from '../../util/datetime';
 import userActions from '../../actions/userActions';
 
-const propTypes = {
-  site: PropTypes.number.isRequired,
-  userActions: PropTypes.arrayOf(USER_ACTION),
-};
 
 class UserActionsTable extends React.Component {
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


frontend/components/site/UserActionsTable.jsx
```diff
@@ -66,7 +62,11 @@ class UserActionsTable extends React.Component {
   }
 }
 
-UserActionsTable.propTypes = propTypes;
+UserActionsTable.propTypes = {
```
el-mapache: I think the general pattern we have been following is to declare all the prop types up at the top of the file, that way its really apparent what input the component accepts.
.
If that seems weird we can start moving them to the bottom, but in general I find the current setup a bit more readable.

jseppi: I agree it would be good to be consistent. This one had part at the top/part at the bottom, so I just moved them both to where I was currently looking (the bottom). I'll file an issue to make these declarations consistent.

jseppi: https://github.com/18F/federalist/issues/1636

Issue #1654, opened 2018-02-26
Error when creating site with any template


line47:
#### Description of feature or bug

When trying to create a site and use any of the templates you get an error that the template doesn't exist.

![screen shot 2018-02-26 at 2 27 01 pm](https://user-images.githubusercontent.com/1449852/36690856-9099d086-1b01-11e8-9f15-2c415fe99c19.png)

#### Definition of done

- [x] when a site can be created using a Federalist template without issue.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

<TODO>

PR #1656, opened 2018-02-26
Fix adding a site from a template

jseppi:
Ref #1654.

In the clean-up from #1633, an incorrect value (the `title` of a template) was passed into the template `name` field.
This was at least partially because `name` was not really an appropriate name for the prop. To help prevent a similar issue in the future, the `name` prop has also been renamed to `templateKey`. The correct value is now passed into the renamed prop.

### DIFF
--- a/api/services/SiteCreator.js
+++ b/api/services/SiteCreator.js
@@ -176,6 +176,7 @@ function createSiteFromSource({ siteParams, user }) {
 
 function createSite({ user, siteParams }) {
   const templateName = siteParams.template;
+
   const template = templateName && TemplateResolver.getTemplate(templateName);
   const newSiteParams = paramsForNewSite(siteParams);
 
--- a/frontend/components/AddSite/TemplateSiteList/index.js
+++ b/frontend/components/AddSite/TemplateSiteList/index.js
@@ -30,13 +30,13 @@ export class TemplateList extends React.Component {
   render() {
     const { templates } = this.props;
 
-    const templateGrid = Object.keys(templates).map((templateName, index) => {
-      const template = templates[templateName];
+    const templateGrid = Object.keys(templates).map((templateKey, index) => {
+      const template = templates[templateKey];
 
       return (
-        <div className="federalist-template-list" key={templateName}>
+        <div className="federalist-template-list" key={templateKey}>
           <TemplateSite
-            name={template.title}
+            templateKey={templateKey}
             index={index}
             thumb={template.thumb}
             active={this.state.activeChildId}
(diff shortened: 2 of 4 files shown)


jseppi:
Another thing I found is that the `TemplateSiteList` and `TemplateSite` components are actually not tested at all. Opening a separate issue for that.

jseppi:
We should hotfix this to prod once approved.

frontend/components/AddSite/index.js
```diff
@@ -12,7 +12,7 @@ import addNewSiteFieldsActions from '../../actions/addNewSiteFieldsActions';
 
 const propTypes = {
   showAddNewSiteFields: PropTypes.bool,
-  user: PropTypes.shape(USER),
+  user: USER,
```
jseppi: fixes the same issue that was corrected for another component in #1634.

api/services/SiteCreator.js
```diff
@@ -176,6 +176,7 @@ function createSiteFromSource({ siteParams, user }) {
 
 function createSite({ user, siteParams }) {
   const templateName = siteParams.template;
+
   const template = templateName && TemplateResolver.getTemplate(templateName);
```
wslack: just to make 100% sure - these variables don't need to be updated, right?

jseppi: no, I just added a newline for readability

frontend/components/AddSite/TemplateSiteList/index.js
```diff
@@ -30,13 +30,13 @@ export class TemplateList extends React.Component {
...
-      const template = templates[templateName];
+    const templateGrid = Object.keys(templates).map((templateKey, index) => {
+      const template = templates[templateKey];
 
       return (
-        <div className="federalist-template-list" key={templateName}>
+        <div className="federalist-template-list" key={templateKey}>
           <TemplateSite
-            name={template.title}
+            templateKey={templateKey}
```
jseppi: this was the relevant change, @wslack. Basically the `title` of the template (like `Report Template`) was incorrectly being passed here.

wslack: right; we shouldn't be pulling this from the template data set. I don't see how you are deriving templateKey unless the key is intended to refer to the entire set of fields about each template.

But we can test that in staging.

jseppi: [line 33](https://github.com/18F/federalist/pull/1656/files/765fb86589f255cd62bf52d29a9d1a9b5011bb01#diff-27d321a5b60a996ab69fc77a401e9c3bR33) shows where `templateKey` comes from.

wslack: cool; I think I get what the code is doing (assigning an identifier to the collection of variables)

PR #1665, opened 2018-03-01
Adds scroll after http error on site settings page

el-mapache:
Site settings page will scroll to the top on error, so the user is aware of it.

**GIF**:
scroll-to-error

I've only added it to this page for now, adding it to all error events means ripping out the error handling logic from the `federalistAPI`, which probably shouldn't be issuing alerts.

Unless people think it should! Open to moving the scroll and handling all errors directly in the api file if that seems cleaner.

### DIFF
--- a/frontend/actions/siteActions.js
+++ b/frontend/actions/siteActions.js
@@ -1,3 +1,5 @@
+/* global window:true */
+
 import federalist from '../util/federalistApi';
 import alertActions from './alertActions';
 
@@ -15,7 +17,10 @@ import {
 } from './dispatchActions';
 import userActions from './userActions';
 
-const alertError = error => alertActions.httpError(error.message);
+const alertError = (error) => {
+  window.scrollTo(0, 0);
+  alertActions.httpError(error.message);
+};
 
 export default {
   fetchSites() {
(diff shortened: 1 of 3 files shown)


wslack:
I defer to James. The other error case is when people can't add a site.

jseppi:
Oh, and to your other point about updating other alerting logic: I think this is fine for now. I suspect there will be larger user experience/visual design changes in the near future that will impact our forms and how want to deal with the alerts in general, so I don't think it's necessarily worth it to spend too much time on it before then.

el-mapache:
@wslack I thought that the user was redirected when adding a site fails? In that case they will always be at the top of the page...but the current code is going to call `window.scrollTo` any time a method in `siteActions` triggers the alert banner.

el-mapache:
I'll add the test, thanks @jseppi

jseppi:
Yup, you're right -- when a site add fails, the user is already at the top of another view.

wslack:
Oh interesting. I don't see that when I test, I have to scroll up.

jseppi:
hrm, are we talking about different things?

error-top

jseppi:
LiceCap

PR #1670, opened 2018-03-05
Er icon changes to sync staging with master

ericronne:


jseppi:
Hrm, I don't think this will work. I think we can just merge `master` into `staging`. I'll open another PR.

wslack:
👍 that makes sense

ericronne:
Cool that was my original thought 😬  godspeed

PR #1655, opened 2018-02-26
Tweaks to github and view-site icons

ericronne:
- Shrunk github and “view site” icons, since they overpowered when
shown multiple times on a page
- Slightly adjusted positioning
- Moved “(you)” into the collaborator link

PR #1671, opened 2018-03-05
Merge `master` into `staging` to sync icon changes

jseppi:
This should get `staging` back in-line with the PR #1655 that was accidentally pulled directly into `master`


wslack:
@jseppi I'll leave it to you to do the "update branch" action but I think that might not be a good idea - that will merge staging into master if memory serves.

jseppi:
No, we should **not** `Update branch` because that will cause a production deploy.

wslack:
yup. let's discuss at standup?

wslack:
Everyone else: do *not* click "update branch"

jseppi:
I can admin-merge this into `staging`.

wslack:
👍

PR #1674, opened 2018-03-05
Don't redirect on add site error

el-mapache:
Errors display directly on the add site screen

### DIFF
--- a/frontend/components/AddSite/index.js
+++ b/frontend/components/AddSite/index.js
@@ -6,21 +6,25 @@ import { connect } from 'react-redux';
 import { USER } from '../../propTypes';
 import TemplateSiteList from './TemplateSiteList';
 import AddRepoSiteForm from './AddRepoSiteForm';
+import AlertBanner from '../alertBanner';
 import { availableEngines } from '../SelectSiteEngine';
 import siteActions from '../../actions/siteActions';
 import addNewSiteFieldsActions from '../../actions/addNewSiteFieldsActions';
 
 const propTypes = {
+  alert: PropTypes.object,
   showAddNewSiteFields: PropTypes.bool,
   user: USER,
 };
 
 const defaultProps = {
+  alert: null,
   showAddNewSiteFields: false,
   user: null,
 };
 
-const mapStateToProps = ({ showAddNewSiteFields, user }) => ({
+const mapStateToProps = ({ alert, showAddNewSiteFields, user }) => ({
+  alert,
   showAddNewSiteFields,
   user,
 });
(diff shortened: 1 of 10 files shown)


frontend/components/AddSite/index.js
```diff
@@ -6,21 +6,25 @@ import { connect } from 'react-redux';
 import { USER } from '../../propTypes';
 import TemplateSiteList from './TemplateSiteList';
 import AddRepoSiteForm from './AddRepoSiteForm';
+import AlertBanner from '../alertBanner';
 import { availableEngines } from '../SelectSiteEngine';
 import siteActions from '../../actions/siteActions';
 import addNewSiteFieldsActions from '../../actions/addNewSiteFieldsActions';
 
 const propTypes = {
+  alert: PropTypes.object,
```
jseppi: Could this be filled out more fully, perhaps as a new `PropType.shape` in our custom `propTypes.js`?

frontend/components/AddSite/index.js
```diff
@@ -88,9 +94,9 @@ export class AddSite extends React.Component {
           </div>
           <div className="usa-content">
             <p>
-              There are a few different ways you can add sites to Federalist.
-              You can start with a brand new site by selecting one of our template sites below.
-              Or you can specify the GitHub repository where your site&#39;s code lives.
+              There are a two different ways you can add sites to Federalist.
```
jseppi: 👍

jseppi:
gif to illustrate what I mean:

repourl

el-mapache:
Normally I would just clear out the field, but a couple of things made me not do that:

* I believe will said that people are primarily copy/pasting the URL into this field
* Its annoying to have to complete the complete URL again, including the `https://`, especially when (IMO) the offending part of the URL is going to be the last part of the path

If those seem stupid to you, I'll change it up

el-mapache:
Good point, I'll look into it.

el-mapache:
@jseppi I made the requested changes! Failing to add a site via the template form does not clear out the form; I'm not sure if the better behavior is to clear those fields out or leave them as is. 

Right now I'm leaning towards preserving the invalid values as the user might want a reference to the username they entered, and I think it's more consistent with our other form errors, although those are also accompanied by inline error messages.

test/frontend/actions/siteActionsTest.js
```diff
@@ -160,7 +171,7 @@ describe('siteActions', () => {
       });
     });
 
-    it('triggers an error and triggers a router update to the site list page when adding a site fails', () => {
+    it('behaves as a no-op qhen a site isnt added, but there is no error', () => {
```
jseppi: Couple typos here

PR #1686, opened 2018-03-12
Home page redesign layout

el-mapache:


ericronne:
I made a number of style adjustments. Here are additional ones which @el-mapache is better qualified to do! Happy to 🍐, as always ✊ thx

**Official banner:**
- [ ] I changed the flag image to us_flag_small.png, which should run at its native size: 16px x 11px (being inexplicably stretched now)
- [ ] Type should be 14px (18 now)
- [ ] Type should be anti-aliased

**Top nav:**
- [ ] Can’t figure out the logo styling; should be 170px wide (height implied)
- [ ] Padding between each nav item should be 30px ($spacing-lg?)
- [ ] Thin down hover underscore to 4px
- [ ] Log in button:
    - [ ] The “button” styles in _actions.scss don’t seem to be having any effect
    - [ ] Text shouldn’t be anti-aliased (since it’s over white)
    - [ ] Button corner radius should be 2px
    - [ ] Font size should be 16px

**Stats box:**
- [ ] Type size should be 19px for all
- [ ] Line height should be 32px for all (see ref image)
- [ ] Reduce padding over the header so that the type is centered between the blue bars
- [ ] We may need to narrow this box a bit, pending final headline content
- [ ] Also pending ^ that ^ is the responsiveness issue

**“Learn more” anchor link:**
- [ ] I added padding to bottom of hero-heading. It’s visually good, but i’m not sure if i used the correct method
- [ ] Underline link in red (remove on hover)
- [ ] Carets should be closer together vertically
- [ ] Add easing to scroll, if we can
- [ ] Add padding above the destination (so it doesn’t hit the top of the browser window)

**More:**
- [ ] Contact button color should match link color, which is now `$color-blue` (#2373a7), with an active color of `$color-blue-dark` (#175c8a)
- [ ] Footer links should be $font-weight-normal
- [ ] H4 is sizing to 1.7rem for some reason, should be 1.5rem (24px) with a line height of 1.5675rem (31px)

el-mapache:
@ericronne Sweet, thanks for updating this!

I just wanted to address a couple of your points / get clarification.

**Top Nav**
Aliasing control (font-smoothing) was removed from the CSS spec, so the support is very limited. I think only Chrome and Safari on OSX support it right now, and that could go away.

**Stats Box**
* Adjusting the line-height for each item, or even just for the `p` elements, doesn't match the reference image, to my eye at least. Here is what 32px renders to in the browser:

screen shot 2018-03-15 at 9 13 08 am

and here is the ref image in Sketch:

screen shot 2018-03-15 at 9 13 48 am

* I probably won't adjust the width of the stats box until we've nailed down the rest of the sizing. Overriding the grid settings for USWDS is a little hairy.

* I'm not sure what you mean by reducing the padding over the header. Do you mean you want the hero copy to live in between the two bars of the callout box? I don't think the ref image reflects that 

In most cases, where there are styles that seem like they should be one size vs another, it's because the specificity of selectors in the USWDS is greater than our specificity, so we need to have explicit rules overwriting them, e.g. for the `h4` font-size, we need to have our own `h4` selector.

ericronne:
Are you sure anti-aliasing is going away?? That would be drastic. The uswds uses it (`-webkit-font-smoothing: antialiased;`) for all instances where white text runs over a dark background. Without it, type looks bad. I use it all the time here.

I think the giant line spacing on the stats box is because margin is being added. If we do a line height of 24px (1.5rem) and add 8px (.5rem) of margin after each item, that will add up to the 32. It'll also ensure that if a line should happen to break (hope not), it will still look okay.

Sorry that my comment about header spacing for the stats box wasn't more clear. The crux is that the bars should appear to be equally distant from the text both above it and below it …

image

el-mapache:
Yep, 100% positive. Also, that prefix indicates that it only works in webkit browsers, which are Safari and older versions of Chrome. Newer versions of Chrome use Blink as their engine, which probably still implement some of the -webkit prefixes, but I'm not certain. Regardless, the property is only implemented on OSX

Here is an MDN link with more info: https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

We can still use it, its just worth knowing that it won't render exactly the same on all browsers and platforms.

thebestsophist:
Updated copy, still needs preview images for downloadable goodies, example sites, and links.

wslack:
Merging to do team review on staging. We will revert this change if there's a hotfix needed before this is ready to go live.

PR #1690, opened 2018-03-14
Adds middleware to redirect from `.fr`

el-mapache:


app.js
```diff
@@ -65,6 +70,16 @@ app.use(methodOverride());
 app.use(flash());
 app.use(responses);
 
+app.use((req, res, next) => {
+  const host = req.get('host');
```
jseppi: Question: would [`req.hostname`](http://expressjs.com/en/api.html#req.hostname) be better instead? I'm a little confused with the difference honestly -- it might just be that `req.hostname` will respect the `X-Forwarded-Host` header.

el-mapache: Yes, as long as we have 'trust proxy' enabled, `hostname` should respect `X-Forwarded-For`. switched!

PR #1694, opened 2018-03-14
Add a highlight to the active link in the side nav

el-mapache:
* The sidenav width is super tiny and breaking the line on the GitHub
Branches link.

**HANDY GIF**:
highlight-link-gif

Let me know your thoughts, this was like 5 minutes of work so its not a big deal if its not what we want to go with

### DIFF
--- a/frontend/sass/_side-nav.scss
+++ b/frontend/sass/_side-nav.scss
@@ -13,7 +13,20 @@
     &:hover svg g {
       fill: $link-color-active;
     }
+
+    &.side-nav-active-item {
+      background-color: #2d476a;
+      color: white !important;
+      padding: 5px !important;
+
+      svg {
+        g {
+          fill: white;
+        }
+      }
+    }
   }
+
 }
 
 .usa-width-one-sixth {
--- a/frontend/components/site/SideNav.jsx
+++ b/frontend/components/site/SideNav.jsx
@@ -13,7 +13,10 @@ const SideNav = ({ siteId }) => (
           const IconComponent = icons[conf.icon];
           return (
             <li key={conf.route}>
-              <Link to={`/sites/${siteId}/${conf.route}`}>
+              <Link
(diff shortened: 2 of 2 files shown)


ericronne:
🤚 may i design? 😬

el-mapache:
yeah design it up!

ericronne:
_onnit!_

ericronne:
Here's one way it could work …
image

Could the nav column could be widened to avoid line breaking (and ensure that there's enough space to the right of each nav item)?

frontend/sass/_side-nav.scss
```diff
@@ -13,7 +13,20 @@
     &:hover svg g {
       fill: $link-color-active;
     }
+
+    &.side-nav-active-item {
+      background-color: #2d476a;
+      color: white !important;
+      padding: 5px !important;
```
line47: Is there a way this can be done without using `!important` ?

ericronne:
@line47 do you mean adopting [this overall style](https://designsystem.digital.gov/components/) to the dashboard? That could be a simple solution.

Regardless, i think the overall design of the dashboard could benefit from a ground-up rethink (rather than piecemeal changes). I wish i had time to address it. :/

line47:
@ericronne not necessarily I was just referring to the left nav in that example link has it's own width and formatting outside of the usa-grid system (which has baked in breakpoints where the grid shifts).

This would give a bit more flexibility for styling the left nav and not having it conform to a grid system.

frontend/sass/_side-nav.scss
el-mapache: Probably, it will just involve a bunch of specificity selectors to override USWDS behavior

el-mapache:
Closing for now as this is going to require a bit more planning to figure out how we want to present the side-nav going forward.

Issue #1662, opened 2018-02-28
Create a metadata image and associated content for better social sharing


ericronne:
#### Description of feature or bug

Currently federalist offers up no preview image or text when shared on social media.
![image.png](https://images.zenhubusercontent.com/555b449c4f7a08fbb7f2de17/a71f8c8d-b106-4614-bb2c-2506d5202f7d)<TODO>

#### Definition of done

Image and text created and implemented

#### Implementation outline:
 - [ ] Review / Edit content
 - [ ] Create image
 - [ ] Implement
 - [ ] Test

jseppi:
FYI: the correct production URL is https://federalist.18f.gov/

ericronne:
👍 
![image](https://user-images.githubusercontent.com/9259185/36805663-9f02ed6e-1c83-11e8-94ce-4edb38774190.png)

jseppi:
That other URL is the old one that we didn't want to break. We have an issue to redirect when a user visits it: #1650

wslack:
close this?

ericronne:
No. We need a sharing image. If we want to wait for the redesign, we can make it a dependency of #1563. Or we can do a placeholder reflective of the current design. Your call

wslack:
I'd say let's use the current logo. I don't expect that to change. That work?

ericronne:
That's easy for now. May want to do something richer down the road

ericronne:
Shooting for something with a modicum of visual interest

![federalist-sharing-twitter](https://user-images.githubusercontent.com/9259185/37303789-8a8547f0-25fd-11e8-8c14-145ad1d52519.png)

ericronne:
cc @wslack

thebestsophist:
How about this for the preview text:

Title (57 characters, max 60): 
> Federalist compliantly hosts federal government websites.

Summary (147 characters, max 200):
> Federalist takes care of the annoying bits of deploying secure, compliant, and scalable government websites so you can focus on serving the public.

jseppi:
How about something like `tedious work` instead of `annoying bits`?

wslack:
How about:

`Federalist expertly manages the backend work of deploying secure, compliant, and scalable government websites, allowing federal offices to focus on their content.`

wslack:
and I 👍 @ericronne's work!

thebestsophist:
Even better, I love it!

wslack:
:shipit:

ericronne:
[...]

PR #1696, opened 2018-03-14
Updates for social sharing

thebestsophist:
Adding social sharing images from @ericronne and twitter cards+facebook open graph descriptions for #1662


views/base.njk
```diff
@@ -42,6 +42,19 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
+    <!-- For Facebook -->
+    <meta property="og:title" content="Federalist compliantly hosts federal government websites." />
+    <meta property="og:type" content="article" />
+    <meta property="og:image" content="/images/social-share/facebook-1200x627.png" />
+    <meta property="og:url" content="https://federalist.18f.gov" />
+    <meta property="og:description" content="Federalist expertly manages the backend work of deploying secure, compliant, and scalable government websites, allowing federal offices to focus on their content." />
+
+    <!-- For Twitter -->
```
jseppi: It looks like several of the the `twitter:` tags will fallback to `og:` counterparts, so maybe we should just specify the ones we want to be different. If I'm reading the [docs](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup) correctly, I think we can leave off `twitter:title` and `twitter:description` since they have the same content in their `og:` counterparts.

docs: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup

thebestsophist: Ah! yes, forgot that it defaults to the OG tags. Removed extra meta tags in 1964bfa

jseppi:
dang, due to our `robots.txt` we can't try this out from `federalist-staging.18f.gov` with [twitter's validator](https://cards-dev.twitter.com/validator)

wslack:
blargh

Should we change federalist-staging.18f.gov's robots.txt briefly? I don't want to confuse search engines.....

PR #1708, opened 2018-03-20
fix build.completeJobErrorMessage has unreached code

amirbey:
condition where err is null or false is unment in Build class completeJobErrorMessage method and is therefore untestable


el-mapache:
So `completeJobErrorMessage` is always called with an error value?

PR #1725, opened 2018-03-22
Add code to transform S3 invalid key error

el-mapache:
* Catch S3 error in dev and test envs to let engineers know they need
to rotate their local keys

* Add test case


test/api/requests/published-branch.test.js
```diff
@@ -118,6 +118,36 @@ describe('Published Branches API', () => {
...
+        cookie: cookiePromise,
+      }).then((promisedValues) => {
+        site = promisedValues.site;
+
+        return request(app)
+          .get(`/v0/site/${site.id}/published-branch`)
+          .set('Cookie', promisedValues.cookie)
+          .expect(400);
+      }).then((response) => {
+        validateAgainstJSONSchema('GET', '/site/{site_id}/published-branch', 400, response.body);
```
jmhooper: We could probably throw and expectation in here to look at the error message.

el-mapache: Good call, we do that in other tests, will add!

Issue #1706, opened 2018-03-20
Images and links for Federalist homepage redesign


el-mapache:
#### Description of feature or bug

We are missing final images / captions / and downloadable files for the Federalist homepage redesign

#### Definition of done

- [x] placeholder images are replaced with real images:
  - [x] images of templates, with descriptive captions
  - [x] images of websites built with Federalist, with descriptive captions
  - [x] images of Federalist partners. This should probably be a static image with all
         the badges, rather than a bunch of individual images
- [x] downloadable content is added, and download links link to downloads

#### Level of effort - <low/medium/high>

**medium**

#### Implementation outline (if higher than "low" effort):

Implementation should be quick, but wrangling of all content will require coordination and might take some time

PR #1728, opened 2018-03-22
adding example sites and partner organizations to new home page

thebestsophist:
For #1706, adding example sites and partner organizations for the new home page. Added 4 example sites to the page, can add more if needed.


el-mapache:
@thebestsophist  Thanks! Do you mind adding a quick screenshot of the changes too?

PR #1730, opened 2018-03-23
Preserve path in rewritten URLs

el-mapache:
Redirect middleware was rewriting all URLs to point to the base federalist site, which may break legacy preview links.

This PR re-appends the path to those rewrites

### DIFF
--- a/app.js
+++ b/app.js
@@ -72,11 +72,10 @@ app.use(responses);
 
 app.use((req, res, next) => {
   const host = req.hostname;
-  const path = req.pathname || '';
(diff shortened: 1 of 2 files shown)


wslack:
Wait a sec - should preview links work under the federalist.18f.gov URL?

el-mapache:
Why shouldn't they?

el-mapache:
This fix preserves whatever the path is, be it `preview` or `random-string`.

el-mapache:
closing as we want this to break

PR #1731, opened 2018-03-23
Updating report config to use new template

line47:
Updating the report config to use the new report-config instead of the separate repo.

**New branch (report-config)**
https://github.com/18F/federalist-uswds-template/tree/report-config

**Old repo** 
https://github.com/18F/federalist-report-template
This repo can be retired once it is determined that the replacement is good to go

### DIFF
--- a/config/templates.js
+++ b/config/templates.js
@@ -19,9 +19,9 @@ module.exports = {
   },
   report: {
     owner: '18f',
-    repo: 'federalist-report-template',
-    branch: 'master',
-    example: 'https://federalist-report-template.18f.gov',
+    repo: 'federalist-uswds-template',
+    branch: 'report-config',
+    example: 'https://federalist-proxy.app.cloud.gov/preview/18f/federalist-uswds-template/report-config/',
     title: 'Basic report',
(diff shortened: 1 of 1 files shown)


wslack:
Should be merged alongside https://github.com/18F/federalist-docs/pull/219

line47:
The goal is to have different sites based on configurations and not custom code that only applies to a single template. For example, you might have a branch that looks like the report template while another branch looks like a one page site all done through configurations. (the `_data` folder YML files)  

Custom functionality should be built at the higher `uswds-jekyll` level so that any of the Federalist templates would be able to use/inherit functionality instead of it only existing for a single repo or template. 

I agree that submitting issues and PRs for a specific style of template would be a bit more involved. 
I created [corresponding labels](https://github.com/18F/federalist-uswds-template/labels) to help with this, but do think that PRs to particular branches could get a bit tricky. 

Hopefully this helps to clear things up 👍

adborden:
Okay, so the idea would be to remove as much code as possible so the only thing in the repo is configuration and content? That makes sense but how does having them all as branches help with this? I feel like using a shared theme (like we're doing), is all you need.

PR #1739, opened 2018-04-02
use sqs user provided service

amirbey:


config/env/production.js
```diff
@@ -31,10 +31,11 @@ if (s3Creds) {
 }
 
 // SQS Configs
-const sqsKey = env.FEDERALIST_AWS_BUILD_KEY;
-const sqsSecret = env.FEDERALIST_AWS_BUILD_SECRET;
-const sqsQueue = env.FEDERALIST_SQS_QUEUE;
-const sqsRegion = env.FEDERALIST_SQS_REGION;
+const sqsCredentials = appEnv.getServiceCreds(`federalist-${process.env.APP_ENV}-sqs-creds`);
```
el-mapache: I think we are going to need a separate `if` block to handle staging here. On production, the app is going to try and find the SQS credentials using this naming convention and won't be able to.

If you've already created this service in the production environment, we should coordinate with @wslack to figure out a good time to test this.

amirbey:
tested catch guard to environment variables

PR #1741, opened 2018-04-03
[WIP] set up homepage for new thumbnails

wslack:
also modifying the text a little bit


wslack:
@el-mapache this should be ready?

Issue #1685, opened 2018-03-12
Iterate on content for landing-page


ericronne:
#### Description of feature or bug

~Much of the content on the [revised landing page](https://github.com/18F/federalist/issues/1563) is placeholder. We've paid close attention to the design, but not as much to the words (and some supporting imagery) as of yet. Let's address in earnest!~ Content on the landing page has been refined and there's more work to be done!

(See image below)

#### Implementation outline (if higher than "low" effort):

- [x] Clarify primary audience
- [ ] Evaluate and remove/pivot/edit content so that it's geared towards primary audience, eliminating content that less technical users consider to be "hard"
- [x] Change `Documentation` navigation title
- [x] Change `Get started` navigation title
- [ ] Add useful links back into page copy, linking to docs site and other valuable artifacts/sites/documentation.

Other steps are TBD as outputs of #1723. We can open new issues for those as we learn more.

ericronne:
![home-content-needs](https://user-images.githubusercontent.com/9259185/37306008-5ca1177c-2604-11e8-8c28-64e5b1c9da1c.png)

ericronne:
cc @el-mapache

ericronne:
cc @ShannonBad

wslack:
I was hoping we could edit this in the PR of the actual site content...thoughts @thebestsophist?

wslack:
@annahsebok to also review the PR from this AM where I flirted with some content changes. in #1743

annahsebok:
👍 on that PR @wslack (with the exception of one small suggested change). Did you want me to take a look at this as a whole again once those changes are merged?

annahsebok:
@wslack are these stats still right? 
![screen shot 2018-04-18 at 10 48 09 am](https://user-images.githubusercontent.com/6728448/38946209-01e89bca-42f6-11e8-99a6-0a80f7f44e40.png)
I know we're up to 121 live sites. Let me know if any of the other numbers need changing.

annahsebok:
As I've been working on #1757, I've been thinking about whether or not it really makes sense to put a glossary on federalist.18f.gov. Given that it's more of a sales site, I'm thinking we should remove any of the "hard" terms from the homepage (via another PR) and make it even simpler, with the terms less experienced users struggle with being mentioned in a second or third layer of the site (and on the docs site). I'm going to gut check this approach with some other content folks today, and will follow up on any additional changes here after that convo.

annahsebok:
[...]

PR #1743, opened 2018-04-03
add compliance memo link and edit some language

wslack:
Merge with https://github.com/18F/federalist-docs/pull/224

### DIFF
--- a/views/home.njk
+++ b/views/home.njk
@@ -7,8 +7,9 @@
       <div class="usa-grid federalist-hero">
         <div class="usa-width-two-thirds hero-heading">
           <h1>
-            Rapidly launch a compliant website for your federal agency, program or office
+            Rapidly launch a compliant website for your program or office
           </h1>
+          <p>Federalist takes care of the heavy lifting for federal web hosting and compliance, allowing you to focus your expertise on your content.</p>
         </div>
         <div class="usa-width-one-third usa-hero-callout">
           <hr class="hr-dark"></hr>
@@ -25,7 +26,7 @@
     </div>
 
     <div class="usa-grid info-block">
-      <h2><a id="page-body"></a>Federalist takes care of the heavy lifting</h2>
+     <a id="page-body"></a>
       <section class="features">
         <div class="feature-group">
           <div class="usa-width-one-sixth">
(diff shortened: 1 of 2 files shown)


wslack:
I don't think I need to edit https://github.com/18F/federalist/blob/wslack-patch-1/test/api/requests/main.test.js , but not 100% sure

wslack:
Removing 
`          <p>Federalist takes care of the heavy lifting for federal web hosting and compliance, allowing you to focus your expertise on your content.</p>
`
Added `<h2><a id="page-body"></a>Federalist takes care of the heavy lifting</h2>` back in

views/home.njk
```diff
@@ -7,24 +7,25 @@
...
-            Rapidly launch a compliant website for your federal agency, program or office
+            Rapidly launch a compliant website for your federal agency, program, or office
           </h1>
         </div>
         <div class="usa-width-one-third usa-hero-callout">
           <hr class="hr-dark"></hr>
           <h5 class="contrast-heading">Federalist Stats:</h5>
           <div class="usa-hero-callout-stats">
+            <p>119 live .gov sites supported</p>
+            <p>8 federal agency customers</p>
```
annahsebok: I thought we were going with `partners`

annahsebok:
Made one small note (partners vs. customers) but other than that, LGTM!

wslack:
ah - I chose "customers" here because the term applied to the entire agency. "Partners" are what I would call people that log into Federalist to manage sites.

"Agency Partners" seems unclear - as if we were working on Federalist with those agencies.

annahsebok:
@wslack Yes, that makes sense. I'd suggest using "partners" to retain consistency, but given the fact that we're not yet working with the folks mentioned here, I'd suggest "potential agency partners" over "federal agency customers." Varying nouns can be disorienting and makes the way we talk about the people we work with or have the potential to work with inconsistent.

wslack:
Now I'm more confused. We work with 8 agencies - and are working actively with all of those agencies. It's just that the "partners" in those agencies - the civil servants - are only sometimes in an office or two within the whole agency.

annahsebok:
@wslack and I synced up on this — we're going to leave this as `customers` — `partners` is a term that we have more context around internally and could confuse folks about the relationships we have with agencies. I'll be talking a broader look at this when I'm working on #1685.

annahsebok:
In the meantime — @wslack this is 👍 from me!

PR #1764, opened 2018-04-11
Use absolute URL for twitter:image

adborden:
Twitter doesn't seem to like relative URLs
https://github.com/18F/federalist/issues/1662

### DIFF
--- a/views/base.njk
+++ b/views/base.njk
@@ -50,7 +50,7 @@
 
     <!-- For Twitter -->
     <meta name="twitter:card" content="summary" />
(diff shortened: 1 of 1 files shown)


adborden:
Not sure if there is a template variable I can use here. I noticed `og:url` is also hardcoded.

jseppi:
Should have @amirbey review instead, he's the lead dev on the project now :)

Also the PR should be targeted to `staging` so it goes through the proper promotion process.

adborden:
I also updated the URL for the Facebook image since that was giving a warning.

PR #1812, opened 2018-05-01
Build log api opt

amirbey:


api/routers/build-log.js
```diff
@@ -2,7 +2,7 @@ const router = require('express').Router();
 const BuildLogController = require('../controllers/build-log');
 const sessionAuth = require('../policies/sessionAuth');
 
-router.get('/build/:build_id/log', sessionAuth, BuildLogController.find);
+router.get('/build/:build_id/log(/page/:page)?', sessionAuth, BuildLogController.find);
```
jmhooper: We should probably add some requests specs to cover pagination

test/frontend/actions/buildLogActionsTest.js
```diff
@@ -37,7 +37,8 @@ describe("buildActions", () => {
     const buildLogsPromise = Promise.resolve(logs);
     const fetchStartedAction = { action: "fetchStarted" }
     const receivedAction = { action: "received" };
-    fetchBuildLogs.withArgs().returns(buildLogsPromise);
+    fetchBuildLogs.withArgs().onCall(0).returns(buildLogsPromise);
+    fetchBuildLogs.withArgs().onCall(1).returns([]);
```
jmhooper: We may want to make this `.returns(Promise.resolve([]))`, right?

jmhooper:
Looks like codeclimate is upset. One thing that helps get ahead of linter issues is installing the ESLint plugin for your editor. It'll callout linting issues as you type.

test/frontend/actions/buildLogActionsTest.js
amirbey: 👍

amirbey:
@jmhooper - please re-review

PR #1829, opened 2018-05-14
route content/examples to cases studies

amirbey:
fix the content/examples route which was broken by adding case-studies route


api/routers/main.js
```diff
@@ -5,6 +5,7 @@ const csrfProtection = require('../policies/csrfProtection');
 
 router.get('/', MainController.home);
 router.get('/case-studies/', MainController.examples);
+router.get('/content/examples', MainController.examples);
```
el-mapache: Should we maybe have a redirect to `/case-studies`? Then there won't be 2 URLs that point to the same content AND it won't break for people with bookmarks or whatever 😄

PR #1835, opened 2018-05-23
migrate to octokit as existing github package is deprecated

amirbey:


api/services/GitHub.js
```diff
@@ -1,69 +1,27 @@
...
-  github.repos.createFromOrg(options, (err, res) => {
-    if (err) {
-      reject(Object.assign(err, { status: err.code }));
-    } else {
-      resolve(res);
-    }
-  });
-});
+const createRepoForOrg = (github, options) =>
+  github.repos.createForOrg(options);
```
jmhooper: This API call works? I can find it anywhere in the octokit source?

jmhooper: Wait, I'm looking at `octokit`, not `@octokit/rest`. How confusing

Issue #1794, opened 2018-04-24
Create features page for federalist.18f.gov


annahsebok:
#### User Story

As someone interested in Federalist, I'm curious what Federalist provides.

#### Background

We heard from users that they tend to assume Federalist is a CMS. They were also confused about the steps they must take in order to use Federalist. We believe that by creating a separate product page (linked to from the homepage navigation), we will be able to streamline the homepage, clarify what Federalist is and (from a high level) how to use it, and create a dedicated space to outline features and services without overwhelming the homepage.

This will involve moving some of the content on the homepage to this page.

#### Acceptance Criteria

- [x] We have created content for a product page that outlines what Federalist is, how to use it (at a high level), and what cool features it includes.
- [x] We have created a product page that is linked from both navigation bars on the homepage that reflects [Features - Desktop copy 2](https://gsa.invisionapp.com/share/YZKCIN2SKEP#/screens/302673508).

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [x] Clarify primary audience
- [x] Outline priorities for page
- [x] Develop content
- [x] Discuss potential visuals with viz designer
- [x] Build page
- [x] Publish content to page
- [x] Remove content from homepage that is duplicative of features page, eliminating content that less technical users consider to be "hard"
- [x] Change `Documentation` navigation title
- [x] Change `Get started` navigation title
- [x] Add useful links back into homepage page copy, linking to docs site and other valuable artifacts/sites/documentation. 
- [x] Corey or Will to review homepage changes
- [x] Will to ask Aaron about the bottom of the features page
- [x] Corey to do one more features page pass.
- [x] Will to ask Jen about bottom of page.
- [x] Will to make a fixing edit for bottom of page.

annahsebok:
I'm _almost_ finished with the content here and will be wrapping this up tomorrow. Big thanks to @amirbey for trying to help me troubleshoot through getting federalist running locally. He's going to help me get a copy of the case studies page created so that I can then drop this new content in. Also planning on filing an issue with the design lab to help with the overall design of the page while I'm away. Another update coming tomorrow!

annahsebok:
This is in a great place for a review from @wslack, and I've flagged a few specific sections where I'd like his input.

I've also filed [this design lab issue](https://github.com/18F/design-lab/issues/11) to jumpstart work on corresponding visuals for the content. Will followup on this when I'm back in the office.

annahsebok:
[...]

Issue #1869, opened 2018-06-18
Update terms in top and bottom navigational toolbars for federalist.18f.gov and federalist-docs


annahsebok:
#### User Story

As a visitor to federalist.18f.gov, I want to be able to easily find features and pricing, example sites, contact information, and support documentation so that I can make decisions about whether or not to buy Federalist.

#### Background

During our IA work in #1723, we found that majority of users sorted Federalist content into one of four areas - `features & pricing`,`support`, and `contact` - along with something representing `example sites` (if we can think of something better). As a part of [our IA reorganization](https://app.mural.co/t/gsa6/m/gsa6/1522093516875/88001f9a465e1c1a57860657f3cbf6cb4708a7a4) based on a card sort with users, we should rename existing menu options on the top and bottom navigational toolbars to reflect the nomenclature above. 

_Note: We used to refer to the `case studies` page as `example sites`. The change to `case studies` was somewhat recent, but no users identified this content using the phrase `case studies`. `Example sites` was by far the most common title users used to identify the content to be included on this page._

#### Acceptance Criteria

- [ ] The top and bottom navigational toolbars have been updated.
- [ ] The top and bottom navigational toolbars match (excluding the log in button which should only be on the top bar) match with the federalist-docs toolbar.

_After evaluating, edit this part:_

wslack:
This now includes the discussion in https://github.com/18F/federalist/issues/1668

annahsebok:
To be clear, when I mentioned in standup that this needs to happen in tandem with homepage updates, I was referring to https://github.com/18F/federalist/pull/1870, not #1820.

amirbey:
features and pricing - where is the pricing information?

Are we getting rid of Federalist Interest Signup form?

wslack:
We haven't added pricing to the features page yet.

wslack:
We haven't added the pricing issue yet.

The interest form should stay in the menu.

File an issue should disappear. It's never used and the source code is linked in the footer.

PR #1870, opened 2018-06-18
Homepage updates

annahsebok:
Per #1794, I've removed content that less technical users have identified as "hard" (which will be covered on our features page) and made a few plain language edits. I've also updated the `contact` bar information given recent changes to our contact page and approach.

### DIFF
--- a/views/home.njk
+++ b/views/home.njk
@@ -7,7 +7,7 @@
       <div class="usa-grid federalist-hero">
         <div class="usa-width-two-thirds hero-heading">
           <h1>
-            Rapidly launch a compliant website for your federal agency or office
+            Rapidly launch a compliant website for your federal agency
           </h1>
         </div>
         <div class="usa-width-one-third usa-hero-callout">
@@ -35,7 +35,7 @@
           </div>
           <div class="usa-width-three-fourths feature-text">
             <h3>Easy, powerful and fast</h3>
-            <p>You don&#8217;t have to be a programmer to get started or spend months working on compliance. We’ve taken care of it using popular and easy-to-use open source tools so you can get right to work.</p>
+            <p>We’ve taken care of it using popular and easy-to-use open source tools so you can get right to work. You don't have to be a programmer to get started or spend months working on compliance. </p>
           </div>
         </div>
 
@@ -47,7 +47,7 @@
           </div>
           <div class="usa-width-three-fourths feature-text">
             <h3>Fully customizable</h3>
-            <p>Bringing your own design team? Use Jekyll, Hugo, or your own tools to deploy custom sites that leverage Federalist’s scalable hosting architecture.</p>
+            <p>Bringing your own design team? They can deploy custom sites that leverage Federalist’s scalable architecture.</p>
           </div>
         </div>
 
@@ -59,7 +59,7 @@
           </div>
           <div class="usa-width-three-fourths feature-text">
             <h3>Secure and scalable</h3>
-            <p>Built on the FedRAMP-certified Amazon Web Services (AWS) GovCloud infrastructure and services from cloud.gov, our websites offer the highest level of security and scalability.</p>
+            <p>Built on an approved and certified infrastructure with services from [cloud.gov](https://cloud.gov/), our websites offer the highest level of security and scalability.</p>
           </div>
         </div>
       </section>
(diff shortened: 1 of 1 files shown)


views/home.njk
```diff
@@ -7,7 +7,7 @@
       <div class="usa-grid federalist-hero">
         <div class="usa-width-two-thirds hero-heading">
           <h1>
-            Rapidly launch a compliant website for your federal agency or office
+            Rapidly launch a compliant website for your federal agency
```
wslack: I think "agency" alone confuses users because it makes them think they have to be representing the entire agency when they call us.

views/home.njk
```diff
@@ -59,7 +59,7 @@
           </div>
           <div class="usa-width-three-fourths feature-text">
             <h3>Secure and scalable</h3>
-            <p>Built on the FedRAMP-certified Amazon Web Services (AWS) GovCloud infrastructure and services from cloud.gov, our websites offer the highest level of security and scalability.</p>
+            <p>Built on an approved and certified infrastructure with services from [cloud.gov](https://cloud.gov/), our websites offer the highest level of security and scalability.</p>
```
wslack: Maybe say "your websites will have the highest level of...."

views/home.njk
```diff
@@ -117,7 +117,7 @@
         <div class="usa-width-two-thirds description">
           <h2>Launch your site in minutes.</h2>
           <p class="copy">
-            Is your code already on GitHub? Federalist can securely deploy a static website from your repository in minutes. Federalist also deploys site previews from GitHub branches, allowing your team to easily prototype site updates.
+            Federalist uses [GitHub](https://github.com/) as our software development platform. If your code is already on GitHub, Federalist can securely deploy a static website from your repository in minutes.
```
wslack: Should we say GitHub or just that Federalist loads site code from a public code hosting service - and then link to Github?

wslack:
Happy with these edits but should wait until we have a features page. We'll also want to link to the features page a few places in the homepage.

wslack:
Hotfix is in but we have one more to deploy tomorrow.

views/home.njk
annahsebok: @wslack have we heard that feedback from users directly? 

I don't mind using a term other than `agency`, but I think this generally could use some simplification given that it's the call to action and the first thing people see. Additionally, as users dig deeper into the site, they learn more and more about what is offered to them. I suggest we pick just one noun here - `agency`, `office`, `team`, or something else.

wslack: I have heard the confusion about "can I get Federalist for just my office" directly, but not about that specific verbiage.

What about "program?"

annahsebok: I like program! Let's try that and test it.

views/home.njk
annahsebok: Yup! Nice catch. Will edit.

annahsebok:
@wslack definitely agree that this shouldn't go live until the features page is ready. I'm not sure an embedded a link to the features page within this page makes sense though — was there a particular place you were thinking it would be linked? My general take is that people will most likely navigate directly to the features page via the nav bars (which should also be updated before these edits goes live via issue #1869). 

FYI: I've added both the features page and nav update issues as blockers for this PR.

wslack:
I was thinking about linking to it in the prose midpage (where someone might be looking for technical details if scrolling)

annahsebok:
@wslack Interesting. I'd suggest we merge this PR first as is and then review the page again to determine where it might fit best. Links throughout the page can start to become overwhelming, so I want to be sure we're intentionally thoughtful about making that addition.

views/home.njk
```diff
@@ -117,7 +117,7 @@
         <div class="usa-width-two-thirds description">
           <h2>Launch your site in minutes.</h2>
           <p class="copy">
-            Is your code already on GitHub? Federalist can securely deploy a static website from your repository in minutes. Federalist also deploys site previews from GitHub branches, allowing your team to easily prototype site updates.
+            Federalist uses a [public code hosting service](https://github.com/) to load and store code for your site. If your code is already on GitHub, Federalist can securely deploy a static website from your repository in minutes.
```
jmhooper: We don't reference GitHub by name in the first sentence, but do in the second. Does that make sense?

wslack:
@adborden your features page PR included this branch, correct?

adborden:
@wslack right, this can be closed. GH usually detects this, but since I rebased the other branch the commit hashes are different so GH doesn't realize they are the same changes.

PR #1891, opened 2018-06-25
Remove collaborator from users list when removed from site as a collaborator

amirbey:


jmhooper:
Looks like CC is complaining. One of them looks like whitespace which should be an easy fix. For the cognitive complexity, I think we can add a comment there to ignore that since it is a reducer function which will, by their nature, be complex.

PR #1897, opened 2018-06-29
test change up update add site form

wslack:
I might have done this wrong.

### DIFF
--- a/frontend/components/AddSite/AddRepoSiteForm.jsx
+++ b/frontend/components/AddSite/AddRepoSiteForm.jsx
@@ -74,11 +74,11 @@ export const AddRepoSiteForm = ({
         </div>
         <div className="form-group">
           <BranchField
-            label="Primary branch"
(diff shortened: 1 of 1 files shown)


wslack:
Darn it. Do I want "input?"(based on https://github.com/18F/federalist/blob/e96b83ca9b8436ef5c16b7ae98b31a93b51f4641/test/frontend/components/Fields/InputWithErrorField.test.jsx#L7-L17)

wslack:
@amirbey should we do this in place of https://github.com/18F/federalist/pull/1900 ?

amirbey:
we can move forward with both PRs (as is) because together they complete what you were initially attempting to do when raising this PR.
#1897 - updates the label
#1900 - sets a default value for the default branch check field

wslack:
I retriggered the build in code climate and no result..... https://codeclimate.com/github/18F/federalist/compare/wslack-patch-1#issues

Issue #1810, opened 2018-04-30
Light-background thumbnails should be slightly darkened


ericronne:
#### One design subtlety not yet implemented is to slightly darken thumbnail images which have white or very-light backgrounds. Consequently these images tend to blend in with the background, e.g. on the home page …
![image](https://user-images.githubusercontent.com/9259185/39449461-bafa8588-4c8d-11e8-9341-986449051e76.png)

### How they should look:
![image](https://user-images.githubusercontent.com/9259185/39449594-1d2ee82a-4c8e-11e8-823e-c6167ee637e4.png)

### Technique:
![image](https://user-images.githubusercontent.com/9259185/39449624-2e7c3862-4c8e-11e8-9709-05a2faee9697.png)

<TODO>

#### Definition of done
Thumbs are darkened, either via code or simply in sketch

<TODO>

#### Level of effort
Probably low if sketch; medium if code-based

#### Implementation outline (if higher than "low" effort):

<TODO>

wslack:
Holding, pending @ericronne.

wslack:
PR is in, moving to done!

PR #1898, opened 2018-06-29
Added brightness overlay on .figure images & removed whitespace

ctrimm:
This addresses issue #1810 in a very simple PR.

The issue is that when scrolling thru the main landing page, the images do not particularly stand out against the white background of the page. So, a simple solution suggested by @ericronne was to add a darker background and then change the transparency of all the images. I took a slightly different approach by adding `filter: brightness(95%)` to the `img` tags inside of `.figure` div's.

Here are some screenshots of what my proposed solution looks like - 
screen shot 2018-06-29 at 3 10 22 pm
screen shot 2018-06-29 at 3 10 19 pm
screen shot 2018-06-29 at 3 10 15 pm

You can see that there isn't really a background on the Agency Logos screenshot - this is because they all happen to be circular `.png` images. Whereas the screenshots of sample sites are rectangular.

If you would like to have the agency logos with a similar gray background, we can simply apply `background-color: $color-gray-lightest;` to the `img` tag styling.

### DIFF
--- a/frontend/sass/_figure_group.scss
+++ b/frontend/sass/_figure_group.scss
@@ -10,18 +10,22 @@
 
 .figure-group {
   @media screen and (max-width: $mobile-size) {
-    margin: 0 auto;    
+    margin: 0 auto;
   }
 }
 
 .figure {
   margin: 0;
 
+  img {
+    filter: brightness(95%);
+  }
+
   figcaption {
     margin-bottom: 6px;
     margin-top: 20px;
     text-align: left;
-    
+
     @media screen and (max-width: $mobile-size) {
       text-align: center;
     }


ctrimm:
This was manually tested in Chrome 67. However, I realized when I ran IE10 on Windows 7 inside of VirtualBox that IE10 and below does not support filter/brightness. This can also be [verified here](https://caniuse.com/#search=CSS%20filter) that it will only work in IE11 and above.

I can dig into this deeper if you all would like. If not, feel free to ask questions/give feedback/etc.

wslack:
Thank you @ctrimm! I am headed out for the day but we will look at this early next week!

ericronne:
Neat! Thanks, @ctrimm!

The agency logos should all be designed to run over white already, so our solution should exclude them.

ctrimm:
@ericronne The `filter` property only affects colored pixels - so, it'll still add the brightness filter to the agency logos - but the change in negligible, in my opinion. You can verify this by inspecting the DOM.

PR #1905, opened 2018-07-03
Fix proptypes issues or add justificaiton for ignoring

jmhooper:
This commit fixes a place where the propType was set to `object` which
made the linter upset. I also added a comment to ignore the rule in 2
places along with a comment providing justification for ignoring it.


wslack:
Reminder: please open a tiny issue for this or make sure its noted in another issue so we can show it off in sprint review!

wslack:
You know what, this is my job.

PR #1915, opened 2018-07-19
Thumbnail tinting

ericronne:
Applied a bit of a tint to the thumbnails via a color "underlayment," as opposed to the lightness filter which had been previously applied. I didn't apply this to the agency seals; they shouldn't be altered in this way.

image
image

### DIFF
--- a/frontend/sass/_figure_group.scss
+++ b/frontend/sass/_figure_group.scss
@@ -44,6 +44,6 @@
   display: inline-block;
   img {
     display: block;
-    opacity: 0.98;
+    opacity: 0.97;
   }
 }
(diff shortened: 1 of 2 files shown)


ericronne:
An ideal version would use a css multiply "transfer mode" technique, but i.e. doesn't support. So this may be our best bet for now.

frontend/sass/_figure_group.scss
```diff
@@ -42,3 +38,12 @@
     }
   }
 }
+
+.tint {
+  background-color: #1c4e80;
```
adborden: This could be a variable at the top of the file, but no big deal.

wslack:
Love it.

PR #1916, opened 2018-07-19
allow user to remove self as a collaborator regardless of repo permis…

amirbey:
…sions


test/api/requests/site.test.js
```diff
@@ -901,6 +901,49 @@ describe('Site API', () => {
       })
       .catch(done);
     });
+
+    it('should allow a user to remove themselves even if htey are not a repo write user', (done) => {
```
jmhooper: nit, but you gotta type here. `htey`

PR #1929, opened 2018-08-03
Site admin only can delete site and add/remove site collaborators

amirbey:


api/controllers/site.js
```diff
@@ -68,7 +68,7 @@ module.exports = {
         siteJSON = json;
         return authorizer.destroy(req.user, site);
       })
-      .then(() => S3SiteRemover.removeSite(site))
+      .then(() => S3SiteRemover.removeSite(site, req.user))
```
jmhooper: What are your thoughts on moving the GH check to the authorizer? That way the `S3SiteRemove` stays focused on removing sites, and the policies for whether someone is authorized to remove a site are colocated.

wslack:
Amir fixing some tests

amirbey:
@jmhooper - thank for the suggestion. changes made .. pls re-review

amirbey:
@jmhooper - pls re-review

PR #1933, opened 2018-08-08
Copyedits to the homepage

coreycaitlin:
This follows the work on #1794.

🚨  👀please — this includes an edit to the big copy on the homepage


coreycaitlin:
I'm hoping it won't — it feels too long if we add "and offices", but maybe it's worth it? I wish we could just say "federal _folks_" but that's weird. 😆

PR #1943, opened 2018-08-13
Show site delete button as disabled after click

### LINKED ISSUE #1922: Site delete button doesn't respond or demonstrate that the user clicked
#### User Story

As a federalist user who needs to delete my site, I am confused when I click the "delete site" button and nothing seems to happen for a few seconds. I want to know that Federalist is processing my click.

#### Background (Optional)

There's no messaging saying - we're deleting it , please hang on. There's no visual indicator (like the button becoming an activity indicator, for example) that the system is trying to do something. 

#### Acceptance Criteria

- [ ] When the user clicks the delete button, it changes in some way to show the command has been sent to Federalist.

adborden:
I moved the delete button into its own form so that Redux Forms would handle the state for us.

Fixes #1922 

peek 2018-08-13 16-11

### DIFF
--- a/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
+++ b/frontend/components/site/SiteSettings/AdvancedSiteSettings.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import AdvancedSiteSettingsForm from './AdvancedSiteSettingsForm';
+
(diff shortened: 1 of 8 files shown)


wslack:
woot! will defer to @amirbey to review for security and etc.

wslack:
Cool I checked with Aaron and accept the risk.

PR #1948, opened 2018-08-17
add javascript build option

amirbey:


wslack:
I don't like calling this option "javascript" - but am not sure what else to use. Maybe "script only?"

PR #1956, opened 2018-08-24
use the page params to fetch site builds

amirbey:


amirbey:
@jmhooper - thanks. resolved

PR #1971, opened 2018-09-05
add text on the form

wslack:
Editing per retro today to add the Google Form.


wslack:
Changed. Isn't "you can sign up at" valid as well?

PR #1981, opened 2018-09-12
provide admin auth if repo does not exist

amirbey:


api/authorizers/site.js
```diff
@@ -25,6 +25,16 @@ const authorizeAdmin = (user, site) => (
     }
     return Promise.resolve(site.id);
   })
+  .catch((error) => {
+    if (error.code === 404) {
+      // authorize user if the site's repo does not exist
```
jmhooper: Could we add something here about why we need to authorize the repo? I think that is the confusing part.

PR #1984, opened 2018-09-18
sitebuilds preview links moved to left column with icon

amirbey:


jmhooper:
Looks like there's 2 lines over the length requirement, but after trimming those down this should be good.

PR #2018, opened 2018-10-24
socket io to user websocket transport only

amirbey:


amirbey:
https://socket.io/docs/using-multiple-nodes/#Sticky-load-balancing

PR #2032, opened 2018-11-05
Allow users to opt-in/out of push notifications

amirbey:


api/controllers/site-user.js
```diff
@@ -0,0 +1,29 @@
+const siteSerializer = require('../serializers/site');
```
jmhooper: What if instead of naming this API route `site-user`, we did something like `site/:id/notifications`? It could all work the same, but with a different name it may be a little bit easier to parse out what this API route does.

api/models/site-user.js
```diff
@@ -0,0 +1,46 @@
...
+      delete object[key];
+    }
+  });
+
+  return object;
+}
+
+module.exports = (sequelize, DataTypes) => {
+  const SiteUser = sequelize.define('SiteUser', {
+    buildNotify: {
```
jmhooper: Thoughts on renaming this to `pushNotificationsPreference` or something like that?

amirbey: changed to buildNotificationSetting

api/controllers/site-user.js
amirbey: done

test/api/unit/models/site-user.test.js
```diff
@@ -0,0 +1,70 @@
+const expect = require('chai').expect;
+// const config = require('../../../../config');
```
jmhooper: We can probs delete this line

PR #2043, opened 2018-11-28
re-add federalist.18f.gov

amirbey:


app.js
```diff
@@ -42,6 +42,8 @@ const sequelize = require('./api/models').sequelize;
 const redirectUrls = [
   'federalist.fr.cloud.gov',
   'federalist-staging.fr.cloud.gov',
+  'federalist.18f.gov',
```
jmhooper: This looks like it will cause `federalist-staging.18f.gov` and `federalist.18f.gov` to redirect to themselves. That will result in endless redirects.

amirbey: great catch ... i think this is addressed now ... pls re-review

app.js
```diff
@@ -82,6 +81,21 @@ app.use((req, res, next) => {
...
+// temporary until federalist.18f.gov is launched
+app.use((req, res, next) => {
+  const host = req.hostname;
+  const redirectUrls = [
+    'federalist.18f.gov',
+    'federalist-staging.18f.gov',
+  ];
+
+  if (redirectUrls.indexOf(host) !== -1) {
+    return res.redirect(301, `https://${host.slice().replace('federalist', 'federalistapp')}`);
```
jmhooper: If this is temporary, we probably want a 302 instead of a 301

PR #2048, opened 2018-12-04
Unify user interface with main federalist home page

eddietejeda:
This is a large pull request, but it's actually very simple. It just required lots of edits to make it happen and requiring deprecating lots of existing code.

In this request we accomplish the following:
- We want switching between the Federalist site and the Federalist app to be as simple as possible. So we track the various staging, development and production URLs in the manifest and allow user the click around the appropriate environment.
- I removed lots of text that was being rendered within the Federalist app. This content has been moved to the main site. This required changing our tests and redirecting old URLs to the new Federalist site pages.
- I changed the styling and navigation of the header to be consistent with the homepage.
- Given that the static site can't tell when a user is logged in, we had to create the login form page that automatically directs users to Github for auth or directly to the app.
 - The tests all had to be updated to take into account the changes in content on the homepage.


api/controllers/main.js
```diff
@@ -87,7 +71,7 @@ module.exports = {
         title: 'Unauthorized',
         message: 'You are not permitted to perform this action. Are you sure you are logged in?',
       });
-      return res.redirect('/');
+      return res.redirect(config.app.homepageUrl);
```
jmhooper: We set a flash message above this, but that isn't going to have any effect if we redirect to the static site.

eddietejeda: Good catch. Fixed in most recent version.

PR #2094, opened 2019-01-16
Adds auto refresh and a button to toggle

davemcorwin:
- refresh interval is 15 seconds
- updated the UI to always show the buttons, and only replace the table of builds with the loading spinner so it looks a bit smoother


amirbey:
during a release or when server is down ... the following error is displayed
screen shot 2019-01-23 at 10 49 37 am

PR #2110, opened 2019-01-23
Temporarily handles http failures on build and log refreshes

davemcorwin:
## Description
- Provide a slightly more helpful but generic error message to the user when an http error occurs (See below)
- Handle missing data a little more gracefully to avoid JS errors when this happens

## Additional Thoughts
This is a quick fix, but implementing a more robust error-handling solution (and retaining the last known data) is more impactful on the app and requires a bit more thought.

## Feedback
- Let me know what changes should be made to the error message

screen shot 2019-01-23 at 12 28 32 pm


eddietejeda:
Have you seen the issue where the error page does not clear?

eddietejeda:
Clearing on every network request seems fine to me. We need this error message to be performant.

PR #2124, opened 2019-01-30
Only include redux logger and devtools in non-production environments

davemcorwin:


frontend/store.js
```diff
@@ -13,23 +13,31 @@ const reducer = combineReducers({
...
+const middlewares = [
+  reroute,
+  createNotifier(notificationSettings),
+];
+
+const enhancers = [
+  applyMiddleware,
+];
+
+if (process.env.NODE_ENV !== 'production') {
```
davemcorwin: The production webpack build runs with `-p` which adds `process.env.NODE_ENV="production"`. In development, this will be `null` or `undefined`.

Issue #1914, opened 2018-07-17
Automatic access list removal


wslack:
#### User Story

As a site owner, when I remove permissions in GitHub from a Federalist user, Federalist should automatically drop access from the user for the Federalist site based on the repo.

#### Background (Optional)

Right now, Federalist access must be dropped manually (either via being dropped from the app as a whole or being dropped from a site). Doing this automatically lowers the percentage change of errors.

#### Acceptance Criteria

- [x] When user access is removed in GitHub, Federalist drops access to the site for that user.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [x] Investigate different possible paths: webhook or polling//background job
- [x] Decide on path
- [x] Implement
- [x] Test on staging (can't be tested during retro)

PR #2140, opened 2019-02-07
temporarily disable audit users till audited in site user actions

amirbey:


amirbey:
Reference - #1914 
disable until site user actions are tracked

PR #2159, opened 2019-02-25
allow branch names to of character length 1

amirbey:


amirbey:
thanks @rahearn for lending your regex expertise 👍

PR #2163, opened 2019-02-27
upgrade dependencies using old new Buffer() constructor

amirbey:


api/controllers/published-file.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
       if (isNaN(siteId)) {
         throw 404;
       }
-      return Site.findById(siteId);
+      return Site.findByPk(siteId);
```
jmhooper: It is really annoying that they didn't backport this.

api/models/build.js
```diff
@@ -80,25 +80,6 @@ function completeJob(err) {
...
-  object.updatedAt = object.updatedAt.toISOString();
-  if (object.completedAt) {
-    object.completedAt = object.completedAt.toISOString();
-  }
-  Object.keys(object).forEach((key) => {
-    if (object[key] === null) {
-      delete object[key];
-    }
-  });
-  delete object.token;
```
jmhooper: Looks like we may still need to find a way to make sure that `toJSON` calls don't include the token.

jmhooper: Ah, I see that you moved this.

PR #2164, opened 2019-02-27
remove old templates and clean up language for 2.0 templates

eddietejeda:
I remove the old templates and revised the language explaining each template in more detail


davemcorwin:
unclear to me why the code coverage would decline, but @eddietejeda has the power to merge!

eddietejeda:
Sounds good.

PR #2179, opened 2019-03-14
upgrade babel

amirbey:


amirbey:
@davemcorwin  - thanks for the feedback
1) i've added browser versions for ie, chrome and safari
2) looks like this require babel core 7 .. which doesn't appear to have production release yet. perhaps we can update this in the near future

thoughts?

pls. re-review

PR #2199, opened 2019-03-29
Ensure mobile nav handles clicks, minor style cleanup

davemcorwin:


frontend/sass/_header.scss
```diff
@@ -20,6 +20,7 @@
   border-bottom: none;
   min-height: 4rem;
   padding: $spacing-thin 0 $spacing-thin 0;
+  z-index: auto;
```
davemcorwin: This fixes the mobile nav issue. `z-index` was being applied on a non-positioned element, causing issues with the nested elements.

frontend/sass/_header.scss
```diff
@@ -57,6 +58,14 @@
           border-width: 3px;
         }
       }
+
+      @media screen and (max-width: 950px) {
```
davemcorwin: now that `hover` is happening, have appropriate styles in place.

PR #2204, opened 2019-04-04
sticky navigation in site-container

amirbey:


amirbey:
make site navigation sticky when scrolling 

remove the scrolling windows for individual build logs

PR #2229, opened 2019-04-24
Add route creation and mapping when creating a site with a private bucket

apburnes:
## Description

When a user creates a new site with a private bucket, a new route will be generated and mapped to the `federalist-proxy`, the proxy will point to their private aws bucket endpoint, and will allow users to view the site deployments as normal.

## Implementation

- [x] Add methods to the CF API client to manage routes
- [x] Create and map routes when a site is created
- [x] Remove route when site is deleted

### DIFF
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -1,3 +1,4 @@
+const _ = require('underscore');
 const authorizer = require('../authorizers/site');
 const S3SiteRemover = require('../services/S3SiteRemover');
 const SiteCreator = require('../services/SiteCreator');
@@ -7,9 +8,9 @@ const siteSerializer = require('../serializers/site');
 const { User, Site, Build } = require('../models');
 const siteErrors = require('../responses/siteErrors');
 
-const sendJSON = (site, res) =>
-  siteSerializer.serialize(site)
-    .then(siteJSON => res.json(siteJSON));
+const sendJSON = (site, res) => siteSerializer
+  .serialize(site)
+  .then(siteJSON => res.json(siteJSON));
 
 module.exports = {
   findAllForUser: (req, res) => {
@@ -24,25 +25,26 @@ module.exports = {
     let site;
 
     Promise.resolve(Number(req.params.id)).then((id) => {
-      if (isNaN(id)) {
+      if (_.isNaN(id)) {
         throw 404;
       }
       return Site.findByPk(id);
-    }).then((model) => {
-      if (model) {
-        site = model;
-      } else {
-        throw 404;
-      }
-      return authorizer.findOne(req.user, site);
     })
-    .then(() => siteSerializer.serialize(site))
-    .then((siteJSON) => {
-      res.json(siteJSON);
-    })
-    .catch((err) => {
-      res.error(err);
-    });
+      .then((model) => {
+        if (model) {
+          site = model;
+        } else {
+          throw 404;
+        }
+        return authorizer.findOne(req.user, site);
+      })
+      .then(() => siteSerializer.serialize(site))
+      .then((siteJSON) => {
+        res.json(siteJSON);
+      })
+      .catch((err) => {
+        res.error(err);
+      });
   },
 
   destroy: (req, res) => {
@@ -51,7 +53,7 @@ module.exports = {
 
     Promise.resolve(Number(req.params.id))
       .then((id) => {
-        if (isNaN(id)) {
+        if (_.isNaN(id)) {
           throw 404;
         }
         return Site.findByPk(id);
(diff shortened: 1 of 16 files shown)


api/controllers/site.js
```diff
@@ -24,25 +25,26 @@ module.exports = {
     let site;
 
     Promise.resolve(Number(req.params.id)).then((id) => {
-      if (isNaN(id)) {
+      if (_.isNaN(id)) {
```
amirbey: is this necessary? this style isNaN is not used throughout the codebase

api/controllers/site.js
```diff
@@ -24,25 +25,26 @@ module.exports = {
...
-      return authorizer.findOne(req.user, site);
     })
-    .then(() => siteSerializer.serialize(site))
-    .then((siteJSON) => {
-      res.json(siteJSON);
-    })
-    .catch((err) => {
-      res.error(err);
-    });
+      .then((model) => {
```
amirbey: indented 1 too many?

api/controllers/site.js
```diff
@@ -51,7 +53,7 @@ module.exports = {
 
     Promise.resolve(Number(req.params.id))
       .then((id) => {
-        if (isNaN(id)) {
+        if (_.isNaN(id)) {
```
amirbey: s this necessary? this style isNaN is not used throughout the codebase

api/controllers/site.js
```diff
@@ -104,7 +107,7 @@ module.exports = {
     const userId = Number(req.params.user_id);
     let site;
 
-    if (isNaN(siteId) || isNaN(userId)) {
+    if (_.isNaN(siteId) || _.isNaN(userId)) {
```
amirbey: s this necessary? this style isNaN is not used throughout the codebase

api/controllers/site.js
```diff
@@ -124,26 +127,23 @@ module.exports = {
 
       return authorizer.removeUser(req.user, site);
     })
-    .then(() =>
-      SiteMembershipCreator.revokeSiteMembership({ user: req.user, site, userId })
-    )
-    .then(() =>
-      UserActionCreator.addRemoveAction({
+      .then(() => SiteMembershipCreator
```
amirbey: is this a linting issue? 100 characters?

api/controllers/site.js
```diff
@@ -124,26 +127,23 @@ module.exports = {
...
+        .revokeSiteMembership({ user: req.user, site, userId }))
```
amirbey: indented 1 too many?

api/controllers/site.js
```diff
@@ -160,51 +160,52 @@ module.exports = {
     const siteId = Number(req.params.id);
 
     Promise.resolve(siteId).then((id) => {
-      if (isNaN(id)) {
+      if (_.isNaN(id)) {
```
amirbey: s this necessary? this style isNaN is not used throughout the codebase

api/controllers/site.js
```diff
@@ -160,51 +160,52 @@ module.exports = {
...
-      site: siteId,
-      branch: model.defaultBranch,
-    }))
-    .then(() => {
-      if (site.demoBranch) {
-        return Build.create({
-          user: req.user.id,
-          site: siteId,
-          branch: site.demoBranch,
+      .then((model) => {
```
amirbey: indented 1 too many?

api/controllers/site.js
```diff
@@ -160,51 +160,52 @@ module.exports = {
...
+      }))
+      .then(() => {
+        if (site.demoBranch) {
+          return Build.create({
+            user: req.user.id,
+            site: siteId,
+            branch: site.demoBranch,
+          });
+        }
+        return null;
```
amirbey: return null or Promise.resolve() ? or is that the same thing? for consistency, i think the codebase uses Promise.resolve()

api/controllers/site.js
apburnes: The `isNaN` was throwing linting error's so I used the underscore version.

api/controllers/site.js
apburnes: The `isNaN` was throwing linting error's so I used the underscore version.

api/controllers/site.js
apburnes: Lint error thrown on spacing.

api/controllers/site.js
apburnes: The isNaN was throwing linting error's so I used the underscore version.

api/controllers/site.js
apburnes: Lint error. No line break after arrow function

api/controllers/site.js
apburnes: Linting error on spacing

api/controllers/site.js
apburnes: The isNaN was throwing linting error's so I used the underscore version

api/controllers/site.js
apburnes: Linting again

api/controllers/site.js
apburnes: This is linting again... But we can update to `Promise.resolve`

PR #2230, opened 2019-04-25
Additional auth flow that allows Netlify CMS to authenticate with Github

davemcorwin:
External authentication implementation that allows Federalist users to authenticate Netlify CMS with Github.

Here we create a second Github authentication strategy for Passport. This allows an application to hit the login url explicitly in a popup window, and when authentication is complete, send a message to the parent window that Netlify CMS is expecting.

## Questions/Concerns
- [x] Do we want to restrict to current Federalist users? **YES**
- This still has negative security implications until sites are deployed to their own subdomains.

## Remaining tasks
- [x] Add automated "request" tests for external authentication urls
- [x] Review security implications/browser support of `window.postMessage` API
- ~[] QA test against non-Federalist user~ Currently the only check is if the user is in an approved org. See follow up issues.

## Before merge to STAGING
- [ ] Add an additional env variable `GITHUB_CLIENT_EXTERNAL_CALLBACK_URL` with a value of `<host>/external/auth/github/callback`

## Before merge to MASTER
- [ ] Update the "callback url" specified in the production Federalist OAuth application in Github to only be the federalist domain.
- [ ] Add an additional env variable `GITHUB_CLIENT_EXTERNAL_CALLBACK_URL` with a value of `<host>/external/auth/github/callback`


davemcorwin:
@eddietejeda the issue we experienced is due to 18F organization-level permissions granted to OAuth applications. The default is to allow read-only access to public data which is consistent with the behavior we observed when trying to use Netlify CMS for an 18F repo: `GET` requests are successful, but `POST`s fail.

In this case we could ask the 18F org to grant more permissions to the Federalist application, but folks who have repos in different orgs would need their org to do the same thing. This is not an issue for repos in personal accounts (not orgs).

Thoughts?

eddietejeda:
I see. Can you check with the Github admins?

davemcorwin:
Oooh, its possible the permissions are only missing from `FederalistLocal`! I am requesting

apburnes:
How can we block authentication (or any Netlify CMS usage) when site is hosted from the shared bucket?

davemcorwin:
@apburnes hmmmm, if this is determined by repo perhaps we can pass that along as a query param and verify thats its not deployed to a shared bucket before authenticating?

apburnes:
@davemcorwin that sounds solid.

davemcorwin:
As outlined in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Security_concerns) the major security risk in leveraging the `postMessage` api revolves around accepting messages from unknown parties and potentially executing malicious code.

I believe the [implementation](https://github.com/netlify/netlify-cms/blob/master/packages/netlify-cms-lib-auth/src/netlify-auth.js) satisfies the security concerns by:
- confirming the origin of the message when receiving a message in the client window
- explicitly checking the entire string message against a regex before `JSON.parse`

PR #2251, opened 2019-05-08
ensure engine select options are valid site.engines options

amirbey:


amirbey:
@davemcorwin - fair enough .. make sense ... i have modified the PR to make sure that the options in the dropdown are all one of the expected possible values

PR #2271, opened 2019-05-30
retry buildreport status if fails

amirbey:


eddietejeda:
@amirbey what's the status on this?

amirbey:
@eddietejeda - Soon after implementing this, I investigated this issue further, failed submission to GitHub of build status doesn't appear to be an issue.  There is no evidence supporting this; therefore, I put this issue/PR on ice.  However, I only see what has been done here as an addition to our codebase.  We can move forward with it. But first, I want to revisit its test cases before submitting for review.

PR #2299, opened 2019-07-19
allow fedearlist-users admins to add any site

amirbey:


api/services/SiteMembershipCreator.js
```diff
@@ -1,23 +1,41 @@
...
-        status: 400,
-      };
-    }
-    return true;
-  });
+const checkGithubRepository = ({ user, owner, repository }) => {
+  let repo;
+  return GitHub.getRepository(user, owner, repository)
+    .then((_repo) => {
+      repo = _repo;
```
apburnes: You don't need to assign `_repo` to `repo`.  You could just `return repo;` after the first `if` and pick it back up in the next promise.

api/services/SiteMembershipCreator.js
```diff
@@ -1,23 +1,41 @@
...
+        return FederalistUsersHelper.federalistUsersAdmins(user.username)
+          .then((admins) => {
+            if (!admins.includes(user.username)) {
+              throw {
+                message: siteErrors.WRITE_ACCESS_REQUIRED,
+                status: 400,
+              };
+            }
+          })
+          .catch(() => {
```
apburnes: You shouldn't need to catch this again since you threw the error just before.

amirbey: just in case, i'm catching ... FederalistUsersHelper.federalistUsersAdmins(user.username)

api/services/SiteMembershipCreator.js
amirbey: good point .. i've combined the blocks

api/services/SiteMembershipCreator.js
```diff
@@ -1,23 +1,36 @@
...
-    }
-    if (!repo.permissions.push) {
-      throw {
-        message: siteErrors.WRITE_ACCESS_REQUIRED,
-        status: 400,
-      };
-    }
-    return true;
-  });
+const checkGithubRepository = ({ user, owner, repository }) =>
```
apburnes: Codeclimate says you've got trailing space here.  🌡 👾

PR #2327, opened 2019-09-06
set preview links to old proxy

amirbey:


amirbey:
site.createdAt basded on https://github.com/18F/federalist/pull/2276

PR #2354, opened 2019-10-04
Try to report status if build user's token is invalid by repo write access user

amirbey:


api/services/GithubBuildStatusReporter.js
```diff
@@ -18,12 +18,14 @@ const checkAccessTokenPermissions = (users, site) => {
...
+      if (permissions.push) {
         return Promise.resolve(user.githubAccessToken);
       }
-
       count += 1;
-
+      return getNextToken(users[count]);
+    })
+    .catch(() => {
+      count += 1;
```
apburnes: Should we log the catch error?

amirbey: i suppose we can ... it may help support conversations

PR #2356, opened 2019-10-08
use internal routing for garden-build to federalist

amirbey:


amirbey:
yeah ... you have to create the internal route and then add a network policy around it

PR #2362, opened 2019-10-09
update build status if there is an error parsing the message

amirbey:


api/controllers/build.js
```diff
@@ -90,6 +90,15 @@ module.exports = {
   status: (req, res) => {
     let message;
     Promise.resolve(decodeb64(req.body.message))
+    .catch((err) => {
```
davemcorwin: I'm not sure this will work as you intend. Typically you would call `resolve` AFTER something has been successful. If the error occurs while providing the argument, I believe it is treated as if it happened BEFORE the promise, which means this will throw before the promise is created and will not be caught.

In general, this method could be simplified quite a bit by refactoring out the unnecessary promise chaining, but if we just start with the addition, perhaps something like:
```
let message;
try {
  message = decode64(req.body.message);
} catch (err) {
  const errMsg = [
    `Error decoding build status message for build@id=${req.params.id}`,
    `build@message: ${req.body.message}`,
    err,
  ];
  logger.error(errMsg.join('\n'));
  message = 'build status message parsing error';
}
return Promise.resolve(Number(req.params.id))
...
...
```

amirbey: i did test this out and it does work as expected.  A promise can be returned from the catch block, and also happens other places in this codebase.  Since it does, work I think this more of a stylistic preference.  Please let me know if this is still an issue for you.

amirbey: i did test this and dit did behave as expected. however, after thinking about it more i do think you have an interesting point. i have incorporated your suggestion.  thanks. pls re-review

PR #2368, opened 2019-10-16
add network policy to deploy script

amirbey:


amirbey:
this is more to add the req'd network policy ..... doing this dynamically, would require further scripting .... 1) we could go with an mvp and do this iteratively ... or 2) pump the brakes and revisit this

PR #2371, opened 2019-10-21
Hotfix: Remove user audit upon login

amirbey:


amirbey:
https://developer.github.com/v3/repos/#list-your-repositories
is no longer returning all repos for user

PR #2373, opened 2019-10-22
consider case when auditing a site users

amirbey:


amirbey:
codecimate linting hung ... to be investigated

PR #2385, opened 2019-11-08
Upgrade Sequelize to latest due to security alert

amirbey:


amirbey:
https://github.com/advisories/GHSA-m9jw-237r-gvfv

PR #2421, opened 2019-12-14
Removed user authorizer

davemcorwin:
The user authorizer was not doing anything here. `req.user` contains the authenticated user from the passport process. It is not necessary to requery for the user and verify that it has the same id since we just queried for it by id! This is different than if the user id came from the UI (ie `req.params.id`, ...)

In addition, there is no need to requery for the user in the serializer, since we already have it. In general, the serializers should be pure functions that accept an object and turn it into an HTTP-friendly one, it should not do any additional querying.


api/controllers/user.js
```diff
@@ -16,10 +15,6 @@ module.exports = {
   },
 
   me: (req, res) => {
-    User.findByPk(req.user.id)
-      .then(model => authorizer.me(req.user, model))
-      .then(currentUser => userSerializer.serialize(currentUser))
-      .then(userJSON => res.json(userJSON))
-      .catch(err => res.error(err));
+    res.json(userSerializer.toJSON(req.user));
```
amirbey: @davemcorwin -  how does this now handle errors? is it the same in the end as the catch block?

davemcorwin: Added generic error handler that will handle this. Above, call next with any errors to forward there.  See https://expressjs.com/en/guide/error-handling.html.

app.js
```diff
@@ -148,30 +149,34 @@ app.use((err, req, res, next) => {
   next(err);
 });
 
-app.use((req, res) => res.status(404).redirect(302, '/404-not-found/'));
+// Generic error handler
+// eslint-disable-next-line no-unused-vars
+app.use((err, _req, res, _next) => {
+  res.status(500).send({ message: `An unexpected error occurred: ${err.message}` });
+});
```
davemcorwin: new generic error handler

app.js
```diff
@@ -148,30 +149,34 @@ app.use((err, req, res, next) => {
...
-})
-.on('error', (err) => {
-  logger.error(`socket auth/subscribe error: ${err}`);
-});
+  .on('connection', (_socket) => {
+    SocketIOSubscriber.joinRooms(_socket);
+  })
+  .on('error', (err) => {
+    logger.error(`socket auth/subscribe error: ${err}`);
+  });
```
davemcorwin: everything else in this file was just eslint related refactoring.

PR #2423, opened 2019-12-26
Removed unused methods of user serializer, some linting cleanup

davemcorwin:


davemcorwin:
@amirbey soooo, code climate is failing with a bunch o stuff, except it won't show it to me because it's too much... Not sure how to resolve. It doesn't correspond to the eslint errors (if i don't run `lint:diff`)

PR #2428, opened 2019-12-31
queued status updated to processing by garden build

amirbey:


api/controllers/build.js
```diff
@@ -88,52 +88,55 @@ module.exports = {
...
+    let buildStatus;
 
-    const getStatusMessage = (statusRequest) => {
-      let statusMessage;
+    const getBuildStatus = (statusRequest) => {
+      let status;
+      let message;
       try {
-        statusMessage = decodeb64(statusRequest.body.message);
+        status = decodeb64(statusRequest.body.status);
```
davemcorwin: I don't think the status is being base64 encoded in the garden build

api/models/build.js
```diff
@@ -44,25 +44,22 @@ const beforeValidate = (build) => {
...
 const sanitizeCompleteJobErrorMessage = message => message.replace(/\/\/(.*)@github/g, '//[token_redacted]@github');
 
-const completeJobErrorMessage = (err) => {
-  let message = 'An unknown error occurred';
-  if (err) {
-    message = err.message || err;
-  }
-  return sanitizeCompleteJobErrorMessage(message);
-};
+const jobErrorMessage = ({ message = 'An unknown error occurred' }) => sanitizeCompleteJobErrorMessage(message);
```
davemcorwin: does this method need to take an object containing the `message` or can it just take the `message` as an argument?

api/models/build.js
```diff
@@ -44,25 +44,22 @@ const beforeValidate = (build) => {
...
 
-const completeJobStateUpdate = (err, build, completedAt) => {
-  if (err) {
+const jobStateUpdate = (buildStatus, build, completedAt) => {
```
davemcorwin: If the garden build returns the expected status ("success" instead of "complete") then this logic can be simplified a bit.

api/services/GithubBuildStatusReporter.js
```diff
@@ -107,7 +107,7 @@ const reportBuildStatus = (build) => {
       context,
     };
 
-    if (build.state === 'processing') {
+    if (['processing', 'queued'].find(s => s === build.state)) {
```
davemcorwin: Could this use `includes` instead? Ex:
```
 if (['processing', 'queued'].includes(build.state))
```

test/api/requests/build.test.js
```diff
@@ -508,7 +508,7 @@ describe('Build API', () => {
         .post(`/v0/build/${options.build.id}/status/${buildToken}`)
         .type('json')
         .send({
-          status: options.status,
+          status: encode64(options.status),
```
davemcorwin: Same as above, I don't think the status is base64 encoded

api/models/build.js
```diff
@@ -79,10 +76,10 @@ const completeJobSiteUpdate = (build, completedAt) => {
...
 
-function completeJob(err) {
+function updateJobStatus(buildStatus) {
   const completedAt = new Date();
 
-  return completeJobStateUpdate(err, this, completedAt)
+  return jobStateUpdate(buildStatus, this, completedAt)
     .then(build => completeJobSiteUpdate(build, completedAt)
       .then(() => build));
 }
```
davemcorwin: Now or as a future PR, consider refactoring a bit to be a little more clear, would love something like:
```
async function updateJobStatus(buildStatus) {
  const completedAt = new Date();
  const build = await jobStateUpdate(buildStatus, this, completedAt);
  await completeJobSiteUpdate(build, completedAt);
  return build;
}
```

Also consider moving the conditional logic if the status is "success" into this method and out of `completeJobSiteUpdate` Ex.
```
async function updateJobStatus(buildStatus) {
  const completedAt = new Date();
  const build = await jobStateUpdate(buildStatus, this, completedAt);
  if (buildStatus.state === 'success') {
    await completeJobSiteUpdate(build, completedAt);
  }
  return build;
}
```

then the job of `completeJobSiteUpdate` becomes much simpler and does one thing and always returns the site.

api/controllers/build.js
amirbey: goog catch .... i pushed this in garden build

api/services/GithubBuildStatusReporter.js
amirbey: good suggestion ... changed ... thanks 👍

test/api/requests/build.test.js
amirbey: encoding pushed in garden build

PR #2434, opened 2020-01-08
Handle new logs with source 'ALL'

davemcorwin:
Update the API to query for new style logs that will have a record per log message. Some refactoring along the way including updates to the `create-dev-data` script.

Update the UI to handle many log record per source.

Add some additional patterns for async request handlers and error handling.

### DIFF
--- a/api/controllers/build-log.js
+++ b/api/controllers/build-log.js
@@ -1,3 +1,4 @@
+const { wrapHandlers } = require('../utils');
 const buildLogSerializer = require('../serializers/build-log');
 const { Build, BuildLog } = require('../models');
 
@@ -8,31 +9,27 @@ function decodeb64(str) {
   return null;
 }
 
-module.exports = {
-  create: (req, res) => {
-    Promise.resolve(Number(req.params.build_id))
-      .then((id) => {
-        if (isNaN(id)) { throw 404; }
-        return Build.findByPk(id);
-      })
-      .then((build) => {
-        if (!build) {
-          throw 404;
-        } else if (build.token !== req.params.token) {
-          throw 403;
-        }
+module.exports = wrapHandlers({
+  create: async (req, res) => {
+    const { build_id: buildId, token } = req.params;
 
-        return BuildLog.create({
-          build: build.id,
-          output: decodeb64(req.body.output),
-          source: req.body.source,
-        });
-      })
-      .then(buildLog => buildLogSerializer.serialize(buildLog))
-      .then((buildLogJSON) => { res.json(buildLogJSON); })
-      .catch((err) => {
-        res.error(err);
-      });
+    const build = await Build.findOne({ where: { id: buildId, token } });
+
+    if (!build) {
+      return res.notFound();
+    }
+
+    const { output, source } = req.body;
+
+    const buildLog = await BuildLog.create({
+      build: build.id,
+      output: decodeb64(output),
+      source,
+    });
+
+    const buildLogJSON = await buildLogSerializer.serialize(buildLog);
+
+    return res.ok(buildLogJSON);
   },
 
   find: async (req, res) => {
@@ -43,7 +40,7 @@ module.exports = {
     const build = await Build.forUser(user).findByPk(buildId);
 
     if (!build) {
-      return res.status(404).send('Not Found');
+      return res.notFound();
     }
 
     const buildLogs = await BuildLog.findAll({
@@ -56,6 +53,6 @@ module.exports = {
 
     const serializedBuildLogs = await buildLogSerializer.serialize(buildLogs);
 
-    return res.json(serializedBuildLogs);
+    return res.ok(serializedBuildLogs);
   },
-};
+});
--- a/api/controllers/build-log.js
+++ b/api/controllers/build-log.js
@@ -1,3 +1,4 @@
+const Seq = require('sequelize');
 const { wrapHandlers } = require('../utils');
 const buildLogSerializer = require('../serializers/build-log');
 const { Build, BuildLog } = require('../models');
@@ -34,6 +35,11 @@ module.exports = wrapHandlers({
 
   find: async (req, res) => {
     const limit = 5;
+    // A page is set to a maximum of 200,000 characters. Assuming
+    // a each log line is about 200 characters, that is roughly
+    // equal to 1,000 lines
+    const lineLimit = 5 * 1000;
+
     const { params, user } = req;
     const { build_id: buildId, page = 1 } = params;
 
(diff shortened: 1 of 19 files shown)


api/controllers/build-log.js
```diff
@@ -9,74 +10,75 @@ function decodeb64(str) {
...
-        build: build.id,
-        output: decodeb64(req.body.output),
-        source: req.body.source,
-      });
-    })
-    .then(buildLog => buildLogSerializer.serialize(buildLog))
-    .then((buildLogJSON) => { res.json(buildLogJSON); })
-    .catch((err) => {
-      res.error(err);
+module.exports = wrapHandlers({
```
davemcorwin: Use `wrapHandlers` to have an easy way of handling errors when using `async` since they need to be explicitly caught. With this helper method, we don't have to remember to always add `try...catch` blocks. See the implementation for more info.

api/controllers/build-log.js
```diff
@@ -9,74 +10,75 @@ function decodeb64(str) {
...
+  create: async (req, res) => {
+    const { build_id: buildId, token } = req.params;
+
+    const build = await Build.findOne({ where: { id: buildId, token } });
```
davemcorwin: A lot of the logic appears to be gone, here's the thought process.
1. Include the token in the query for the build so we just get a 404 back and don't have to do an explicit check. This follows a line of thought that says to just return a 404 instead of 403 for authorization failure to prevent any information from being communicated to anyone with bad intent.
2. Allow the type conversion to be handled in Sequelize and throw an error which is eventually handled in a common place (see the change to the `error` response handler) and results in a 404.

api/controllers/build-log.js
```diff
@@ -9,74 +10,75 @@ function decodeb64(str) {
...
+    if (buildLogs.length === 0) {
+      buildLogs = await BuildLog.findAll({
+        where: {
+          build: build.id,
+          source: { [Seq.Op.ne]: 'ALL' },
+        },
+        order: [['id', 'ASC']],
+        offset: (limit * (page - 1)),
+        limit,
+        include: [Build],
```
davemcorwin: this could definitely be refactored, perhaps into a scope on the model, but I think this is ok for now. We'll see if we run into any bottlenecks.

api/responses/error.js
```diff
@@ -10,6 +12,10 @@ module.exports = (error = {}, { res }) => {
       status: 400,
       message: 'The request parameters were invalid.',
     };
+  } else if (error instanceof DatabaseError) {
+    finalError = {
+      status: 404,
+    };
   }
```
davemcorwin: handle the sequelize type conversion error.

scripts/create-dev-data.js
```diff
@@ -1,14 +1,162 @@
...
+      user: user1.id,
+      token: 'fake-token',
+    }),
+    Build.create({
+      branch: 'dc/fixes',
+      source: 'fake-build',
+      site: site2.id,
+      user: user1.id,
+      token: 'fake-token',
+    }).then(build => build.update({ commitSha: '57ce109dcc2cb8675ccbc2d023f40f82a2deabe1' })),
```
amirbey: curious , why is token updated and not part of create object?

davemcorwin: I dunno, I just copy/pasted it....

test/api/requests/build-logs.test.js
```diff
@@ -176,103 +172,134 @@ describe('Build Log API', () => {
...
 
     it('should response with a 404 if the given build does not exist', (done) => {
       authenticatedSession().then(cookie => request(app)
         .get('/v0/build/-100/log')
         .set('Cookie', cookie)
         .expect(404)).then((response) => {
-          validateAgainstJSONSchema('GET', '/build/{build_id}/log', 404, response.body);
-          done();
-        }).catch(done);
+        validateAgainstJSONSchema('GET', '/build/{build_id}/log', 404, response.body);
```
amirbey: indentation was removed?

test/api/unit/models/build.test.js
```diff
@@ -71,14 +72,12 @@ describe('Build model', () => {
 
   describe('.completeJob(message)', () => {
     it('should mark a build errored with a message', (done) => {
-      factory.build().then(build =>
-        build.updateJobStatus({ status: 'error', message: 'this is an error' })
-      ).then((build) => {
+      factory.build().then(build => build.updateJobStatus({ status: 'error', message: 'this is an error' })).then((build) => {
```
amirbey: is this formatting better? i find the former easier to read

test/api/unit/models/build.test.js
```diff
@@ -89,28 +88,27 @@ describe('Build model', () => {
...
 
         return promisedValues.build.updateJobStatus({ status: 'success' });
       }).then(() => Site.findByPk(site.id))
-      .then((model) => {
-        expect(model.publishedAt).to.be.a('date');
-        expect(new Date().getTime() - model.publishedAt.getTime()).to.be.below(500);
-        done();
-      })
-      .catch(done);
+        .then((model) => {
```
amirbey: i'm seeign a lot of indention changes? this is your editor?  this seems to not be in sync with previous lines

test/api/unit/models/build.test.js
```diff
@@ -89,28 +88,27 @@ describe('Build model', () => {
...
+          done();
+        })
+        .catch(done);
     });
 
     it('should sanitize GitHub access tokens from error message', (done) => {
-      factory.build().then(build =>
-        build.updateJobStatus({ status: 'error', message: 'http://123abc@github.com' })
-      ).then((build) => {
+      factory.build().then(build => build.updateJobStatus({ status: 'error', message: 'http://123abc@github.com' })).then((build) => {
```
amirbey: all on on line a bit harder to read

test/api/unit/models/build.test.js
```diff
@@ -119,28 +117,24 @@ describe('Build model', () => {
       Build.create({
         user: 1,
         site: null,
-      }).then(() =>
-        done(new Error('Expected a validation error'))
-      ).catch((err) => {
+      }).then(() => done(new Error('Expected a validation error'))).catch((err) => {
```
amirbey: can this line be multi-line?

test/api/unit/models/build.test.js
```diff
@@ -119,28 +117,24 @@ describe('Build model', () => {
...
     });
 
     it('should require a user object before saving', (done) => {
       Build.create({
         user: null,
         site: 1,
-      }).then(() =>
-        done(new Error('Expected a validation error'))
-      ).catch((err) => {
+      }).then(() => done(new Error('Expected a validation error'))).catch((err) => {
```
amirbey: can this be multiline?

test/api/requests/build-logs.test.js
davemcorwin: weird formatting, fixing.

test/api/unit/models/build.test.js
davemcorwin: some of this was probably auto-formatted...

davemcorwin:
@amirbey tried to update formatting.

PR #2435, opened 2020-01-10
track when build started and display to user

amirbey:


api/models/build.js
```diff
@@ -46,19 +46,24 @@ const sanitizeCompleteJobErrorMessage = message => message.replace(/\/\/(.*)@git
 
 const jobErrorMessage = (message = 'An unknown error occurred') => sanitizeCompleteJobErrorMessage(message);
 
-const jobStateUpdate = (buildStatus, build, completedAt) => {
-  let error = null;
+const jobStateUpdate = (buildStatus, build, timestamp) => {
```
davemcorwin: Looks good. Food for thought for future work: we have a good ol' fashioned state machine here and as the complexity increases we may want to consider making this a bit more explicit.

test/api/unit/models/build.test.js
```diff
@@ -52,16 +52,44 @@ describe('Build model', () => {
   });
 
   describe('after create hook', () => {
+    it('should send a build new build message', (done) => {
+      const startedAt = new Date();
+      factory.build({ state: 'processing', startedAt })
+        .then((build) => {
+          expect(build.completedAt).to.be.null;
+          return build.updateJobStatus({ status: 'success' });
```
davemcorwin: If the subject of this test is the `afterCreate` hook, why do we need to call `updateJobStatus` here? Testing that functionality should probably be in a separate spec.

test/api/unit/models/build.test.js
```diff
@@ -52,16 +52,44 @@ describe('Build model', () => {
   });
 
   describe('after create hook', () => {
+    it('should send a build new build message', (done) => {
```
davemcorwin: Thoughts on using the newer `async/await` syntax here to be consistent with other new tests?

test/api/unit/models/build.test.js
```diff
@@ -52,16 +52,44 @@ describe('Build model', () => {
...
+
+          done();
+        })
+        .catch(done);
+    });
+
     it('should send a build new build message', (done) => {
       factory.build()
-        .then(build => build.updateJobStatus({ status: 'complete' }))
+        .then((build) => {
```
davemcorwin: it looks like these two specs have the same name? It also looks like here we're testing `updateJobStatus` again, maybe tests for that should be elsewhere?

test/api/unit/models/build.test.js
```diff
@@ -71,11 +99,16 @@ describe('Build model', () => {
 
   describe('.completeJob(message)', () => {
```
davemcorwin: should this be `describe('.updateJobStatus'...`?

test/api/unit/models/build.test.js
```diff
@@ -71,11 +99,16 @@ describe('Build model', () => {
...
     it('should mark a build errored with a message', (done) => {
-      factory.build().then(build =>
-        build.updateJobStatus({ status: 'error', message: 'this is an error' })
-      ).then((build) => {
+      factory.build({ state: 'processing', startedAt: new Date() })
```
davemcorwin: Might be enough logic that it's worth testing all the different state transitions

test/api/unit/models/build.test.js
```diff
@@ -71,11 +99,16 @@ describe('Build model', () => {
...
+      .then((build) => {
+        expect(build.completedAt).to.be.null;
+        expect(build.startedAt).to.be.a('date');
```
davemcorwin: when possible, let's use the actual date instead of `.to.be.a.('date')`

test/api/requests/build.test.js
```diff
@@ -623,6 +623,127 @@ describe('Build API', () => {
...
+          status: 'success',
+          message: '',
+        });
+      })
+      .then(() => Site.findByPk(siteId))
+      .then((site) => {
+        expect(site.publishedAt).to.be.a('date');
+        return Build.findByPk(buildId);
+      })
+      .then((build) => {
```
davemcorwin: It looks like testing the effects on the build are handled in a different spec (`should mark a build successful`), consider omitting them here or more explicitly combining them into one spec.

test/api/requests/build.test.js
```diff
@@ -623,6 +623,127 @@ describe('Build API', () => {
       .catch(done);
     });
 
+    it('should update the publishedAt from processing -> success status', (done) => {
```
davemcorwin: Generally or these, I would like to see the `async/await` syntax used here. At least for me, it's a lot easier to see what is actually under test. With the promises, the actual method is buried in the promise chain a bit.

Happy to pair on it if you're not totally comfortable right now.

In some of these specs we are testing the effects of the request on both the site and the build. Consider either completely separating them OR making it more explicit in the spec name and code that we are testing both here.

frontend/components/site/siteBuilds.js
```diff
@@ -185,7 +185,7 @@ class SiteBuilds extends React.Component {
                       <td data-title="Completed"><span>{ timeFrom(build.completedAt) }</span></td>
                       <td data-title="Duration">
                         <span>
-                          { duration(build.createdAt, build.completedAt) }
+                          { duration(build.startedAt || build.createdAt, build.completedAt) }
```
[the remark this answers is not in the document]
amirbey: you have a point  ... i thought about the blank duration while it queues; however, the duration shows the amount of time the job is waiting in the queue.  when status is in progress it show the appropriate time for that status. i started to think it was a value add. if you think its confusing, i'm fine with scrapping the queued time. thoughts???

api/models/build.js
amirbey: agreed ... i'm not a fan of this function either .... could user refactoring

frontend/components/site/siteBuilds.js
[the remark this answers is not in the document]
davemcorwin: Do folks care about the queued time? There is nothing they can do about it and it's not actionable like build times mostly). We'll still have the data if we want to evaluate queue times, just not sure users will care. For Ex. when things are queued in Circle, are the times displayed?

test/api/unit/models/build.test.js
amirbey: it appears implementing this change affects many of the other test because of the affects on the SQS stub ... i'd rather wait to overhaul this test file

test/api/unit/models/build.test.js
amirbey: i spent much time debugging this this morning ... even @apburnes joined in. we uncovered that it it was introduced as a delay ... i have replaced it with a different delay

test/api/unit/models/build.test.js
```diff
@@ -52,54 +53,74 @@ describe('Build model', () => {
...
   describe('after create hook', () => {
-    it('should send a build new build message', (done) => {
-      factory.build()
-        .then(build => build.updateJobStatus({ status: 'complete' }))
-        .then((build) => {
-          const queuedBuild = sendMessageStub.getCall(0).args[0];
-          const buildCount = sendMessageStub.getCall(0).args[1];
+    it('should send a build new build message', async () => {
+      const startedAt = new Date();
+      let build = await factory.build();
```
davemcorwin: prefer `const` over `let` unless the variable will be reassigned, I think linting should catch this...

test/api/unit/models/build.test.js
```diff
@@ -52,54 +53,74 @@ describe('Build model', () => {
...
+      // create delay while s3 infra create will be removed with 1 bucket federalist
+      await factory.site();
```
davemcorwin: If the purpose of `await factory.site();` is to create the delay mentioned in the comment above, maybe we can replace this with something more semantic utility method like
```
await wait();
```
where
```
function wait(ms = 100) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
```

test/api/unit/models/build.test.js
```diff
@@ -52,54 +53,74 @@ describe('Build model', () => {
...
+        factory.build()
+          .then((build) => {
+            expect(build.state).to.eql('queued');
+            expect(build.startedAt).to.be.null;
+            return build.updateJobStatus({ status: 'processing' });
+          })
+          .then((build) => {
+            expect(build.completedAt).to.be.null;
+            expect(build.startedAt).to.be.a('date');
+            expect(build.startedAt.getTime()).to.above(build.createdAt);
```
davemcorwin: should this be `...to.be.above...`?

test/api/unit/models/build.test.js
```diff
@@ -52,54 +53,74 @@ describe('Build model', () => {
...
+            return build.updateJobStatus({ status: 'error', message: 'this is an error' });
+          }).then((build) => {
+            expect(build.state).to.equal('error');
+            expect(build.error).to.equal('this is an error');
+            expect(build.completedAt).to.be.a('date');
+            expect(build.completedAt.getTime()).to.above(build.startedAt.getTime());
+            return Site.findByPk(build.site);
```
davemcorwin: `to.be.above`

test/api/unit/models/build.test.js
```diff
@@ -52,54 +53,74 @@ describe('Build model', () => {
...
+        factory.build()
+        .then((build) => {
+          expect(build.state).to.eql('queued');
+          expect(build.startedAt).to.be.null;
+          return build.updateJobStatus({ status: 'processing' });
+        }).then((build) => {
+          expect(build.completedAt).to.be.null;
+          expect(build.startedAt).to.be.a('date');
+          expect(build.startedAt.getTime()).to.above(build.createdAt.getTime());
+          return build.updateJobStatus({ status: 'success' });
```
davemcorwin: `to.be.above`

test/api/unit/models/build.test.js
```diff
@@ -1,5 +1,6 @@
 const { expect } = require('chai');
 const { stub } = require('sinon');
+const _ = require('underscore');
```
davemcorwin: hmm, i can't see where this is used...

amirbey: removed .... we were trying to use the wait functionality and opted out

test/api/unit/models/build.test.js
```diff
@@ -52,54 +52,75 @@ describe('Build model', () => {
...
+        factory.build()
+          .then((build) => {
+            expect(build.state).to.eql('queued');
+            expect(build.startedAt).to.be.null;
+            return build.updateJobStatus({ status: 'processing' });
+          })
+          .then((build) => {
+            expect(build.completedAt).to.be.null;
+            expect(build.startedAt).to.be.a('date');
+            expect(build.startedAt.getTime()).to.above(build.createdAt);
```
davemcorwin:
```suggestion
            expect(build.startedAt.getTime()).to.be.above(build.createdAt);
```

PR #2451, opened 2020-01-23
Build url per build

amirbey:


api/services/GithubBuildStatusReporter.js
```diff
@@ -80,6 +80,7 @@ const loadBuildUserAccessToken = build =>
 const reportBuildStatus = (build) => {
   let site;
   let options = {};
+  let viewLink;
 
   return new Promise((resolve, reject) => {
```
davemcorwin: another good target for refactoring!

frontend/components/branchViewLink.js
```diff
@@ -7,32 +7,33 @@ import { IconView } from './icons';
...
 
-const getUrlAndViewText = (branchName, site, completedAt) => {
-  if (isDefaultBranch(branchName, site)) {
-    return { url: site.viewLink, viewText: 'View site' };
-  }
-  if (isDemoBranch(branchName, site)) {
-    return { url: site.demoViewLink, viewText: 'View demo' };
-  }
+const getUrlAndViewText = (branchName, buildURL, site, completedAt) => {
+  const urlWithSlash = (rawUrl) => {
```
davemcorwin: Instead of duplicating this logic in the frontend, can we just add the viewLink when serializing the build?

test/api/unit/models/build.test.js
```diff
@@ -132,6 +132,63 @@ describe('Build model', () => {
...
 
+  describe('build.url settting', () => {
+    let site;
+    before(async () => {
+      site = await factory.site({ defaultBranch: 'master', demoBranch: 'staging' });
+    });
+
+    it('default branch url start with site', async () => {
+      let build = await factory.build({ branch: site.defaultBranch, site });
+      build = await Build.findByPk(build.id);
```
davemcorwin: why do we need to requery for the build?

migrations/20200117094246-add-url-at-to-build.js
```diff
@@ -0,0 +1,5 @@
+exports.up = (db, callback) =>
+  db.addColumn('build', 'url', { type: 'string' }, callback);
```
apburnes: Will we need to insert the `url` based on the site's `awsBucketName` column and the AWS bucket url.

ie `<awsBucketName><S3 Bucket Region>.amazonaws.com`.

amirbey: if i understand your statement correctly ...  yes, because of the way we are using the serializers .... it seems so ... let me know if you still see otherwise .... plus @davemcorwin feels very strongly about storing the deployment urls in the database

davemcorwin: @apburnes Are you asking whether we need to save THIS url (the one with the bucket name) vs. what's called the `viewLink` or the prettier one?

apburnes: My bad on `<S3 Bucket Region>.amazonaws.com`, dismiss that...

The migration is not populating the old builds with a `url` for either the corresponding `<awsBucketName>.app.cloud.gov` or `federalist-proxy.app.cloud.gov`. Is this happening in the serializer and I'm just missing it?

test/api/unit/models/build.test.js
amirbey: initially url was assigned in aftercreate hook .. this is old junk i failed to remove

migrations/20200117094246-add-url-at-to-build.js
amirbey: we agreed we would update the old urls before the bucket migration to backfill those values

api/services/SQS.js
```diff
@@ -25,18 +25,10 @@ const siteConfig = (build) => {
...
-    return `site/${build.Site.owner}/${build.Site.repository}`;
-  }
-  if (demoBranch(build)) {
-    return `demo/${build.Site.owner}/${build.Site.repository}`;
-  }
-  return `preview/${build.Site.owner}/${build.Site.repository}/${build.branch}`;
-};
-
-const baseURLForCustomDomain = (rawDomain) => {
+const baseURLForDomain = (rawDomain) => {
```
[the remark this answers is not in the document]
amirbey: it's referring to the baseURL environment variable

api/utils/build.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function buildUrl(build, site) {
+  let path = `/preview/${site.owner}/${site.repository}/${build.branch}`;
+  if (build.branch === site.defaultBranch) {
+    path = `/site/${site.owner}/${site.repository}`;
+  } else if (build.branch === site.demoBranch) {
+    path = `/demo/${site.owner}/${site.repository}`;
+  }
+  return `${[`https://${site.awsBucketName}.app.cloud.gov`, path].join('')}`;
```
davemcorwin: maybe just old school?
```suggestion
  return `https://${site.awsBucketName}.app.cloud.gov` +  path;
```

frontend/components/branchViewLink.js
```diff
@@ -7,32 +7,18 @@ import { IconView } from './icons';
...
-    url: `https://${site.awsBucketName}.app.cloud.gov/preview/${site.owner}/${site.repository}/${branchName}/`,
-    viewText: 'Preview site',
-  };
+  return { url: viewLink, viewText };
 };
 
-export const BranchViewLink = ({ branchName, site, showIcon, completedAt }) => {
-  const { url, viewText } = getUrlAndViewText(branchName, site, completedAt);
+export const BranchViewLink = ({ branchName, viewLink, site, showIcon, completedAt }) => {
+  const { url, viewText } = getUrlAndViewText(branchName, viewLink, site, completedAt);
```
davemcorwin: Doesn't look like we need to pass in `viewLink` here, we just spit it back out as `url`...

public/swagger/Build.json
```diff
@@ -64,6 +64,12 @@
     "updatedAt": {
       "type": "string"
     },
+    "url": {
```
davemcorwin: Conceptually `url` is the "internal" url and `viewLink` is the "external" url, I'm wondering if we don't want to serialize and send `url` anymore since we don't really want folks using it. Do we use it anywhere in the UI? Would need a corresponding change in the serializer.

test/api/unit/utils/validators.test.js
```diff
@@ -53,4 +53,24 @@ describe('validators', () => {
...
+      expect(validators.isEmptyOrUrl(null)).to.be.undefined;
+    });
+    it('should allow empty string', () => {
+      expect(validators.isEmptyOrUrl('')).to.be.undefined;
+    });
+    it('should allow url', () => {
+      expect(validators.isEmptyOrUrl('https://hello.world')).to.be.undefined;
+    });
+    it('should not allow non url', () => {
+      try {
```
davemcorwin: I think this can be
```
      expect(() => validators.isEmptyOrUrl('hu')).to.throw(Error, 'URL must start with https://');
```

test/frontend/components/branchViewLink.test.js
```diff
@@ -8,44 +8,58 @@ describe('<BranchViewLink/>', () => {
...
-    demoViewLink: 'https://demo-url.com',
+    domain: 'https://prod-url.com',
+    demoDomain: 'https://demo-url.com',
     owner: 'test-owner',
     repository: 'test-repo',
     awsBucketName: 'test-bucket',
     s3ServiceName: 'federalist-production-s3',
   };
 
+  const getViewLink = (build) => {
```
davemcorwin: The UI is just spitting out  whatever the backend is giving it, does it care what it actually is? Do the tests that are using this actually care? They look like they are just checking that the `href` matches `viewLink` so let's just test that, I don't think we need to logic here to generate the correct url.

api/models/build.js
```diff
@@ -59,26 +61,25 @@ const jobStateUpdate = (buildStatus, build, timestamp) => {
...
 };
 
-const completeJobSiteUpdate = (build, completedAt) => {
-  const { Site } = build.sequelize.models;
-  return Site.update(
-    { publishedAt: completedAt },
-    { where: { id: build.site } }
-  );
-};
+const completeJobSiteUpdate = (site, completedAt) => site.update({ publishedAt: completedAt });
```
davemcorwin: This might make more sense as a function on the Site now!

test/frontend/components/branchViewLink.test.js
amirbey: are you suggesting bringing the Build model into the frontend?

api/models/build.js
amirbey: i don't think this really requires its own function. i have removed the function and just call site.update

test/frontend/components/branchViewLink.test.js
davemcorwin: No, i'm saying all we're testing here is that the `href` reflects the value passed in by the `viewLink` prop, so it can be any random string.

davemcorwin: So we don't need a function to generate a "real" url.

api/utils/build.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function buildUrl(build, site) {
+  let path = `/preview/${site.owner}/${site.repository}/${build.branch}`;
+  if (build.branch === site.defaultBranch) {
+    path = `/site/${site.owner}/${site.repository}`;
+  } else if (build.branch === site.demoBranch) {
+    path = `/demo/${site.owner}/${site.repository}`;
+  }
+  return `https://${site.awsBucketName}.app.cloud.gov` +  path;
```
davemcorwin: there was an extra space
```suggestion
  return `https://${site.awsBucketName}.app.cloud.gov` + path;
```

PR #2512, opened 2020-03-11
removing new relic

amirbey:


amirbey:
closed. instead moved new relic env variables to CF .... allowing quick enable/disable of new relic

PR #2519, opened 2020-03-27
Move new relic config from ups to cf env

amirbey:


amirbey:
yes, you're right ... i was just trying to be consistent with the use of the env variables in other parts of the app and the config was already defined in app.js

PR #2520, opened 2020-03-27
Update the logging time in  S3Helper.js

amirbey:


amirbey:
@apburnes - was the start date intentionally the same? if so, pls this PR

apburnes:
@amirbey no. Looks good.

PR #2521, opened 2020-03-30
fix buildlog limit and log server errors

amirbey:


davemcorwin:
@amirbey can you add some context to this?

amirbey:
intermittently we have been seeing the 500 error on the build logs page.  I was peeking at the code and noticed the limit attribute may be improperly set so i fixed it. and then added testing to make sure we are testing pagination for the new logging system. i was also able to re-create the 500 error in the browser but was not seeing any errors while tailing the logs.  In the middleware error handler i added logging.

app.js
```diff
@@ -141,6 +141,7 @@ app.use(expressErrorLogger);
 
 // eslint-disable-next-line
 app.use((err, req, res, next) => {
+  logger.error(err.stack);
```
davemcorwin: a couple lines above this you should see
```
app.use(expressErrorLogger);
```

This should log any errors and then send them along for further processing (such as by THIS error handler). This is configured in `./winston.js` and perhaps we need to update this configuration to provided better logging of errors?

PR #2530, opened 2020-03-31
index foreign keys

amirbey:


amirbey:
i initially wrote it that way ... but then i changed it b/c i wanted to  IF NOT EXISTS across our environments .... potentially less headache

also, there is also use of db.runSql in our migrations as well

PR #2550, opened 2020-04-09
Proxy on the edge

amirbey:


amirbey:
Proxy can be demoed at - federalistapp-proxy-edge-dev.fr.cloud.gov

api/controllers/site.js
```diff
@@ -73,6 +89,12 @@ module.exports = {
       })
       .then(() => S3SiteRemover.removeSite(site))
       .then(() => S3SiteRemover.removeInfrastructure(site))
+      .then(() => {
+        if (Features.enabled(Features.Flags.FEATURE_PROXY_EDGE_DYNAMO)) {
+          return ProxyDataSync.removeSite(site)
```
davemcorwin, line 94: This doesn't necessarily need to be resolved right now.

There is a slight difference in behavior between this implementation, where the promise from `ProxyDataSync.removeSite(site)` is returned as part of the chain, and below when using `async`/`await`.

Here, the next step will not run until `ProxyDataSync.removeSite(site)` completes, but the `async`/`await` version, treats it as "fire and forget".

Perhaps we should make sure these behave consistently.

api/services/DynamoDBDocumentHelper.js
```diff
@@ -0,0 +1,54 @@
...
+class DynamoDBDocumentHelper {
+  constructor(credentials) {
+    this.docClient = new AWS.DynamoDB.DocumentClient({
+      accessKeyId: credentials.accessKeyId,
+      secretAccessKey: credentials.secretAccessKey,
+      region: credentials.region,
+    });
+  }
+
+  put(TableName, Item) {
```
davemcorwin, line 20: Does not need to be resolved now, and I assume you did this for conciseness, but I think the function arguments should be camelcase

config/app.js
```diff
@@ -3,4 +3,6 @@ module.exports = {
   app_env: process.env.APP_ENV || 'development',
   homepageUrl: process.env.HOMEPAGE_URL || 'http://localhost:4000',
   s3ServicePlanId: process.env.S3_SERVICE_PLAN_ID || 'myFederalistS3BrokerGuid',
+  proxySiteTable: process.env.PROXY_SITE_TABLE || 'federalist-proxy-dev',
+  proxyPreviewHost: process.env.FEDERALIST_PREVIEW_HOSTNAME || 'http://*.sites-local.localhost:1337',
```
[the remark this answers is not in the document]
amirbey: you're right ... the intent is  just to generate a url on the local env

api/services/DynamoDBDocumentHelper.js
amirbey: yeah .. .it can be changes ... this titlezed variables is the way the objects are defined in all of the AWS documentation I've seen.  we could create untitlized variables with the same name for uniformity

PR #2551, opened 2020-04-10
Wait for credential availability for S3 client, add generic helpers, …

davemcorwin:
…use promises.

### DIFF
--- a/test/api/unit/utils/utils.test.js
+++ b/test/api/unit/utils/utils.test.js
@@ -317,4 +317,73 @@ describe('utils', () => {
       });
     });
   });
+
+  describe('.wait', () => {
+    it('pauses execution for `time` milliseconds', async () => {
+      const time = 100;
+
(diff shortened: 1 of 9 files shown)


test/api/unit/utils/utils.test.js
```diff
@@ -317,4 +317,73 @@ describe('utils', () => {
...
+
+      it('rejects with the error with which `fn` rejects', async () => {
+        const result = await utils.retry(spy, { maxAttempts }).catch(err => err);
+        expect(result).to.eq(error);
+      });
+    });
+
+    describe('when `fn` eventually resolves', () => {
+      const value = 'hi';
+      const fn = (() => {
```
davemcorwin, line 368: This is particularly ugly because we are using Sinon 1.x instead of the most recent 9.x where `resolves`, `rejects`, `onFirstCall`, etc are available.

PR #2570, opened 2020-04-15
create buckets in federalist aws account using federalist s3 broker

amirbey:
appears federalist is creating buckets in cloud.gov s3 account using cloud.gov default s3 broker


amirbey:
@davemcorwin - yes, that was correct.  we had previously reverted back to the cloud.gov s3 broker when we reached the limit for s3 buckets in our account.  looks like the manfiest didn't et updated back to our borker once the s3 buckets were re-allocated to our account

PR #2585, opened 2020-04-28
allow branch names to contain apostrophes

amirbey:


eddietejeda:
I take it that the ' is being satanized elsewhere and can't be use escape out of the string and potentially execute code. Can you confirm?

PR #2586, opened 2020-04-29
Revert "allow branch names to contain apostrophes"

amirbey:
Reverts 18F/federalist#2585


amirbey:
reverting ... to deal with this later in a more broad context

PR #2660, opened 2020-05-27
2659 new redis broker

davemcorwin:
Updates redis configuration to support the BETA redis service, if present.

Once deployed, just need to rename the services and restage.


package.json
```diff
@@ -35,6 +35,7 @@
     "pg": "^8.2.1",
     "pg-native": "^3.0.0",
     "rc": "^1.2.8",
+    "redis": "~2.8.0",
```
davemcorwin, line 38: This is to keep in sync with the redis version required by `socket.io-redis`

PR #2666, opened 2020-05-29
Upgrade to Node 12

davemcorwin:
- Upgrades to Node 12 the current LTS
- renames Socket IO instance `socket` to `socketIO` to not clobber existing `response.socket`
- use `http.createServer`


davemcorwin:
Woohoo, found the issue! We were assigning the Socket IO instance to response.socket, overwriting the existing socket attribute. The corking and uncorking of the socket when writing to the http stream was added in Node 16 so it wasn't triggered until now. Nuts.

PR #2691, opened 2020-06-11
Basic auth

amirbey:


api/utils/validators.js
```diff
@@ -76,6 +76,10 @@ function isEmptyOrUrl(value) {
   }
 }
 
+const validBasicAuthUsername = s => /^[a-zA-Z0-9_-]{4,15}$/.test(s);
+
+const validBasicAuthPassword = s => /^[a-zA-Z0-9_\-$!@]{6,25}$/.test(s);
```
davemcorwin, line 81: How were the regexes determined?

frontend/components/site/SiteSettings/BasicAuthSettings.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+const infoContent = (
+  <Fragment>
+    If you would like to add basic authenticaiton to your site previews, please provide a
+    username and password for which you will be prompted to enter when previewing your site.
+    For more information see
+    {' '}
+    <a
+      href="https://federalist.18f.gov/documentation/basic-auth"
+      rel="noopener"
```
davemcorwin, line 23: maybe add `target="_blank"` ?

frontend/components/site/SiteSettings/BasicAuthSettings.jsx
```diff
@@ -0,0 +1,129 @@
...
+import {
+  saveBasicAuth,
+  removeBasicAuth,
+  fetchBasicAuth,
+} from '../../../actions/basicAuthActions';
+import { BASIC_AUTH } from '../../../propTypes';
+
+const infoContent = (
+  <Fragment>
+    If you would like to add basic authenticaiton to your site previews, please provide a
```
davemcorwin, line 17:
```suggestion
    If you would like to add basic authentication to your site previews, please provide a
```

frontend/components/site/SiteSettings/BasicAuthSettings.jsx
```diff
@@ -0,0 +1,129 @@
...
+const warningContent = (
+  <Fragment>
+    Federalist is a
+    <b> FISMA Low </b>
+    system, do NOT reuse credentials from other secure systems. See
+    {' '}
+    <a
+      href="https://csrc.nist.gov/Projects/Risk-Management/Risk-Management-Framework-(RMF)-Overview/Security-Categorization"
+      rel="noopener noreferrer"
+    >
```
davemcorwin, line 41: `target="_blank"`?

frontend/selectors/basicAuth/index.js
```diff
@@ -0,0 +1,4 @@
+export function siteBasicAuth(state, siteId) {
```
davemcorwin, line 1: This is not used, but if it was, it wouldn't work because the state is not keyed by site id...

frontend/reducers/basicAuth.js
```diff
@@ -0,0 +1,59 @@
...
+
+    case REMOVED: {
+      return {
+        isLoading: false,
+        data: {},
+      };
+    }
+
+    default:
+      return state;
```
davemcorwin, line 57: For the UEVs, the state is keyed by site Id, but these aren't, is there a reason why they should be treated differently? This implies that the state contains the basic auth credentials for the *current* site, but this is a global store and could lead to problems if the site changed but the BasicAuth component(s) was/were NOT remounted.

That said, since this data is part of the site schema, why not just include it when fetching the site?

frontend/components/site/SiteSettings/BasicAuthSettingsForm.jsx
```diff
@@ -0,0 +1,72 @@
...
+          id="basicAuthUsernameInput"
+        />
+        <Field
+          name="password"
+          type="password"
+          label="Password:"
+          component={InputWithErrorField}
+          required
+          validate={[validateBasicAuthPassword]}
+          placeholder="**********"
```
davemcorwin, line 51: Maybe omit the placeholder, users can find them confusing and we already have a nice label

frontend/components/site/SiteSettings/BasicAuthSettingsForm.jsx
```diff
@@ -0,0 +1,72 @@
...
+          To enable basic authentication, please submit a username and password credentials required to preview your site builds.
+        </p>
+        <Field
+          name="username"
+          type="text"
+          label="Username:"
+          component={InputWithErrorField}
+          required
+          validate={[validateBasicAuthUsername]}
+          placeholder="username"
```
davemcorwin, line 41: Maybe omit the placeholder, users can find them confusing and we already have a nice label

frontend/selectors/basicAuth/index.js
amirbey: thanks for caching this .. meant to delete this file

api/utils/validators.js
amirbey: @davemcorwin - init just keeping it simple. do you have suggestions to augment it?

davemcorwin: Just wondering where it came from. It feels too restrictive to me and wondering if there are reasons for that. For example, my email address wouldn't work as a username.

amirbey: upated based on https://owasp.org/www-community/OWASP_Validation_Regex_Repository safetext and password

frontend/actions/siteActions.js
```diff
@@ -109,4 +113,26 @@ export default {
...
   },
+
+  removeBasicAuthFromSite(siteId) {
+    return federalist.removeBasicAuthFromSite(siteId)
+      .then((site) => {
+        if (site) {
+          dispatchSiteBasicAuthRemovedAction(site);
+        }
+      })
+      .then(() => alertActions.alertSuccess('Successfully removed basic authentcation.'))
```
davemcorwin, line 124:
```suggestion
      .then(() => alertActions.alertSuccess('Successfully removed basic authentication.'))
```

frontend/util/validators.js
```diff
@@ -2,6 +2,10 @@ const yaml = require('js-yaml');
 
 const validBranchName = s => /^[\w._]+(?:[/-]*[\w._])*$/.test(s);
 
+const validBasicAuthUsername = s => /^(?!.*[:])(?=.*[a-zA-Z0-9]).{4,255}$/.test(s) ? undefined : 'Value is required and must be at least 4 characters. Allowed symbols: .-_@';
+
+const validBasicAuthPassword = s => /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z]).{4,255}$/.test(s) ? undefined : 'Value is required and must be at least 4 characters. At leaset 1 uppercase, 1 lowercase and 1 number is required';
```
davemcorwin, line 7:
```suggestion
const validBasicAuthPassword = s => /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z]).{4,255}$/.test(s) ? undefined : 'Value is required and must be at least 4 characters. At least 1 uppercase, 1 lowercase and 1 number is required';
```

frontend/util/validators.js
```diff
@@ -2,6 +2,10 @@ const yaml = require('js-yaml');
 
 const validBranchName = s => /^[\w._]+(?:[/-]*[\w._])*$/.test(s);
 
+const validBasicAuthUsername = s => /^(?!.*[:])(?=.*[a-zA-Z0-9]).{4,255}$/.test(s) ? undefined : 'Value is required and must be at least 4 characters. Allowed symbols: .-_@';
```
davemcorwin, line 5: Is the allowed symbols in the message still accurate?

public/swagger/Site.json
```diff
@@ -67,6 +68,17 @@
     "previewLink": {
       "type": "string"
     },
+    "basicAuth": {
```
davemcorwin, line 71: should this just be a reference to the BasicAuth.json schema?

amirbey: i don't think its necessary to maintain a separate file for basicAuth. initially BasicAuth.json was its own file because it was an api response all on its own.  now its part of the site response

frontend/actions/siteActions.js
```diff
@@ -109,4 +113,26 @@ export default {
...
+  },
+
+  saveBasicAuthToSite(siteId, basicAuth) {
+    return federalist.saveBasicAuthToSite(siteId, basicAuth)
+      .then(site => {
+        if (site) {
+          dispatchSiteBasicAuthSavedAction(site);
+        }
+      })
+      .then(() => alertActions.alertSuccess('Successfully added basic authentcation.'))
```
davemcorwin, line 135:
```suggestion
      .then(() => alertActions.alertSuccess('Successfully added basic authentication.'))
```

public/swagger/BasicAuth.json
```diff
@@ -0,0 +1,11 @@
+{
```
davemcorwin, line 1: Should this file be removed?

frontend/actions/siteActions.js
amirbey: @davemcorwin  - good point. what about instead of "Disable" we change the button to "Delete" or "Remove"?

PR #2709, opened 2020-06-17
Move nightly builds to run from Circle and some refactoring

davemcorwin:


api/services/ScheduledBuildHelper.js
```diff
@@ -52,7 +63,16 @@ const nightlyBuilds = () => Site.findAll({
...
+};
+
+const nightlyBuilds = async () => {
+  const sites = await Site.findAll(query);
+
+  const builds = sites
+    .map(siteBranches)
+    .flat();
+
+  return Promise.allSettled(builds);
```
davemcorwin, line 75: use `allSettled` to ensure all of them complete in case one fails and report all of the information back.

.circleci/config.yml
```diff
@@ -113,4 +159,15 @@ workflows:
...
+                - master
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - staging
+                - master
```
davemcorwin, line 171: I don't see why we wouldn't want these to run in staging as well, would like to keep the envs as close as possible and maybe we'll find out if something is broken in staging first.

PR #2727, opened 2020-06-25
add timestamps of when the site was updated and when the dynamo recor…

amirbey:
…d updated


amirbey:
- add timestamps to assist with future data sync debugging
- capitalize dynamoDB table attributes in accord with common styling for dynamoDB usage

PR #2767, opened 2020-07-16
add subdomain colum to sites and update proxy db script to include site.config

amirbey:


davemcorwin:
@amirbey It looks like the purpose here is to decouple the proxy subdomain from the CF S3 service name, which sounds great to me! That said, if the subdomain is now decoupled from the service name and bucket name, why does the logic for computing the subdomain differ for dedicated vs shared buckets? Shouldn't it be the same since we now have a mapping in the table from subdomain to bucket?

amirbey:
@davemcorwin - we are using the s3service name as the subdomain/site key in the proxyDBTable.  In staging, many of the sites are still in the shared bucket and are causing errors in the proxyDB sync script because the s3ServiceName/subdomains/site keys are not currently unique.  The goal is to make sure each site has its own unique subdomain whether the site even if it's in the shared bucket.

davemcorwin:
1. Is the logic in the migration script equivalent to just setting ALL subdomains to the result of `generateS3ServiceName()` since the existing service names for dedicated buckets already use that logic?
2. Since we are decoupling, would it make sense to have an additional `generateSubdomain()` function even if it only calls `generateS3ServiceName()` for now?
3. If 1 and 2, maybe we should just set ALL subdomains to the result of `generateSubDomain()` since the decoupling will have the effect of being indifferent to whether the site is using a dedicated or shared bucket without creating an artifact that is specific to the current state of the staging environment.
4. Perhaps we should migrate all the sites in staging so it is NOT different than production...

amirbey:
@davemcorwin - i have added the geenerateSubdomain function.
agree we should either migrate or delete the other sites

davemcorwin:
Cool. Should the migration be simplified as well?

amirbey:
@davemcorwin - what is the simplification you seek?

migrations/20200716100000-set-site-subdomain.js
```diff
@@ -0,0 +1,17 @@
+const yaml = require('js-yaml');
+const { generateSubdomain } = require('../api/utils');
+
+const cmdUpdateDedicatedBucketSitesSubdomain = 'UPDATE site SET subdomain = "s3ServiceName" WHERE "s3ServiceName" !~ \'federalist-(production|staging)-s3\'';
+const getSharedBucketSites = 'SELECT id, owner, repository from site WHERE "s3ServiceName" ~ \'federalist-(production|staging)-s3\'';
+const cmdUpdateSiteSubdomain = site => `UPDATE site SET subdomain = '${generateSubdomain(site.owner, site.repository)}' WHERE id = ${site.id}`;
+const cmdUnsetAllSiteSubdomains = 'UPDATE site SET subdomain = null';
+
+exports.up = (db, callback) => db.runSql(cmdUpdateDedicatedBucketSitesSubdomain)
```
davemcorwin, line 9: Why do we need to 2 different update statements, can we just update all the sites to the result of `generateSubdomain`? It shouldn't matter what the existing value is, right?

PR #2783, opened 2020-07-20
2765 feature flags

davemcorwin:
Here is an initial try adding environment-specific feature flags. Hopefully this survives contact with the enemy, let me know your thoughts and if the Readme content is helpful.


README.md
```diff
@@ -207,6 +207,54 @@ docker-compose run app yarn lint:diff
 docker-compose run app node_modules/.bin/eslint --fix path/to/file.js
 ```
 
+## Feature Flags
+Environment-specific feature flags are supported for both the api and frontend. Flagged features are assumed to be "off" unless the flag exists (and the value is truthy), thus feature flag conditions should always check for the presence or truthiness of the flag, not for it's absence. Environment variables for feature flags *MUST* be prefixed by `FEATURE_`, ex. `FEATURE_BRING_THE_AWESOME`.
```
apburnes, line 211: Suggestions for the example feature variable name [`FEATURE_BRING_THE_NOISE`](https://www.youtube.com/watch?v=kl1hgXfX5-U) or [`FEATURE_BRING_DA_RUCKUS`](https://www.youtube.com/watch?v=LRhRNgqPlyk) 😝

README.md
```diff
@@ -207,6 +207,54 @@ docker-compose run app yarn lint:diff
...
+### Frontend feature flags
+Frontend feature flags are evaluated at *compile* time NOT at runtime, resulting in unused codepaths being exluded from the built code. Environment variables with the `FEATURE_` prefix are available globally within the codebase at compile time. This magic is performed by `webpack.DefinePlugin` and minification.
+
+
+Example:
+
+Given environment variable `FEATURE_AWESOME_SAUCE='true'`
+
+```js
+if (FEATURE_AWESOME_SAUCE === 'true') {
```
apburnes, line 245: For consistency, what are you thoughts on webpack's `getFeatureFlags` removing `FEATURE_` from the env variable like the api flags or vice versa?

README.md
```diff
@@ -207,6 +207,54 @@ docker-compose run app yarn lint:diff
...
+  //...
+  AWESOME_SAUCE: 'AWESOME_SAUCE',
+}
+```
+2. Check if the feature is enabled:
+```js
+// some code in a child folder of /api
+const Features = require('../features');
+
+if (Features.enabled(Features.Flags.AWESOME_SAUCE)) {
```
apburnes, line 231: For consistency, what are you thoughts on the api's `getFeatureFlags` keeping `FEATURE_` in the env variable like the webpack flags or vice versa?

README.md
davemcorwin: I def want to leave the `FEATURE_` part of the env var on the client so it's clear what the global is and so there are no potential collisions.

README.md
davemcorwin: I can update this to retain the `FEATURE_` prefix, i just removed it because it seemed repetitive with the extra stuff we have around it, but maybe keeping it would be clearer even if it's more verbose.

PR #2786, opened 2020-07-22
Cleanup frontend filenames and tests

davemcorwin:
- Rename all React files to have .jsx extension for consistency
- Fix tests
  - many of the tests were resulting in uncaught failed promises. These were actually failing tests but the tests were not written properly so the suite was not failing. I fixed the test syntax as well as updated the tests and potentially actual code to resolve the issues
  - fixed incorrect props in React tests
- Only run test coverage by default in CI
  - added new scripts for test processed ending in `:cover` and updated CircleCI config to use `test:cover`

Going forward, if a test results in an uncaught exception that is logged it should be treated as a FAILED test.

Actual changes:
- Dispatch the action and refetch the user from the api when a user is removed from a site (`frontend/actions/siteActions.js`)
- Show an alert to the user when fetching the user fails (`frontend/actions/userActions.js`)


davemcorwin:
The code climate stuff are false positives, but I'm seeing the locally as well even though I shouldn't. This may take a little more research but it seems like there should not be errors.

PR #2789, opened 2020-07-25
docker-db trust connect w/o user/pass

amirbey:


apburnes:
I never got this to work... I have to set a password variable.

amirbey:
really @apburnes ? i spoke to @davemcorwin and it sounded like I was having the same issue as you and this resolved it for me.  if i build the db with this version of docker-compose.yml it works ... did you try this version?

apburnes:
Tried it on the latest staging branch when doing admin-client work and still getting connection denied when migrating the db. I've just been setting a password config for development for my work around.

davemcorwin:
Not sure if I misspoke earlier, but using "trust" did NOT work for me when updating the db in the garden build docker setup either, https://github.com/18F/federalist-garden-build/blob/staging/docker-compose.yml

davemcorwin:
Updated postgres version in the db image and set a password. The key to upgrading is to blow away the docker volume containing the data so it starts from scratch. (Perhaps there is a way to  use `pg_upgrade` but since this is local dev data I don't see the point)

`docker volume ls` to view volumes
`docker volume rm <volume>` to remove

davemcorwin:
can't add @amirbey as a reviewer since he authored...

apburnes (approved):

PR #2790, opened 2020-07-25
implement proxy edge as feature

amirbey:


.cloudgov/manifest.yml
```diff
@@ -22,6 +22,7 @@ applications:
       - federalist-((env))-uev-key
       - federalist-deploy-user
       - federalist-site-wide-error
+      - federalist-((env))-dynamodb-creds
```
amirbey, line 25: federalist-production-dynamodb-creds has been added with empty credentials

PR #2803, opened 2020-07-28
Add admin auth

davemcorwin:
Add admin auth behind a feature flag

### DIFF
--- a/api/admin/index.js
+++ b/api/admin/index.js
@@ -4,22 +4,16 @@ const cors = require('cors');
 const methodOverride = require('method-override');
 const session = require('express-session');
 
-const sessionConfig = require('../init/sessionConfig');
 const { expressErrorLogger } = require('../../winston');
 const responses = require('../responses');
 
 const router = require('./routers');
 const passport = require('./passport');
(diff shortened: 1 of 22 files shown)


api/admin/index.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+// eslint-disable-next-line no-unused-vars
+function errorHandler(err, _req, res, _next) {
+  res.error(err);
+}
+
+const app = express();
+maybeAddCORS(app);
```
davemcorwin, line 33: The assumption here is that the client will be deployed to a subdomain of federalistapp.18f.gov, ex. `admin.federalistapp.18f.gov`.

api/admin/routers/auth.js
```diff
@@ -0,0 +1,19 @@
+const { Router } = require('express');
+const passport = require('../passport');
+
+const onSuccess = (req, res) => {
+  const script = `
+    <script>
+      window.opener.postMessage("success", "http://localhost:3000")
```
davemcorwin, line 7: This will need to be replaced by an env var with the client url OR find another way to get the opener in a secure way.

PR #2820, opened 2020-08-03
Whoops, double slash not good

davemcorwin:
A double slash at the beginning of the path was preventing logins...


api/services/passport.js
```diff
@@ -53,6 +53,7 @@ passport.logout = (req, res) => {
 passport.callback = (req, res) => {
   passport.authenticate('github')(req, res, () => {
     if (req.user) {
+      console.log({ session: req.session });
```
amirbey, line 56: is this intentional or for debugging?

amirbey, line 56:
```suggestion
```

PR #2842, opened 2020-08-15
Add site page to admin client

apburnes:
When an admin goes to a site page in the admin client, they will have access to all of the attribute data and builds for that site.

## Acceptance Criteria
- [x] Add 'site/:id' to page routes
- [x] Slim down sites list for easier viewing
- [x] View site attibutes
- [x] View site users
- [x] View builds


admin-client/src/components/LabeledItem.svelte
```diff
@@ -0,0 +1,13 @@
+<script>
+  export let label = 'label';
```
amirbey, line 2: does this value get redefined?

admin-client/src/components/LabeledItem.svelte
```diff
@@ -0,0 +1,13 @@
...
+  export let value = null;
```
amirbey, line 3: does this value get redefined?

admin-client/src/components/ExternalLink.svelte
```diff
@@ -24,7 +25,7 @@
 <a
   rel="noreferrer"
   target="_blank"
-  class="font-code-2xs padding-x-1 text-primary text-ls-neg-3 text-uppercase"
+  class="font-code-2xs text-primary text-ls-neg-3 text-uppercase {className}"
```
amirbey, line 28: does this value get redefined?

admin-client/src/components/UserList.svelte
```diff
@@ -0,0 +1,21 @@
...
+<SectionHeader>
+  Users
+</SectionHeader>
+<div class="grid-row grid-gap">
+  {#each users as user}
+    <div class="grid-col-12 tablet:grid-col-auto margin-y-05">
+      <div class="padding-x-1 padding-y-05 bg-gray-5">
+        <LabeledItem label="username" value={user.username} />
+        <LabeledItem label="email" value={user.email} />
+      </div>
```
amirbey, line 18:
```suggestion
        <LabeledItem label="signed in" value={user.signedInAt} />
      </div>
```

admin-client/src/components/LabeledItem.svelte
apburnes: Ya, that's the svelte way of creating a default value for a property.  Any variable in a component with the `export` syntax is a value you can pass to a component and the passed value will take precedent in the prop. 

ie
```svelte
// passing a value 
<LabeledItem label={'Hello'} value={'World'} />

// Shorthand version
<script>
  const label = 'Hello';
  const value = 'World';
</script>

<LabelItem {label} {value} />
```

admin-client/src/components/UserList.svelte
apburnes: 👍

admin-client/src/components/ExternalLink.svelte
apburnes: Yup. Playing with the idea that you can add additional classes to the anchor for customization.  Using it with the '<SiteCard />' to add padding to the center external link.

PR #2859, opened 2020-08-18
use FEATURE_PROXY_EDGE_LINKS env var in frontend

amirbey:


davemcorwin:
https://webpack.js.org/plugins/define-plugin/

PR #2863, opened 2020-08-19
Add ability to edit container config

davemcorwin:
- Edit container configuration in admin
- Fix notification/alert components so it disappears automatically
- Remove users list from sites page (up for debate)
- Add admin site fetch/update routes
- Update site serializer

### DIFF
--- a/admin-client/src/components/Alert.svelte
+++ b/admin-client/src/components/Alert.svelte
@@ -1,0 +1,38 @@
+<script>
+  import { onMount } from 'svelte';
+  import { fly } from 'svelte/transition';
+  import { notification } from '../stores';
+
+  import GridContainer from './GridContainer.svelte';
+  import GridRow from './GridRow.svelte';
+
+  export let alert;
+
+  onMount(() => {
+    setTimeout(() => { notification.clear(); }, 2000);
+  });
+</script>
+
+<div
+  transition:fly={{ y: 100, duration: 750 }}
+  class="position-fixed mobile-lg:margin-top-2 tablet:margin-top-4
+  width-full z-top">
+  <GridContainer>
+    <GridRow>
+      <div
+        class="grid-row usa-alert usa-alert--{alert.type}
+        usa-alert--slim width-full flex-row shadow-4">
+        <div class="grid-col flex-fill usa-alert__body">
+          <p class="usa-alert__text">{alert.message}</p>
+        </div>
+        <div class="grid-col flex-auto">
+          <button
+            class="usa-button usa-button--outline"
+            on:click|preventDefault={notification.clear}>
+            Close
+          </button>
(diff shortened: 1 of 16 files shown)


admin-client/src/components/Alert.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
```
davemcorwin, line 1: Moved this to it's own component so the `onMount` functionality actually fires correctly!

admin-client/src/components/SiteMetadata.svelte
```diff
@@ -3,7 +3,6 @@
   import ExternalLink from './ExternalLink.svelte';
```
davemcorwin, line 3: Sooo, i removed this bc I wasn't including the users in the query and it was causing problems. Trying to think of why we would want/need to see the users...

api/models/site.js
```diff
@@ -149,6 +149,24 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.STRING,
       allowNull: false,
     },
+    basicAuth: {
```
davemcorwin, line 152: apparently the `getterMethods` section is deprecated. Also, using setters allows the other code to interact with the site model in a more natural way, without needing to worry about these fields actually stored in the `config` field.

apburnes: 👍 👍 The virtual data type looks a lot cleaner and definitely makes interacting with `config` more natural.

api/serializers/site.js
```diff
@@ -1,47 +1,74 @@
...
+const viewLinks = {
+  demoViewLink: 'demo',
+  previewLink: 'preview',
+  viewLink: 'site',
+};
 
-  delete object.site_users__user_sites;
-  delete object.config; // may contain sensitive info ie: basicAuth
+// Eventually replace `serialize`
+function serializeNew(site, isAdmin = false) {
```
apburnes, line 40: For future context, maybe we're more semantic with `isAdmin` ie (`isPlatformAdmin`, `isFederalistAdmin`) so we're not to be confused with the repository admin.

apburnes: Though `isAdmin` might be obsolete soon if we let folks set container config. 🤷

davemcorwin: updating to `isSystemAdmin`

davemcorwin:
@amirbey If we want to test `siteToItem` then we should add a test for it, but that's not really what this particular test is trying to achieve. Looking at this now, a better approach may have been to stub out `siteToItem`. Either way, can we handle that improvement in an additional PR?

amirbey:
@davemcorwin - that's fine. it's already checked in the SiteToItem test so your right ... no need to verify here

PR #2877, opened 2020-08-26
generate site.subdomains without periods

amirbey:
subdomains with "." generate preview urls that are insecure and not valid for https wildcard domains


migrations/20200930102246-update-event-created-at-to-timestamp.js
```diff
@@ -2,6 +2,6 @@ exports.up = (db, callback) => db.changeColumn('event', "createdAt", {
   type: 'timestamp',
 }, callback);
 
-exports.down = (db, callback) => db.addColumn('event', "createdAt", {
+exports.down = (db, callback) => db.changeColumn('event', "createdAt", {
```
amirbey, line 5: good catch ... thanks

api/controllers/site.js
```diff
@@ -163,6 +163,7 @@ module.exports = {
         res.json(siteJSON);
       })
       .catch((err) => {
+        console.error(err)
```
amirbey: is  this for testing?

davemcorwin: I caught this one already!

api/utils/index.js
```diff
@@ -65,8 +65,31 @@ function generateS3ServiceName(owner, repository) {
   return slicedServiceName;
 }
 
+function toSubdomainPart(str) {
+  const characters = 'abcdefghijklmnopqrstuvwxyz';
+  let subdomain = str
+    // replace all invalid chars with '-'
+    .replace(/[^a-zA-Z-]+/g, '-')
```
davemcorwin, line 72: allow numbers

apburnes (approved):

davemcorwin (approved):

PR #2879, opened 2020-08-27
Update sqs message, add admin testing and cache control, update tests

davemcorwin:
Support configuring container in admin interface and passing in sqs message. Additional tests and refactoring, configure appropriate api caching.

I removed a `.only` from the test suite that was preventing certain tests from being run which revealed some failing tests. I addressed them, however, there is still one unhandled promise rejection that I haven't been able to track down.

Otherwise, I think this is gtg.


test/api/unit/services/BuildResolver.test.js
```diff
@@ -96,51 +96,43 @@ describe('BuildResolver', () => {
...
 
-    beforeEach(() => {
-      pValues = Promise.props({
-        user,
-        site,
-      });
-    });
+    it('calls out to github if the branch cannot be found locally', async () => {
+      const user = await factory.user();
+      const site = await factory.site({ users: Promise.resolve([user]) });
```
amirbey, line 102: should this be Promise.all([user]) ?

davemcorwin: Not in this case because we already `await`ed the promise in the line above.

davemcorwin: This expects the value of the `users`  key to be a promise that resolves to an array of users, both `Promise.all([userPromise])` and `Promise.resolve([user])` do this..

Issue #2835, opened 2020-08-11
admin interface can remove sites


amirbey:
#### User Story

An admin should be able to remove a site  from Federalist

#### Background (Optional)

<TODO>

#### Acceptance Criteria

- [ ] <TODO>
- [ ] Change made live via <TODO>.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

amirbey:
implemented

PR #2916, opened 2020-09-10
Admin site functionality to delete a site

apburnes:
The UI implementation to close out #2835 which allows admins to delete a site from the admin site page.

## Open questions to get your opinions
- Should the admin api see or search for soft deleted sites?
- If so, do we want to return the deleted site page?

## Acceptance Criteria
**Admin Client**
- [x] Adds a site delete form to the site page
- [x] Adds a verification modal to confirm you want to delete the site
- [x] Sends a delete method to `/site/{id}` to delete a site

**Backend API**
- [x] Add a route to the admin sites 
- [x] Add a controller to the admin sites route
- [x] Test delete request


apburnes:
I've got a couple open questions about UX that we can clarify, but its ready to go for the most part.

api/admin/controllers/site.js
```diff
@@ -69,4 +73,25 @@ module.exports = wrapHandlers({
...
+    try {
+      site = await fetchSiteById(id);
+      await S3SiteRemover.removeSite(site);
+      await S3SiteRemover.removeInfrastructure(site);
+
+      if (Features.enabled(Features.Flags.FEATURE_PROXY_EDGE_DYNAMO)) {
+        await ProxyDataSync.removeSite(site);
+      }
+
+      await site.destroy();
```
davemcorwin, line 90: Do all of the steps here need to happen serially, or could they happen concurrently?

api/admin/controllers/site.js
```diff
@@ -69,4 +73,25 @@ module.exports = wrapHandlers({
...
+      return res.json(serializeNew(site, true));
+    } catch (error) {
+      logger.error([`site@id=${site.id}`, error.message, error.stack].join('\n\n'));
```
davemcorwin, line 93: I'm not a huge fan of putting these messages in the logs since no one ever sees them. We need to find a better way of surfacing when these things happen...

api/admin/controllers/site.js
davemcorwin: also, this logic is identical to the logic in the user client, perhaps it should be refactored to a common service...

davemcorwin (commented):
Thanks @apburnes . Some thoughts:
1. I don't think we need to call out that we're "soft" deleting the site, this is what "delete" means in the context of the system, it's no different than when a user deletes a site. I think this also applies to noting that the infrastructure will be torn down, I'm not sure we need to call it out. Maybe the statement could be more generic and say something like "this remove the site and all associated infrastructure for all users".

2. If we want/need to see "deleted" sites, then maybe we should have an option on the Sites page to show them, since the query itself has to explicitly ask for them. I think when it is deleted, we should redirect to the sites page.

3. The delete confirmation modal is SLICK, but I think forcing the user to enter the repo name is overkill, maybe just a simple confirmation?

4. This failed to delete the site for me locally, but looks like there was an error specific to the local env.

additional comments below

api/admin/controllers/site.js
apburnes: 👍 Pulled out a `SiteDestroy` service

api/admin/controllers/site.js
apburnes: Put a `To Do` until we role out events.

apburnes:
Cleaned up the site delete flow but won't work locally unless you comment out the delete infrastructure (Maybe work on local dev feature flag 🤷).  Also simplified the delete verification with just a click.

PR #2929, opened 2020-09-15
refactor tests for federalistUsersHelper and allow federalist-users org owner to delete a site

amirbey:


api/authorizers/site.js
```diff
@@ -39,6 +40,23 @@ const authorizeAdmin = (user, site) => (
     })
 );
 
+const authorizeFederalistUsersAdmin = (user) =>
+  FederalistUsersHelper.federalistUsersAdmins(user.githubAccessToken)
+    .then((admins) => {
+      if (!admins.includes(user.username)) {
+        return Promise.reject({
```
davemcorwin, line 47: If we want the promise to reject just `throw`, there is no need to call `Promise.reject`. In general, when we `throw` we should only be throwing instances of `Error` (or subclasses), but if this works with just the object that's fine for now.

api/authorizers/site.js
```diff
@@ -39,6 +40,23 @@ const authorizeAdmin = (user, site) => (
...
+          message: siteErrors.ADMIN_ACCESS_REQUIRED,
+          status: 403,
+        });
+      }
+    })
+    .catch((error) => {
+      return Promise.reject({
```
davemcorwin, line 54: same here about `throw`ing.

davemcorwin (approved):
thanks @amirbey ! Approved with a couple comments below.

api/authorizers/site.js
```diff
@@ -18,10 +19,10 @@ const authorizeAdmin = (user, site) => (
...
     .then((permissions) => {
       if (!permissions.admin) {
-        return Promise.reject({
+        throw {
           message: siteErrors.ADMIN_ACCESS_REQUIRED,
           status: 403,
-        });
+        };
       }
       return Promise.resolve(site.id);
```
davemcorwin, line 27: same with the `Promise.resolve()`, it's not necessary, just return the value.

api/authorizers/site.js
```diff
@@ -32,13 +33,30 @@ const authorizeAdmin = (user, site) => (
...
         status: 403,
-      });
+      };
     })
 );
 
+const authorizeFederalistUsersAdmin = (user) =>
+  FederalistUsersHelper.federalistUsersAdmins(user.githubAccessToken)
+    .then((admins) => {
+      if (!admins.includes(user.username)) {
```
davemcorwin, line 46: not a big deal, but this is a bit repetitive, you could just throw `new Error()` here since it will be caught in the next `catch` and throw the error with the message there, or make a function?

davemcorwin (approved):
Other minor comments, still gtg.

api/authorizers/site.js
amirbey: removed! 👍

PR #2938, opened 2020-09-18
LINT ALL THE THINGS

davemcorwin:
- resolves ALL lint issues
- removes lint diff so we catch all new errors (prep your editors folks)

A couple opinions included on dealing with spread props for React components / dom and commenting out other things. Happy to discuss!!!

The update to the `ButtonLink` component results in a change to the ui in 2 places where we see actual buttons now instead of links for the "Remove user" and "Remove" buttons below:

Screen Shot 2020-09-18 at 2 44 05 PM
Screen Shot 2020-09-18 at 2 43 41 PM


apburnes (approved):
WOW. Doing all the dirty work today... As for the button link, it spices up the UI and is now more semantic. :kudos:

PR #2952, opened 2020-09-25
Actually cancel build tasks, refactor

davemcorwin:
Looks good locally, looking forward to testing in staging.

Currently looking for builds that have been `processing` for more than 45 mins or `tasked` for more than 3 minutes, but not looking at `queued` state. Thoughts on if and how to handle this? I don't necessarily want to cancel builds when there is a backup...

### DIFF
--- a/test/api/unit/services/TimeoutBuilds.test.js
+++ b/test/api/unit/services/TimeoutBuilds.test.js
@@ -4,42 +4,81 @@ const sinon = require('sinon');
 
 const factory = require('../../support/factory');
 const timeoutBuilds = require('../../../../api/services/TimeoutBuilds');
-const { Build } = require('../../../../api/models');
+const { Build, sequelize } = require('../../../../api/models');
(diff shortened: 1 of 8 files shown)


test/api/unit/services/TimeoutBuilds.test.js
```diff
@@ -4,42 +4,81 @@ const sinon = require('sinon');
...
-const { Build } = require('../../../../api/models');
+const { Build, sequelize } = require('../../../../api/models');
+const CFApi = require('../../../../api/utils/cfApiClient');
+
+const { Processing, Queued, Tasked } = Build.States;
+
+function setBuildUpdatedAt(build, date) {
+  return sequelize.query(
+    'UPDATE build set "updatedAt" = ? WHERE id = ?',
+    { replacements: [date, build.id] }
```
amirbey, line 15: 👍

amirbey (approved):

PR #2967, opened 2020-09-30
fix event createdAt column to be a timestamp and not a date

amirbey:


davemcorwin (approved):
Not a huge deal since we don't really rollback migrations, but if we did with this one, wouldn't it fail?

PR #2968, opened 2020-09-30
create logout event when the express response finishes

amirbey:
create logout event when the express response finishes. currently the redirect appears to fail


davemcorwin:
@amirbey hmmmm, can we touch base on this?

davemcorwin:
@amirbey can this be closed?

amirbey:
was not necessary .... the issue was the redirect location for the staging/dev env only

PR #2974, opened 2020-10-02
process github org webhooks and remove org members not using Federalist

amirbey:


davemcorwin:
@amirbey what is the purpose of this PR? Is there an issue?

amirbey:
@davemcorwin - https://github.com/18F/federalist/issues/3023

api/controllers/webhook.js
```diff
@@ -44,6 +50,37 @@ const findSiteForWebhookRequest = (request) => {
...
+    logger.warn(`Not a ${config.federalistUsers.orgName} membership action:\t${JSON.stringify(payload)}`);
+    return;
+  }
+
+  const { user: { login } } = membership;
+  const username = login.toLowerCase();
+  const user = await User.findOne({ where: { username } });
+
+  if (['member_added', 'member_removed', 'member_invited'].includes(action)) {
+    EventCreator.audit(Event.labels.FEDERALIST_USERS, user || User.build({ username }), payload);
```
davemcorwin, line 68: hmmm, do we care about this if there is no user? also `User.build` will only build a user in memory so there won't be an `id`...

api/controllers/webhook.js
```diff
@@ -44,6 +50,37 @@ const findSiteForWebhookRequest = (request) => {
     });
 };
 
+const organizationWebhookRequest = async (payload) => {
+  const {
+    action, membership, organization,
+  } = payload;
+  const { login: orgName } = organization;
+  if (orgName !== config.federalistUsers.orgName) {
+    logger.warn(`Not a ${config.federalistUsers.orgName} membership action:\t${JSON.stringify(payload)}`);
```
davemcorwin, line 59: Do really care about this? When will we try to find this in the logs?

api/controllers/webhook.js
```diff
@@ -122,10 +159,23 @@ module.exports = {
...
+        logger.error(err);
+        if (err.message) {
+          res.badRequest(err);
+        } else {
+          res.badRequest();
+        }
+      });
+  },
+  organization(req, res) {
+    signWebhookRequest(req)
```
davemcorwin, line 171: I'm assume the logic here was just copied from above, but this probably isn't the way we want to handle the response to github. If we can verify the signature and there is nothing wrong with the request, we should respond with a success (204). It looks like Github stores the response to the webhooks, I don't think we really want to return the model or a stack trace if there is an error, we handle this in other ways...

davemcorwin (commented):
Thanks @amirbey !

Have a bunch of questions and comments, happy to chat through them as well.

0. Something we might want to consider here is that folks could actively use Federalist without actually logging into the web application. I'm wondering if we should consider a github push from a user the same as a sign in...

1. Why are we using an `isActive` attribute instead of just (soft) deleting the user? Is it because we want to retain the relationships to the sites?

2. I'm not following why the `removeInactiveMembers` process is so complicated. What are we missing by just trying to remove all users where:
- isActive = true
- createdAt < cutoff
- signedIn = null or signedIn < cutoff
?

Looking at the webhook handling, maybe this is complicated by using `isActive` to track the `member_added` event as well, but I don't see the point. I thought all we care about is whether they have logged into to Federalist.

3. Can we use the name `username` instead of `login` where appropriate so it's a bit clearer what the value is?

4. Are we relying on receiving the webhook to mark the user as inactive even when we initiate the removal?

5. Do we need to create events in our system for the org events since they are already available (https://github.com/organizations/federalist-users/settings/audit-log) ? The only we may care about is when we actually update the user...

api/controllers/webhook.js
amirbey: good catch... yes ... copied and reused ... i will address this ... thanks

api/controllers/webhook.js
```diff
@@ -13,7 +16,10 @@ const findUserForWebhookRequest = (request) => {
 
   return User.findOrCreate({
     where: { username: username.toLowerCase() },
-    defaults: { username },
+    defaults: {
+      username,
+      active: false,
```
amirbey, line 21:
```suggestion
      isActive: false,
```

amirbey:
@davemcorwin 

    Something we might want to consider here is that folks could actively use Federalist without actually logging into the web application. I'm wondering if we should consider a github push from a user the same as a sign in...
    ** this is true ... how can a user actively use Federalist if they cannot log in? Federalist will build their commits if the repo is active and i think that is fine ***

    Why are we using an isActive attribute instead of just (soft) deleting the user? Is it because we want to retain the relationships to the sites?
    ** we are using is active to to track whether or not a user can authenticate into Federalist and whether they can be checked to be removed from Federalist users ... it also helps to minimize the payload from querying for users who can currently authenticate into the app **

    I'm not following why the removeInactiveMembers process is so complicated. What are we missing by just trying to remove all users where:

    isActive = true
    createdAt < cutoff
    signedIn = null or signedIn < cutoff
    ?
    ** pls elaborate on which part complicated. i don't want to assume but i'm guessing you are speaking to the queries around the "member_added" related stuff (if i'm wrong pls clarify or summarize how you think it should be simpler.)?  in addition to what you posted, I am trying to account for the users who have join the github org but who have never authenticated ... if this was not accounted for a user would have to login to Federalist right after joining the org or be removed from the org ***

Looking at the webhook handling, maybe this is complicated by using isActive to track the member_added event as well, but I don't see the point. I thought all we care about is whether they have logged into to Federalist.
**The user's isActive event is not altered in the webhook handling.  It is set to false which is the  model default.  In theory, I could remove that line as it serves no purpose.**

    Can we use the name username instead of login where appropriate so it's a bit clearer what the value is?
    **GitHub uses login and our system uses username.  I see how this could be confusing and am fine with just using one. However, using the name username and login allows the ability to identify where the value is coming from.  We could do that more explcitly if necessar** 

    Are we relying on receiving the webhook to mark the user as inactive even when we initiate the removal?
    **Yes we are relying on the webhook to remove the user from the federalist users group and mark them in active.  This is due to the scenariou that a federalist user org member has never logged into FEderalist but has builds and a user record in the database**

    Do we need to create events in our system for the org events since they are already available (https://github.com/organizations/federalist-users/settings/audit-log) ? The only we may care about is when we actually update the user...
    **The audit-log is dump which is compiled and evenetually downlaodable.  I do not see an easy way for hte app to use this dump. The webhook reports the information that we need to discern removing members from the federalist-users org**

api/controllers/webhook.js
```diff
@@ -13,7 +16,10 @@ const findUserForWebhookRequest = (request) => {
 
   return User.findOrCreate({
     where: { username: username.toLowerCase() },
-    defaults: { username },
+    defaults: {
+      username,
+      isActive: false,
```
amirbey, line 21:
```suggestion
```

api/controllers/webhook.js
amirbey: the EventCreator.audit expects a model.  This is User related data ... it is fine that it has no id

api/services/FederalistUsersHelper.js
```diff
@@ -51,7 +55,124 @@ const audit18F = ({ auditorUsername, fedUserTeams }) => {
...
+      },
+      returning: ['id', 'isActive'],
+    });
+  const users = activeUsers.concat(inActiveUsers);
+  users.map(user => EventCreator.audit(Event.labels.UPDATED, user, {
+    action: { isActive: user.isActive },
+  }));
+};
+
+const removeMember = (githubAccessToken, login) => GitHub
```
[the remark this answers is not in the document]
amirbey: let's do it!

davemcorwin (commented):
thanks @amirbey !!

1. Maybe the default value for `isActive` for a User should be `true`, then we wouldn't have to specify it everywhere. This might make the migration more annoying, but seems to be the "correct" approach going forward.
2. Aside from a one-time cleanup, why would we need to call `removeMembersWhoAreNotUsers`? Going forward, a user is created when a member is added and a user is inactivated when a member is removed so there shouldn't be any.
3. Same question with `refreshIsActiveUsers`, a user is active when created, they become inactivated when their membership is revoked.
4. I'm going to work on PR to further refactor the webhook handlers, but not essential for now

api/controllers/webhook.js
amirbey: The EvenCreator.audit has error handling but i can move it below

api/services/FederalistUsersHelper.js
```diff
@@ -1,10 +1,14 @@
...
 const GitHub = require('./GitHub');
 const { logger } = require('../../winston');
 const config = require('../../config');
+const EventCreator = require('./EventCreator');
+const { User, Event } = require('../models');
 
-const { User } = require('../models');
-
-const federalistOrg = 'federalist-users';
+const FEDERALIST_ORG = config.federalistUsers.orgName;
```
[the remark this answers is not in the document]
amirbey: Done!

scripts/sync-active-users.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-console */
+const { refreshIsActiveUsers } = require('../api/services/FederalistUsersHelper');
+
+refreshIsActiveUsers()
+  .catch((err) => {
+    console.error(err);
+    throw err;
```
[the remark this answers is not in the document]
amirbey: point taken ... done!

api/services/FederalistUsersHelper.js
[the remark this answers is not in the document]
amirbey: done 👍

api/models/event.js
```diff
@@ -10,6 +10,7 @@ module.exports = (sequelize, DataTypes) => {
     REMOVED: 'removed',
     UPDATED: 'updated',
     AUTHENTICATION: 'authentication',
+    FEDERALIST_USERS: 'federalist-users',
```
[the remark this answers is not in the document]
amirbey: Done

amirbey:
>     3\. refreshIsActiveUsers

I have removed "removeMembersWhoAreNotUsers" from the script remove-inactive-users b/c you are correct it *should not* need to be called more than once.  i had added it b/c i didn't see the harm only help (unlikely). removed ...

api/models/event.js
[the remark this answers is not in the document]
davemcorwin: did you mean to only update the key but not the value?

api/services/FederalistUsersHelper.js
```diff
@@ -51,7 +55,96 @@ const audit18F = ({ auditorUsername, fedUserTeams }) => {
...
+        [Op.lt]: cutoff.toDate(),
+      },
+      pushedAt: {
+        [Op.lt]: cutoff.toDate(),
+      },
+      createdAt: {
+        [Op.lt]: cutoff.toDate(),
+      },
+    },
+  });
```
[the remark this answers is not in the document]
davemcorwin: thoughts on this lil change?

migrations/20201001094246-add-is-active-to-user.js
```diff
@@ -0,0 +1,5 @@
+exports.up = (db, callback) => db.addColumn('user', 'isActive', { type: 'boolean', notNull: true, defaultValue: false })
```
davemcorwin, line 1: should the default value here be `true` to match the one in the Sequelize model?

api/controllers/webhook.js
```diff
@@ -44,6 +30,30 @@ const findSiteForWebhookRequest = (request) => {
...
+  } = payload;
+  const { login: orgName } = organization;
+
+  const { login } = membership.user;
+  const username = login.toLowerCase();
+  const user = await User.findOne({ where: { username } });
+
+  if (orgName === config.federalistUsers.orgName) {
+    if (['member_added', 'member_removed'].includes(action)) {
+      EventCreator.audit(Event.labels.FEDERALIST_USERS_MEMBERSHIP, user || User.build({ username }), payload);
```
davemcorwin, line 45: 1. I still don't understand the advantage of creating the event without the user, when we can do it with the user. Github already has an audit trail of these events, what our audit events are tracking are changes to the things INSIDE of our system and the expectation (at least my expectation) is that audit events will always tie back to a model instance, I actually thought we had put in a validation to require this, but now I don't see it.
For example, if we leave this as is and then query "everything that happened to user X", this event will not show up.

2. Also consider returning early here for readability and doing a bit less work, ex:
```
  if (orgName !== config.federalistUsers.orgName
      || !['member_added', 'member_removed'].includes(action)) {
    return;
  }

  const { login } = membership.user;
  const username = login.toLowerCase();
  const user = await User.findOne({ where: { username } });
  ...
```

api/controllers/webhook.js
```diff
@@ -84,50 +100,45 @@ const createBuildForWebhookRequest = async (request) => {
...
-          res.badRequest();
-        }
-      });
-  },
+  github: (req, res) => signWebhookRequest(req)
+    .then(() => {
+      if (req.body.commits && req.body.commits.length > 0) {
+        return createBuildForWebhookRequest(req);
+      }
+      return Promise.resolve();
```
davemcorwin, line 124: not necessary to return a "Promise", this is the same as `return undefined` but with more work.

davemcorwin (commented):
1. I asked this previously, but why do we need to run `removeMembersWhoAreNotUsers()` daily?
2. Other comments below

api/controllers/webhook.js
amirbey: linting rules require something to be returned

migrations/20201001094246-add-is-active-to-user.js
amirbey: i preferred letting the app handle making it true but i updated it to be false in DB ... my opinion isn't super strong on this

amirbey:
>     1. I asked this previously, but why do we need to run `removeMembersWhoAreNotUsers()` daily?
> 
>     2. Other comments below

@davemcorwin - post release we need to cleanse the federalist-users GitHub org of stale users ... because they are in the org does not mean they are users nor contributors

davemcorwin:
> > ```
> > 1. I asked this previously, but why do we need to run `removeMembersWhoAreNotUsers()` daily?
> > 
> > 2. Other comments below
> > ```
> 
> @davemcorwin - post release we need to cleanse the federalist-users GitHub org of stale users ... because they are in the org does not mean they are users nor contributors

ahh, my bad, it's not being run in Circle...

package.json
```diff
@@ -81,6 +81,8 @@
     "create-dev-data": "NODE_ENV=development node ./scripts/create-dev-data.js",
     "update-proxy-db": "node ./scripts/proxy-edge.js",
     "timeout-builds": "node ./scripts/timeout-builds.js",
+    "remove-inactivate-federalist-users": "node ./scripts/remove-inactive-federalist-users.js",
```
davemcorwin, line 84:
```suggestion
    "remove-inactive-federalist-users": "node ./scripts/remove-inactive-federalist-users.js",
```

.circleci/config.yml
```diff
@@ -329,6 +329,14 @@ workflows:
             branches:
               only: main
 
+      - run-task:
+          name: remove-inactive-federalist-users-production
+          app: federalistapp
+          task-name: remove-inactive-federalist-users
+          task-command: yarn remove-inactivate-federalist-users
```
davemcorwin, line 336:
```suggestion
          task-command: yarn remove-inactive-federalist-users
```

migrations/20201001094246-add-is-active-to-user.js
davemcorwin: I would think we would want the application enforcing the same defaults as the database, otherwise things could get confusing

api/controllers/webhook.js
amirbey: @davemcorwin 
1. i have switched to creating the event after the user exists ... no longer with empty user
2. done

davemcorwin (approved):

PR #3006, opened 2020-10-16
Dc/admin events

davemcorwin:


admin-client/src/components/Nav.svelte
```diff
@@ -89,6 +89,13 @@
             </a>
           {/if}
         </li>
+        <li class="usa-nav__primary-item">
+          {#if authenticated}
```
apburnes, line 93: Should we move this out to cover all of the authenticated links?

apburnes (commented):
Looks good.  Threw in a question about consolidating the nav's `authenticated` if statement.

admin-client/src/components/Nav.svelte
davemcorwin: yeah, looks like they are stacked from the right so it looks ok!

davemcorwin: done

apburnes (approved):
👍

PR #3019, opened 2020-10-21
Archive build logs daily

davemcorwin:
- task for archiving logs on a daily basis
- task for migrating historical logs
- update internal docs
- upgrade eslint-plugin-react for some reason

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,6 +291,7 @@ workflows:
               only:
                 - staging
                 - main
+
     jobs:
       - run-task:
           name: archive-build-logs-staging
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -295,23 +285,46 @@ workflows:
...
                 - staging
                 - main
+
     jobs:
-      - nightly-builds:
+      - run-task:
+          name: archive-build-logs-staging
+          app: federalistapp-staging
+          task-name: archive-build-logs-$(date "%Y-%m-%d")
+          task-command: yarn archive-build-logs-daily $(date "%Y-%m-%d")
```
davemcorwin, line 300: not sure if this date thing is gonig to work

scripts/archive-build-logs-daily.js
```diff
@@ -0,0 +1,56 @@
...
+          [Op.gte]: startDate.toDate(),
+          [Op.lt]: endDate.toDate(),
+        },
+      },
+    });
+
+    console.log(`\nFound ${builds.length} builds.`);
+
+    const { errors } = await PromisePool
+      .withConcurrency(5)
```
davemcorwin, line 33: Maybe we can handle more concurrent requests?

.circleci/config.yml
apburnes: You could just retrieve the date in the function during execution since its only run once a day for that day.

apburnes (approved):
🚀

scripts/archive-build-logs-daily.js
apburnes: Once it's up and running, we should have a better idea on tuning concurrency.

davemcorwin:
@apburnes yeah, we could add a default date if none provided , but i'm curious to see if this works.

PR #3104, opened 2020-11-25
store both webhook hash and clone hash per build

amirbey:
Each build in federalist should store the actual hash of the build when it was cloned.


api/services/GithubBuildStatusReporter.js
```diff
@@ -77,10 +77,10 @@ const reportBuildStatus = (build) => {
   let options = {};
 
   return new Promise((resolve, reject) => {
-    if (!build || !build.commitSha) {
-      reject(new Error('Build or commit sha undefined. Unable to report build status'));
-    } else {
+    if (build && (build.webhookCommitSha || build.cloneCommitSha)) {
```
davemcorwin, line 80: Is there a scenario when we would have `cloneCommitSha` without `webhookCommitSha` ?

api/services/GithubBuildStatusReporter.js
```diff
@@ -98,7 +98,7 @@ const reportBuildStatus = (build) => {
       options = {
         owner: site.owner,
         repo: site.repository,
-        sha: build.commitSha,
+        sha: build.cloneCommitSha || build.webhookCommitSha,
```
davemcorwin, line 101: Hmmm, how does this work right now? Does Github show the latest reported status OR the reported status for the latest commit included in the PR? (looking at the API docs, I wonder if they are using https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-combined-status-for-a-specific-reference which would seem to be just the latest) Before this change I could see this being a bit wonky in that we may report a status for a different (out of date) commit!

This seeeeeems fine?

Makes me think we *really* should build specific commits...

frontend/propTypes.js
```diff
@@ -34,7 +34,7 @@ export const BUILD = PropTypes.shape({
   state: PropTypes.string,
   error: PropTypes.string,
   branch: PropTypes.string,
-  commitSha: PropTypes.string,
+  webhookCommitSha: PropTypes.string,
```
davemcorwin, line 37: you're sending BOTH values to the frontend, right?

davemcorwin (commented):
Thanks @amirbey . I would have suggested NOT renaming the existing column just for the sake of churn/risk, but I also realize that even with the change, it's still not totally accurate since some of the builds are not triggered from webhooks. **IF** we are going to rename it, maybe something like `requestedCommitSha` would be more accurate, but I think it's safer to leave it as is and just create the new column.

api/services/GithubBuildStatusReporter.js
amirbey: point taken

test/api/requests/build.test.js
```diff
@@ -179,54 +183,19 @@ describe('Build API', () => {
           })
           .then((response) => {
             validateAgainstJSONSchema('POST', '/build', 200, response.body);
-            return Build.findOne({
+            return Build.findAll({
```
davemcorwin, line 186: hmmmm, I *think* `findOne` is the same things a `findAll` except it just returns only the first record...

davemcorwin (approved):
thanks @amirbey ! question about `findOne` and `findAll` below, but otherwise good!

test/api/requests/build.test.js
amirbey: changed

Issue #3114, opened 2020-11-30
Use GSA SecureAuth the authenticate non-prod and non-user facing (admin) sites


davemcorwin:
~- [ ] `Federalist` as admin - Production~
- [x] `Federalist` as admin - Staging
~- [ ] `Federalist Admin` as admin - Production~
- [x] `Federalist Admin` as admin - Staging
- [x] `Federalist` as user - Staging

Removing UAA as an "admin" option in the customer-facing app as all admin functionality will be moved to admin interface.

davemcorwin:
References
- https://cloud.gov/docs/management/leveraging-authentication/
- https://cloud.gov/docs/services/cloud-gov-identity-provider/

PR #3129, opened 2020-12-07
Use UAA for staging auth

davemcorwin:
Addresses #3114 

Allows authentication service for the Federalist web app to be configured to Github or SecureAuth. For now only admin users will be allowed to authenticate using SecureAuth. 

- Configures SecureAuth for staging and local environments. 
- `create-dev-data` has been updated to seed admin users for current Federalist operator github usernames.
- admin users have been manually seeded in staging and production
- `uaa-client` CF Service has been created, configured and bound in staging and production
- make sure to update your local credentials (see `local.sample.js`) with the `uaa-client-key-local` service key in `staging` to run locally.
- for a handful of the auth routes, the tests will only run the configured idp, i tested both locally, but this isn't feasible long term so we need to update to be able to test both

This will work for now, but is not a permanent solution as our users still require Github auth at some point. We will need to refactor to a "connection" based strategy for maintaining the relationship with Github.


api/services/passport.js
```diff
@@ -1,85 +1,92 @@
...
+    EventCreator.audit(Event.labels.AUTHENTICATION, user, { action: 'login' });
 
-passport.logout = (req, res) => {
-  const { user } = req;
-  req.logout();
-  if (user) {
-    const eventBody = { action: 'logout' };
-    EventCreator.audit(Event.labels.AUTHENTICATION, user, eventBody);
+    // should we wait for this?
+    RepositoryVerifier.verifyUserRepos(user);
```
davemcorwin, line 40: should we wait for this to finish?

api/services/passport.js
```diff
@@ -1,85 +1,92 @@
...
+
+    await user.update({
+      signedInAt: new Date(),
+    });
+
+    EventCreator.audit(Event.labels.AUTHENTICATION, user, { action: 'login' });
+
+    // should we wait for this?
+    // do i need to do this?
+    RepositoryVerifier.verifyUserRepos(user);
```
davemcorwin, line 66: should we wait for this to finish?

api/routers/auth.js
```diff
@@ -1,8 +1,35 @@
...
 
-router.get('/auth/github', AuthController.github);
-router.get('/auth/github/callback', AuthController.callback);
-router.get('/logout', AuthController.logout);
+const idp = config.env.authIDP;
+
+const opts = {
+  failureRedirect: '/',
+  failureFlash:
+    'Apologies; you don\'t have access to Federalist! Please contact the Federalist team if this is in error.',
```
amirbey, line 10:
```suggestion
    'Apologies; you are not authorized to access Federalist! Please contact the Federalist team if this is in error.',
```

api/services/passport.js
amirbey: i don't have a strong opinion on this but I lean towards not waiting.   i mainly see the benefit in waiting for a user who has not logged in for quite a while

api/services/passport.js
amirbey: i don't have a strong opinion on this but I lean towards not waiting. i mainly see the benefit in waiting for a user who has not logged in for quite a while

config/env/production.js
```diff
@@ -139,11 +139,13 @@ module.exports.userEnvVar = {
...
+  clientSecret: uaaCredentials.client_secret,
+};
 module.exports.passport = {
   uaa: {
-    options: {
-      clientID: uaaCredentials.client_id,
-      clientSecret: uaaCredentials.client_secret,
-    },
+    options: uaaOptions,
+    adminOptions: uaaOptions,
```
amirbey, line 149: options and adminOptions are temporarily the same?

amirbey (approved):
comments provided

config/env/production.js
davemcorwin: yes, the credentials are the same since we;re just using a single instance of the uaa service. The callback urls are different but those are configured in `./config/passport.js`

PR #3130, opened 2020-12-09
upload build federalist.json to proxy database

amirbey:
site_repo/federalist.json is currently used in the garden-build to set custom headers for files.  In the https://github.com/18F/federalist-proxy-edge custom headers and other branch specific functionality should occur in the the lambda and loaded via the database.


api/services/GithubBuildHelper.js
```diff
@@ -131,4 +131,19 @@ const reportBuildStatus = (build) => {
...
+  if (!build.clonedCommitSha) {
+    throw new Error(`Build or commit sha undefined. Unable to fetch ${path} for build@id=${build.id}`);
+  }
+
+  const site = await Site.findByPk(build.site);
+  if (!site) {
+    throw new Error(`Unable to find a site for build@id=${build.id}`);
+  }
+
+  const accessToken = await loadBuildUserAccessToken(build);
```
davemcorwin, line 144: Let's refactor `loadBuildUserAccessToken` so it doesn't have to query anything. We should have (or are able to get beforehand) the Site, Build and User...

api/services/GithubBuildHelper.js
```diff
@@ -131,4 +131,19 @@ const reportBuildStatus = (build) => {
     });
 };
 
-module.exports = { reportBuildStatus };
+const fetchContent = async (build, path = null) => {
```
davemcorwin, line 134: it looks like `path` is required so we shouldn't have a default argument

api/services/GitHub.js
```diff
@@ -280,4 +282,25 @@ module.exports = {
 
   getCollaborators: (accessToken, owner, repo) => githubClient(accessToken)
     .then(github => getNextCollaborators(github, owner, repo)),
+
+  getContent: async (accessToken, owner, repo, path, ref = null) => {
```
davemcorwin, line 286: 👍

test/api/requests/build.test.js
```diff
@@ -16,14 +19,16 @@ const requestedCommitSha = 'a172b66c31e19d456a448041a5b3c2a70c32d8b7';
 const clonedCommitSha = '7b8d23c07a2c3b5a140844a654d91e13c66b271a';
 
 describe('Build API', () => {
-  let sendMessageStub;
 
   beforeEach(() => {
-    sendMessageStub = stub(SQS, 'sendBuildMessage').returns(Promise.resolve());
+    process.env.FEATURE_PROXY_EDGE_DYNAMO = 'true';
+    sinon.stub(SQS, 'sendBuildMessage').returns(Promise.resolve());
```
davemcorwin, line 25: shouldn't this be `.resolves()` as well?

api/services/ProxyDataSync.js
```diff
@@ -1,10 +1,15 @@
...
+const url = require('url');
 const { DynamoDBDocumentHelper } = require('./DynamoDBDocumentHelper');
 const config = require('../../config');
 
-const tableName = config.app.proxySiteTable;
-const siteKey = 'Id';
+const TABLE_NAME = config.app.proxySiteTable;
+const SITE_KEY = 'Id';
 const getSiteKey = site => site.subdomain;
+const getBranchKey = (build) => {
```
davemcorwin, line 9: in terms of performance or simplicity in the proxy, would there be any value in using the `subdomain` and `path` or something that re-uses the key we use for the Site?

test/api/requests/build.test.js
```diff
@@ -530,34 +535,85 @@ describe('Build API', () => {
     });
 
 
-    it('should report the build\'s status back to github', (done) => {
+    it('should report the build\'s status back to github', async () => {
+      nock.cleanAll();
+      const statusNock = githubAPINocks.status({ status: 'pending', commitSha: clonedCommitSha });
+      const fetchContentStub = sinon.stub(GithubBuildHelper, 'fetchContent').resolves('{}');
+      const saveBuildStub = sinon.stub(ProxyDataSync, 'saveBuild').returns();
```
davemcorwin, line 542: should this (and references in other tests) be `.resolved()`?

test/api/requests/build.test.js
```diff
@@ -530,34 +535,85 @@ describe('Build API', () => {
     });
 
 
-    it('should report the build\'s status back to github', (done) => {
+    it('should report the build\'s status back to github', async () => {
+      nock.cleanAll();
```
davemcorwin, line 539: maybe we should just do this in the `beforeEach` instead of repeating it in every test

test/api/unit/services/GitHub.test.js
```diff
@@ -676,4 +676,77 @@ describe('GitHub', () => {
       expect(result).to.be.undefined;
     });
   });
+
+  describe('.getContent', () => {
+    /* eslint-disable camelcase */
+    it('returns a file based on the supplied parameters', async () => {
+      const accessToken = 'token';
```
davemcorwin, line 683: for the values that don't change from test to test, you can just define them once in the `describe` block...

test/api/unit/services/GithubBuildHelper.test.js
```diff
@@ -761,15 +761,48 @@ describe('GithubBuildStatusReporter', () => {
...
-          loggerSpy.restore();
           done();
         })
         .catch(done);
       });
     });
   });
+
+  describe('fetchContent(build, path)', () => {
+    it('should fetch the content requested', async() => {
```
davemcorwin, line 776:
```suggestion
    it('should fetch the content requested', async () => {
```

davemcorwin (commented):
awesome work @amirbey !! Some comments below, this is cool!

api/controllers/build.js
```diff
@@ -126,30 +139,31 @@ module.exports = {
...
+
+      await GithubBuildHelper.reportBuildStatus(build);
+
+      if (Features.enabled(Features.Flags.FEATURE_PROXY_EDGE_DYNAMO)) {
+        await saveBuildToProxy(build, buildStatus.status);
+      }
+
+      res.ok();
+    } catch (err) {
+      EventCreator.error(Event.labels.UPDATED, ['Error build status reporting to GitHub', err, err.stack].join('\n'));
```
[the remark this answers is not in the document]
amirbey: this controller function is `updating` the build based on the status ... perhaps a new label is more appropriate. i can think on it. any suggestions @davemcorwin

api/services/GithubBuildHelper.js
amirbey: nice catch 👍

api/services/ProxyDataSync.js
amirbey: as discussed ... for now, we will continue with build url origna nd path fur use by origin response

api/controllers/build.js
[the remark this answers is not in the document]
davemcorwin: I feel like `BUILD_STATUS` or something would be more appropriate, so maybe general, the "label" should represent the "subject" or "context" of the event 🤷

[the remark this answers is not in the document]
amirbey: agreed.... updated to match the other build_status logging events ... thanks

davemcorwin (approved):
thanks! @amirbey See the comments on the Event labels. Also, this doesn't have to be fixed now, but when handling the webhooks, our logic should be something like:
1. Check signature
  - if valid, respond `ok` and continue
  - if invalid respond with `badRequest` and halt
2. Do the rest of the stuff 

so we respond as soon as possible and only respond w an error if the actual request is bad, NOT is something that we are doing fails.

PR #3135, opened 2020-12-10
Remove slash from branch names in build URLs

amirbey:


amirbey:
to be replaced with double hyphen ('--') insted of single hyphen ('-')

davemcorwin (commented):
@amirbey I think we may need to do more research to ensure this won't cause problems. 
1) We always tell folks not to bookmark the preview urls, but sometimes they do, this could cause them to view a stale deployment
2) Netlify CMS creates branches like `cms/dailyMessages/dec-13-national-guard`, it's possible this is an assumption baked into the product so this might break this completely, this pattern might be configurable.

Is this only an issue with the Edge Proxy? If so, is there a more localized solution?

PR #3156, opened 2020-12-18
update socket.io and socket.io-client packages

amirbey:


api/socketIO.js
```diff
@@ -15,14 +15,15 @@ if (redisConfig) {
...
-  socketIO.adapter(redisAdapter({ pubClient, subClient }));
-
   pubClient.on('error', (err) => {
     logger.error(`redisAdapter pubClient error: ${err}`);
   });
   subClient.on('error', (err) => {
     logger.error(`redisAdapter subClient error: ${err}`);
   });
+
+  socketIO.adapter(redisAdapter({ pubClient, subClient }));
```
davemcorwin, line 25: does moving this have an effect?

davemcorwin (changes requested):
I think there is more work to do here: https://github.com/socketio/socket.io/releases/tag/3.0.0

> `use` has been removed....

api/socketIO.js
amirbey: not sure ... i was just assigning the error handling before attaching the client to the adapter

amirbey: it appaers socket.use is no longer used ... but io.use still appears to be used ... in this io is socketIO

davemcorwin (approved):
LGTM

PR #3160, opened 2020-12-23
log controllers via winston and EventCreator for controllers

amirbey:


davemcorwin (approved):
this is :100 with the understanding that we're going to refactor this to move the logging into the `EventCreator`

api/controllers/build.js
```diff
@@ -32,14 +30,26 @@ const emitBuildStatus = build => Site.findByPk(build.site)
...
 
 const saveBuildToProxy = async (build) => {
   const buildSettings = await GithubBuildHelper.fetchContent(build, BUILD_SETTINGS_FILE);
   if (buildSettings) {
     const settings = JSON.parse(buildSettings);
     await ProxyDataSync.saveBuild(build, settings);
-    EventCreator.audit(Event.labels.BUILD_STATUS, build, `${BUILD_SETTINGS_FILE} saved to proxy database`);
+    const body = {
+      filename: BUILD_SETTINGS_FILE,
+      content: settings,
```
davemcorwin, line 49: do we need to store the entire file here? Maybe just the commit sha if you're worried about being able to track this down later.

davemcorwin (commented):
thanks @amirbey !

1. If we're always passing an error to `EventCreator.error`, let's make this explicit by passing the error as an additional argument, then we can be sure to handle it in the same way every time.
2. Similarly, for `audit` and `error` events, if we're always adding a `message`, we should consider making this explicit.
3. I don't think that trying to identify and label EVERY type of error is feasible/necessary and I also see some of the same logic repeated. Perhaps errors that occur as part of the typical request/response lifecycle can be handled generically in our error handler, where we can always add the path, params, and body and we won't have to worry about missing any of them. If we need to add additional context to certain situations, we can just `throw` an error with a helpful message. Errors that occur OUTSIDE of this: webhook, tasks, socketio, etc... will have to be handled separately, but there may be analogous, semi-generic ways to do it.
4. This will also help with what I'm feeling are too many labels...
5. In general, I think we should be particular about WHAT metadata we need to store with the error/audit info and only store what we need. Ex. we don't need to store the entire contents of the federalist.json file in the audit event. If we need to track this down, we could just store the commit sha.


thoughts?

api/controllers/build.js
```diff
@@ -30,38 +30,41 @@ const emitBuildStatus = build => Site.findByPk(build.site)
...
-      .then(() => Build.findAll({
-        attributes: ['id'],
-        where: { site: site.id },
-        order: [['createdAt', 'desc']],
-        limit: 100,
-      }))
-      .then(builds => buildSerializer.serialize(builds))
-      .then(buildJSON => res.json(buildJSON))
-      .catch(res.error);
+module.exports = wrapHandlers({
```
apburnes, line 52: Could you use `wrapHandlers` function where the controllers are associated to the route?  Would that make them less nested and composable?

api/controllers/webhook.js
```diff
@@ -1,148 +1,22 @@
...
-    user = await User.findOne({ where: { username } });
-    if (user) {
-      await site.addUser(user);
-    }
-  }
-  if (user) {
-    await user.update({ pushedAt: new Date(pushedAt * 1000) });
-  }
+module.exports = {
+  ...wrapHandlers({
```
apburnes, line 5: Spreading the `wrapHandlers` here but the other controllers you're not.  Just wondering about convention and consistency.

apburnes (commented):
There's a lot to unpack but looks good. Had some questions on convention of wrapHandlers. Maybe worth running it over together. 👍 👍

api/controllers/webhook.js
davemcorwin: it's spread because I was exporting extra functions, but then i removed them, so I will clean this up.

api/controllers/build.js
davemcorwin: one could use `wrapHandlers` or `wrapHandler` in this way. For example:
```
router.post('/somepath', wrapHandler(Controller.action));
```
or
```
const Controller = wrapHandlers(require('./controller));
router.post('/somepath', Controller.action);
```

Express has built in functionality for handling errors, in that if you add an error handler via `app.use(someErrorHandler)` where `someErrorHandler` is a function with 4 arguments, when an error is thrown or  passed to `next(error)` in middleware, the error handlers will be called in order. However, this only works for synchronous controller actions. For async ones, we need to do a little more work and that is where `wrapHandler` comes in. All it does take an async function, add a `catch` to it and then call `next(error)`. Otherwise we would have to explicitly do this ourselves for each one with `catch` or `try/catch`.

Function definition:
```
function wrapHandler(fn) {
  // calls `next(error)`
  return (...args) => fn(...args).catch(args[2]);
}
```

This works because the third argument passed to the handler is `next`, we could make this more explicit:
```
function wrapHandler(fn) {
  return (req, res, next) => fn(req, res).catch(next);
}
```

davemcorwin: I updated the function definition as above.

api/controllers/webhook.js
apburnes: Copy that

api/controllers/build.js
apburnes: Sounds good.

apburnes (approved):
Let's let it fly

PR #3165, opened 2020-12-31
Move node modules and yarn cache to docker volume

davemcorwin:
Greatly increases yarn performance!!!!!!

This moves the yarn cache and node modules folders into volumes. They will persist across containers and can be managed explicitly via docker/docker-compose volume functions.

This means they will not longer appear on your local file system.

After pulling this down:
- `rm -rf .yarn-cache`
- `rm -rf node_modules`
- `rm -rf admin-client/node_modules`
- `docker-compose build`
- `docker-compose run --rm app yarn`
- `docker-compose run --rm admin-client yarn`


amirbey (approved):
Lgtm

PR #3202, opened 2021-01-27
Add users and user pages, enhance event filters

davemcorwin:


apburnes (approved):
💯 Looks good and functioning as expected.  Our seed data is pretty light.  Maybe worth putting in a task make development data more robust.

PR #3215, opened 2021-02-05
Just get full logs, user will have to refresh if in progress

davemcorwin:


apburnes (approved):
You fought the good fight

PR #3241, opened 2021-02-26
Make sure remove S3 is run after bucket contents completely removed

apburnes:
When deleting a site an error gets thrown because the cf service for the S3 bucket cannot be deleted because all of the contents are not removed yet.  Change the promises to make sure remove s3 content is finished before deleting bucket.


davemcorwin (approved):
ugh

PR #3242, opened 2021-02-26
3234 init organizations

### LINKED ISSUE #3234: Add initial Organization data model

davemcorwin:
Resolves #3234 

Initial data model and admin querying to support Organizations. This is not feature complete, I believe we will want to enhance this to be able to create/delete/manage organizations and associated users and sites, but wanted to get this in front of some additional eyes before continuing. Definitely some cleanup and standardization to be done on the front end as well.


apburnes (approved):
Seems like a good start and runs as expected.

PR #3249, opened 2021-03-02
Dc/refactor paginated

davemcorwin:
Establish some patterns for index pages and filtering. Make this generic and easy enough to build these pages and add querying options easily. In the current draft state, this is only implemented for the Users page (check out links from Orgs page as well). If the patterns here look good, will apply to other pages.


admin-client/src/pages/Users.svelte
```diff
@@ -1,49 +1,28 @@
...
   } from '../components';
-  
-  const limits = [
-    '10', '25', '50', '100',
-  ];
 
-  const defaultParams = {
-    limit: '25',
-    page: 1,
+  const fields = {
```
apburnes, line 8: Like the fields filter and it's portability between different models. We could add `role` for users here.

apburnes (approved):
Great start and it's a straight forward paradigm we can easily scale across our models. A `role` scope for users would be cool.

admin-client/src/pages/Users.svelte
davemcorwin: wellll, a `role`  is only defined for a user in the context of an organization, so it would require organization to be applied, might be more than I want to bite off right now :)

apburnes: 👍

apburnes (approved):
🚀

PR #3252, opened 2021-03-05
Add UAA Identity table to support authentication

apburnes:
To support our transition to using UAA and UA groups, use a uaa identity table and the uaa api to verify and save information for our users.

To Do:
- [x] Add uaa_identity table and one-to-one relationship to user
- [x] Add verification of uaa user's access to the pages group for users and admins
- [x] Create a local UAA dev env

### DIFF
--- a/api/models/uaa-identity.js
+++ b/api/models/uaa-identity.js
@@ -1,0 +1,57 @@
+const associate = ({ UAAIdentity, User }) => {
+  UAAIdentity.belongsTo(User, {
+    foreignKey: 'userId',
+    allowNull: false,
+  });
+};
+
+module.exports = (sequelize, DataTypes) => {
+  const UAAIdentity = sequelize.define('UAAIdentity', {
+    uaaId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+      unique: true,
+      validate: {
+        notEmpty: true,
+      },
+    },
+    userName: {
+      type: DataTypes.STRING,
+      unique: true,
+      allowNull: false,
+    },
+    email: {
+      type: DataTypes.STRING,
+      allowNull: false,
+      unique: true,
+      validate: {
+        isEmail: true,
+      },
+    },
+    origin: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    verified: {
+      type: DataTypes.BOOLEAN,
+      defaultValue: false,
+    },
+    accessToken: {
+      type: DataTypes.TEXT,
+    },
+    refreshToken: {
+      type: DataTypes.TEXT,
+    },
+    userId: {
+      type: DataTypes.INTEGER,
+      references: 'User',
+    },
+  }, {
+    paranoid: true,
+    tableName: 'uaa_identity',
+  });
+
+  UAAIdentity.associate = associate;
+
+  return UAAIdentity;
+};
--- a/api/models/uaa-identity.js
+++ b/api/models/uaa-identity.js
@@ -9,7 +9,6 @@ module.exports = (sequelize, DataTypes) => {
   const UAAIdentity = sequelize.define('UAAIdentity', {
     uaaId: {
       type: DataTypes.STRING,
-      allowNull: false,
       unique: true,
       validate: {
         notEmpty: true,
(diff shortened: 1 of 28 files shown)


api/utils/uaaClient.js
```diff
@@ -0,0 +1,51 @@
...
+    const { groups, origin, verified } = await this.request('GET', `/Users/${userId}`);
+    if (origin === 'cloud.gov' && !verified) {
+      return false;
+    }
+
+    return groups.filter(group => groupNames.indexOf(group.display) > -1).length > 0;
+  }
+
+  request(method, path, json) {
+    const host = config.env.uaaHost === 'http://localhost:9000'
```
apburnes, line 20: Want to find a cleaner way of setting this for the dev with docker network.

api/models/uaa-identity.js
```diff
@@ -0,0 +1,56 @@
...
+    allowNull: false,
+  });
+};
+
+module.exports = (sequelize, DataTypes) => {
+  const UAAIdentity = sequelize.define('UAAIdentity', {
+    uaaId: {
+      type: DataTypes.STRING,
+      unique: true,
+      validate: {
```
davemcorwin, line 13: I thought we'll have the uaa id when we create this record so can't we make this not nullable?

api/services/uaaStrategy.js
```diff
@@ -33,4 +35,34 @@ function createUAAStrategy(options, verify) {
...
+    return null;
+  }
+
+  const identity = await UAAIdentity.findOne({ where: { email } });
+
+  if (!identity) {
+    return null;
+  }
+
+  const user = await identity.getUser();
```
davemcorwin, line 54: maybe eager load the user when fetching the identity. Also, the user being null shouldn't be possible, we don't need to handle, just let it fail eventually since it would be a problem.

api/services/uaaStrategy.js
```diff
@@ -33,4 +35,34 @@ function createUAAStrategy(options, verify) {
...
+
+  if (!user) {
+    return null;
+  }
+
+  identity.uaaId = uaaId;
+  identity.accessToken = accessToken;
+  identity.refreshToken = refreshToken;
+  await identity.save();
```
davemcorwin, line 63: meh, maybe preference, but I think you can: 
```
await identity.update({ uaaId, accessToken, refreshToken });
```

api/utils/uaaClient.js
```diff
@@ -0,0 +1,48 @@
...
+    this.accessToken = accessToken;
+  }
+
+  async verifyUserGroup(userId, groupNames = []) {
+    const { groups, origin, verified } = await this.request('GET', `/Users/${userId}`);
+    if (origin === 'cloud.gov' && !verified) {
+      return false;
+    }
+
+    return groups.filter(group => groupNames.indexOf(group.display) > -1).length > 0;
```
davemcorwin, line 16: maybe the following would be more semantic:
```
return groups.some(group => groupNames.includes(group.display));
```

api/utils/uaaClient.js
davemcorwin: hmmmm, i don't understand the issue here, isn't this request always made from the server? It's the redirect uri that needs to say localhost. What am I missing?

davemcorwin (approved):
Awesome @apburnes . In general looks solid, a few questions, the big one about the nullability of the uaa id, otherwise gtg

api/models/uaa-identity.js
apburnes: I had it at non nullable but changed due to not wanting to create placeholder values in the local dev seed data. It think I'll add it back for integrity.

api/utils/uaaClient.js
apburnes: Currently, the UAA host env creates the redirect path for `authorizationURL` for the browser to navigate to and it needs to be the localhost while the server does the rest of the calls to the docker network. Looking at it with fresh eyes, I'll just update the `authorizationURL` to be local host via the local config and we should be good.

api/models/uaa-identity.js
apburnes: Updated

api/services/uaaStrategy.js
apburnes: Copy that.

api/utils/uaaClient.js
apburnes: Duly noted.

api/services/uaaStrategy.js
apburnes: Check one two.

PR #3263, opened 2021-03-09
Dc/concourse

davemcorwin:
- Migrates staging continuous integration to Concourse
- removes automatic db migrations after `yarn install`


ci/pipeline.yml
```diff
@@ -0,0 +1,492 @@
...
+          <<: *node-image
+        inputs: [name: src-api]
+        run:
+          dir: src-api
+          path: bash
+          args: [-c, yarn lint]
+
+    - task: test-api
+      privileged: true
+      config:
```
apburnes, line 60: Maybe worth splitting these larger tasks into separate files later on to get the pipeline file a bit smaller.

apburnes (approved):
Great start. The pipeline file is long and could be refactored but I think that's down the road as we start to work with it more and expand the jobs and tasks.

ci/pipeline.yml
davemcorwin: yes, there is a LOT of refactoring that can happen here...

PR #3284, opened 2021-04-01
Initial sites grouped per organization update

### LINKED ISSUE #3283: Update a user's view so their sites belong to their organizations
In order to support a user's sites that belong to their organizations, a user's sites page should display their sites based on the organizations they belong to and the sites that are apart of those organizations.

Acceptance Criteria:
- [ ] The `/sites` (maybe use `/organizations` to facilitate transition) page should show all of the user's organizations and each of the sites that belong to the organization.
- [ ] When a user that is an org manager create's a new site, make sure they add which organization should own the site.
- [ ] Allow users that are org managers and have admin privileges to the site's github repository, add the site to an organization
- [ ] Use a feature flag to only enable this view in the staging environment.

apburnes:
Closes #3283

Acceptance Criteria:
- [x] Create organization route for api and list user's organizations
- [x] Group sites by organization in `/sites` view
- [x] Associate site to organization when creating a new site

### DIFF
--- a/api/controllers/organization.js
+++ b/api/controllers/organization.js
@@ -1,0 +1,17 @@
+const organizationSerializer = require('../serializers/organization');
+const { User, Organization } = require('../models');
+const { wrapHandlers } = require('../utils');
+const { fetchModelById } = require('../utils/queryDatabase');
+
+module.exports = wrapHandlers({
+  async findAllForUser(req, res) {
+    const user = await fetchModelById(req.user.id, User, { include: [Organization] });
+
+    if (!user) {
+      return res.notFound();
+    }
+
+    const siteJSON = await organizationSerializer.serializeMany(user.Organizations);
+    return res.json(siteJSON);
+  },
+});
--- a/api/controllers/organization.js
+++ b/api/controllers/organization.js
@@ -10,7 +10,7 @@ module.exports = wrapHandlers({
       return res.notFound();
     }
 
-    const siteJSON = await organizationSerializer.serializeMany(organizations);
+    const siteJSON = organizationSerializer.serializeMany(organizations);
     return res.json(siteJSON);
   },
 });
--- a/api/controllers/organization.js
+++ b/api/controllers/organization.js
@@ -1,17 +1,16 @@
 const organizationSerializer = require('../serializers/organization');
-const { User, Organization } = require('../models');
+const { Organization } = require('../models');
 const { wrapHandlers } = require('../utils');
-const { fetchModelById } = require('../utils/queryDatabase');
 
 module.exports = wrapHandlers({
-  async findAllForUser(req, res) {
-    const user = await fetchModelById(req.user.id, User, { include: [Organization] });
+  async findAllForUser({ user }, res) {
+    const organizations = await Organization.forUser(user).findAll();
 
-    if (!user) {
+    if (!organizations) {
       return res.notFound();
     }
 
-    const siteJSON = await organizationSerializer.serializeMany(user.Organizations);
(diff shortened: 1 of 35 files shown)


api/controllers/organization.js
```diff
@@ -0,0 +1,17 @@
+const organizationSerializer = require('../serializers/organization');
+const { User, Organization } = require('../models');
+const { wrapHandlers } = require('../utils');
+const { fetchModelById } = require('../utils/queryDatabase');
+
+module.exports = wrapHandlers({
+  async findAllForUser(req, res) {
+    const user = await fetchModelById(req.user.id, User, { include: [Organization] });
```
davemcorwin, line 8: Not a big deal and I think this is lifted from other places, but semantically, we want to be querying for `Organization`s that belong to a particular user, similar to how that `forUser` scope is implemented on a `Site`:

```
  Site.addScope('forUser', user => ({
    include: [{
      model: User,
      required: true,
      where: {
        id: user.id,
      },
    }],
  }));
```

api/controllers/organization.js
```diff
@@ -0,0 +1,17 @@
...
+
+    if (!user) {
+      return res.notFound();
+    }
+
+    const siteJSON = await organizationSerializer.serializeMany(user.Organizations);
```
davemcorwin, line 14: `organizationSerializer.serializeMany` is not asynchronous

```suggestion
    const siteJSON = organizationSerializer.serializeMany(user.Organizations);
```

api/controllers/site.js
```diff
@@ -27,7 +29,9 @@ const stripCredentials = ({ username, password }) => {
 
 module.exports = wrapHandlers({
   async findAllForUser(req, res) {
-    const user = await fetchModelById(req.user.id, User, { include: [Site] });
+    const user = await fetchModelById(req.user.id, User, {
```
davemcorwin, line 32: same as my comment on querying the organizations, but I know this was already here.

davemcorwin (commented):
I'm looking at the attached image of the site list. Trying to thing about how folks will be looking for their sites, will they think org then site? just site?

What about just adding the organization to the site card instead of grouping and maybe allowing the user to filter by org if they want? Thinking that we can be more helpful with sort here too, looks like we're sorting by id, maybe either alphabetical or (with more work, last activity). Thoughts?

Screen Shot 2021-04-02 at 10 24 08 AM

apburnes:
I like it. Org filter on top and an org element in the site card.

apburnes:
An updated interface for sites page.
org-select

api/controllers/organization.js
apburnes: Scoped

api/controllers/site.js
apburnes: Are you thinking of creating a scope that gets sites based on user and organization relationships? Or just use a scoped query based on user?

davemcorwin: just based on user

frontend/components/siteList/siteListItem.jsx
```diff
@@ -41,6 +41,16 @@ const SiteListItem = ({ site, user }) => (
         </Link>
         {' '}
       </h4>
+      {
+        organization ? (
+          <h6>
+            <Link to={`/organizations/${organization.id}`} title="View organization settings">
```
davemcorwin, line 47: This is currently a link to a page that doesn't exist, maybe let's not make it a link for now and present any "org" settings differently for org managers

davemcorwin (approved):

PR #3291, opened 2021-04-08
We do not really need inquirer

davemcorwin:
We don't really need this, remove all the deps!


apburnes (approved):
Just take away the only question that makes me feel warm and fuzzy.

PR #3297, opened 2021-04-09
Clean up redux state to properly display org info when user has orgs

apburnes:
Fixes the UI to properly display organization related components when a user has associated organizations.

Acceptance Criteria:
- [x] Fix UI
- [x] Add component tests

### DIFF
--- a/frontend/components/AddSite/AddRepoSiteForm.jsx
+++ b/frontend/components/AddSite/AddRepoSiteForm.jsx
@@ -8,6 +8,7 @@ import UserOrgSelect from '../organization/UserOrgSelect';
 import SelectSiteEngine from '../SelectSiteEngine';
 import AlertBanner from '../alertBanner';
 import { hasOrgs } from '../../selectors/organization';
+import { validAddRepoSiteForm } from '../../util/validators';
 
 const showNewSiteAlert = () => {
   const message = (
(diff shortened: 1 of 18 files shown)


davemcorwin (changes requested):
Thanks @apburnes ! As I comb through this, I realize I may not have gone through the first PR that closely.

1. Whether or not a user "has orgs" seems more like a property of the user, then of just having no orgs, I wonder if we look at it this way it simplifies this a bit. Then we could:
  - add a selector for `hasOrgs` or `inOrg` and provide the result as a prop
  - leave everything that required an array of orgs as it was before, keep it required and as an array.

2. In the forms where they user has to choose an org, I don't think we should provide a default of the first org in their orgs, I think that if they have a single org, it's fine to pop it in, but if they have more than one, we should force them to choose, adding to the wrong org would be annoying, we still may have to provide functionality to move a repo between orgs, but I'd rather minimize the possibility.

3. The last comment may address this a bit but the `getOrgIdFromOrgData` function scares me a bit, can we be more explicit in these situations where we may or may not already have an org id? This runs the risk of just popping in an incorrect org id when creating a site. For example, I think in practice this would be fine, but knowing the possibility scares me:

```
  onCreateSiteSubmit({ repoUrl, engine, siteOrganizationId }) {
    const { orgData } = this.props;
    const { owner, repository } = getOwnerAndRepo(repoUrl);
    const organizationId = getOrgIdFromOrgData(siteOrganizationId, orgData);
    siteActions.addSite({
      owner, repository, engine, organizationId,
    });
  }
```

4. `useState`!!!! Missed this before. There are some funky semantics around the `useXXX` methods and I'm not sure if makes sense to start incorporating them into the codebase right now.

I am happy to make some of these changes if you don't have the bandwidth.

apburnes:
All the suggestions make sense to me other than not using the `useState` hook. The `siteList` is already a functional component and the hooks scope is limited to the use of a filtering items. Hooks are mature and we are using their minimum functionality.

apburnes:
Updated with a `hasOrgs` selector, add validation on org selection inputs, and added an orgless user to the dev db.

davemcorwin (approved):
Thanks @apburnes looks good!

Only thing I think we should add is when creating a site on the backend, enforcing that an organization is required for a site to be created if the user is in an org, like we do on the frontend, can def be a separate PR.

PR #3313, opened 2021-04-20
Add authorizer for site creation

### LINKED ISSUE #3310: Require organization to be apart of the API's validation on site creation when a user has organizations
When a user with organizations creates a new site, add validation to the API's create site route to make sure they have specified a site organization and they belong to the organization they specify or else return an error.

### Acceptance Criteria
- [x] Add authorization to the the create site endpoint to validate an organization is specified when creating a new site if the user has organizations.
- [x] Add authorization to the the create site endpoint to only allow user to create a new site and add it to an organization if they are apart of the organization.

apburnes:
Closes #3310


test/api/unit/authorizers/site.test.js
```diff
@@ -1,216 +1,272 @@
...
 const siteErrors = require('../../../../api/responses/siteErrors');
 const FederalistUsersHelper = require('../../../../api/services/FederalistUsersHelper');
 
-describe("Site authorizer", () => {
-  describe(".create(user, params)", () => {
-    it("should resolve", done => {
+describe('Site authorizer', () => {
+  describe('.create(user, params)', () => {
+    beforeEach(() => {
+      Promise.all([
```
davemcorwin, line 14: I think these should actually return the promise https://mochajs.org/#working-with-promises

test/api/unit/authorizers/site.test.js
```diff
@@ -1,216 +1,272 @@
...
+      const params = {
+        owner: crypto.randomBytes(3).toString('hex'),
+        repository: crypto.randomBytes(3).toString('hex'),
+        defaultBranch: 'main',
+        engine: 'jekyll',
+      };
+
+      await org.addUser(user, { through: { roleId: role.id } });
+
+      try {
```
davemcorwin, line 76: When using try/catch as here, if this doesn't actually throw an error, the test will still pass. For this to work you would have to do something like:
```
let error;
try {
  await authorizer.create(user, params);
} catch (e) {
  error = e;
}
expect(error.message).to.equal(siteErrors.ORGANIZATION_REQUIRED);
```

The pattern we have been using for async errors is:
```
const error = await authorizer.create(user, params).catch(e => e);
expect(error).to.be.an('Error'); // probably not 100% necessary
expect(error.message).to.equal(siteErrors.ORGANIZATION_REQUIRED);
```
where the extra check for an 'Error' addresses the wacky case where it resolves to something with a matching`message` property, which seems highly unlikely

test/api/unit/authorizers/site.test.js
```diff
@@ -1,216 +1,272 @@
...
+      const user = await factory.user();
+      const site = await factory.site({ users: Promise.all([user]) });
+      const expected = await authorizer.findOne(user, site);
+
+      return expect(expected).to.equal(site.id);
+    });
+
+    // eslint-disable-next-line consistent-return
+    it('should reject if the user is not associated with the site', async () => {
+      const user = await factory.user();
```
davemcorwin, line 120: could `Promise.all` here to speed it up! Obvi one of thees doesn't make a different, but in the aggregate it can.

api/authorizers/site.js
```diff
@@ -56,7 +56,26 @@ function authorizeFederalistUsersAdmin(user) {
 }
 
 // create is allowed for all
-const create = () => Promise.resolve();
+const create = async (user, siteParams) => {
+  const { organizationId } = siteParams;
+  const organizations = await Organization.forUser(user).findAll();
+
```
davemcorwin, line 62: definitely some personal preference but might be simpler to have
```
if (organizations.length === 0) return undefined;
```
here, then you don't have to keep checking it below.

davemcorwin (changes requested):
thanks @apburnes ! In general looks 💯 , but rejecting because the tests as written may provide false positives.

test/api/unit/authorizers/site.test.js
apburnes: I went down this route but since we are using `throw`, it doesn't return the error object.  To clean it up, I can use the `catch` appended to the async function like `await authorizer.create(...).catch(err => expect(err)...)`

api/authorizers/site.js
apburnes: If they are not apart of any organizations, should we also make sure they do not specify an `organizationId`? 

```
if (organization.length === 0 && !organizationId) return undefined;
```

davemcorwin (approved):

PR #3326, opened 2021-04-29
3264 admin create org

davemcorwin:
Adds the ability to create an organization along with an org manager from the admin interface. Since this was a bit in the weeds, I tried to be very explicit in naming variables and adding JSDoc comments that provide types and context.

Since we don't currently have the ability to send emails, the admin user will be presented with a link to send to the customer to authenticate with UAA.

This will def need some testing in staging and I think we really need to implement reset password flows (here or leveraging  UAA extras) before rolling this out.

Also:
- lots of general refactoring
- reverting to a single UAA client

Deployment Todos:
- update UAA client configuration
- update uaa client configuration in CF

@apburnes - the uaa config changes the scopes given to users, will we have to recreate those users?


davemcorwin:
Feedback from review:

Todo
- check in UAA for user first
- check on UAA group
- fix error messaging
- check fix does not redirect to app after sign up/accept invite

Future
- add Events
- save invite link
- expose invite user
- add/remove user to org

admin-client/src/pages/user/Show.svelte
```diff
@@ -0,0 +1,96 @@
...
+        <p>User does not have a UAA Identity</p>
+      {/if}
+    </div>
+
+    <h3>Github Info</h3>
+    <div class="grid-row">
+      {#if user.hasGithubAuth}
+        <div class="tablet:grid-col-fill padding-bottom-1">
+          <LabeledItem label="github email" value={user.email} />
+          <LabeledItem label="github username" value={user.uaa ? '' : user.username} />
```
amirbey, line 79: why empty string if user.uaa?

api/services/organization/Organization.js
```diff
@@ -0,0 +1,256 @@
...
+    let user = await this.findUserByUAAIdentity(targetUserEmail);
+
+    if (user) {
+      return [user];
+    }
+
+    const invite = await this.inviteUAAUser(currentUserUAAIdentity, targetUserEmail);
+
+    user = await User.findOne({
+      where: { username: targetUserGithubUsername && targetUserGithubUsername.toLowerCase() },
```
amirbey, line 104: all github usernames are inserted as lowercase ... curious, did you experience otherwise?

api/utils/uaaClient.js
```diff
@@ -1,43 +1,300 @@
...
+   *
+   * Utility
+   *
+   */
+
+  /**
+   * @param {[{display: string}]} userGroups - groups that the user belongs to
+   * @param {[string]} groupNames - allowed UAA group names, ex: ['pages.user']
+   */
+  userInGroup(userGroups, groupNames) {
```
amirbey, line 86: method name userInGroups (plural) would be helpful ... i had to do a double take ... 
or could have both userInGroup and userInGroups methods?

amirbey (approved):

admin-client/src/pages/user/Show.svelte
davemcorwin: good catch, this was from before we reverted to having this always be the Github username.

api/utils/uaaClient.js
davemcorwin: ha, yeah, I dunno. We're checking to see if the user is in at least one group of the list provided. I could def go either way.

api/services/organization/Organization.js
amirbey: doh! ... you are 💯  correct!

api/utils/uaaClient.js
amirbey: yeah ... userInAtLeastOneGroup is a bit wordy 😄 
i think right now it requires a dev to already know this function to use it

amirbey (approved):

davemcorwin:
thanks @amirbey . Merging will have to coincide with https://github.com/cloud-gov/cg-deploy-cf/pull/548

PR #3334, opened 2021-05-07
scheduled jobs via bull/workers

amirbey:


davemcorwin (commented):
Thanks @amirbey! I have some general feedback and there are specific comments below.

## Architecture
### Separation and Reusability
The Bull queues are going to be used in multiple places:
- web
  - add jobs to queues
  - monitor queues (maybe?)
- worker
  - add jobs to queues
  - process jobs from queues
  - monitor queues
- monitoring (ex. Bull Board)
  - monitor queues

So our architecture and how we breakdown responsibilies should reflect this, while taking into account the current state of affairs.

What this means in practice is that the creation of the queues (ie `new BullQueueClient()`), should be separate from adding callbacks or jobs to the queues. So perhaps having a `queues` file or folder that creates and exports the queues, that can be imported from each application and used as appropriate.

### Jobs and Error Handling
The previous implementation tried to separate the logic of actually "doing the thing" from the logic of reporting what happened, which is what the separation of concerns was between the service and the script. This PR moves some of that logic into the service and removes other logic that reported on the aggregate results. Borrowing from what I remember of using Sidekiq and Delayed Job in Rails apps, we often maintained this separation by creating a specific "Job" for each service that handled some of that logic that the scripts previously did and I'm wondering if we should take a similar approach. If we do this we can also keep the configuration with the job instead of having a centralized configuration which allows for easier reusabilty. Combined with the above changes, the startup code for the worker could look something like:

```
// api/worker/index.js
const { Queue1, Queue2 } = require('../queues');
const { job1, job2 } = require('../jobs');

Queue1.process(job1.process);
Queue1.add({}, job1.options);

Queue2.process(job2.process);
Queue2.add({}, job2.options);

[Queue1, Queue2].forEach(queue => {
  queue.on('completed', () => { /* do something */ });
  //...
});
//...
```

where a job looks something like:

```
// api/jobs/job1.js
const { REPEAT_NIGHTLY } = require('../queueUtil');
const doSomething = require('../services/SomeService');

const job = {
  options: { ...REPEAT_NIGHTLY },
  process: (data = {}) => doSomething(data),
};

module.exports = job;
```

### Queues
We don't need to have a 1-to-1 relationship from queue to processor and it might be simpler to run groups of jobs on shared queues. For example, maybe we run all the nightly jobs on one queue, and the more frequent stuff and/or realtime stuff on different queues. Both `queue.process` and `queue.add` support passing a name as the first argument AND we can configure the job defaults for the queue, so the above worker startup could look partially like:

```
// api/worker/index.js
const { NightlyQueue } = require('../queues');
const { nightlyBuilds, removeInactiveFederalistUsers } = require('../jobs');

[nightlyBuilds, removeInactiveFederalistUsers].forEach(({ name, process }) => {
  NightlyQueue.process(name, process);
  NightlyQueue.add(name, {});
});

//...
```

where we have specified the cron arguments when creating the queue.

This may make sense when controlling for concurrency, error handling or other configuration.

Ref:
- https://github.com/OptimalBits/bull/blob/v3.22.5/REFERENCE.md#queue
- https://github.com/OptimalBits/bull/blob/v3.22.5/REFERENCE.md#queueprocess

## Error Handling
For jobs that aggregate tasks for many sites/users we need to think about how we want to track success/failure. Does it make sense to create an event for each item, or for the aggregate result with details? What is the better way to track this? Bull Board? Our Events? Something else? We also may want to look at making this error handling generic enough to pput in the worker startup or in the jobs as opposed to in the services. For example, I may want to run one of the services as a one-off, but I may not to create an Event for it.

PR #3348, opened 2021-05-11
Always return user attributes when finding or inviting uaa user

### LINKED ISSUE #3347: when creating an organization check to see if org managers is in uaa before generating uaa invite
#### User Story

BUG - When inviting a user to the platform, if the user exists in UAA and an invite is not generated, but the user DOES NOT exist in Federalist, `findOrCreateUser` fails since it needs the user attributes to create the user in Federalist.

This is required to NOT fail so admins can invite users to the platform and create organizations.

#### Background (Optional)

https://github.com/18F/federalist/blob/93864efe8da756281ce08967995897bdef34911a/api/services/organization/Organization.js#L100

Currently, when creating an organization the organization service attempts to invite user to uaa regardless of whether they are already in uaa. if the uaaClient finds they are already in uaa, then it returns null and does not have an email (invite) to create a new user.

#### Level of effort - Low

davemcorwin:
Resolves #3347 

This seemed to easiest fix for now.

### DIFF
--- a/api/admin/controllers/organization.js
+++ b/api/admin/controllers/organization.js
@@ -41,12 +41,12 @@ module.exports = wrapHandlers({
       user,
     } = req;
 
-    const [org, invite] = await OrganizationService.createOrganization(
+    const [org, uaaUserAttributes] = await OrganizationService.createOrganization(
       user, name, managerUAAEmail, managerGithubUsername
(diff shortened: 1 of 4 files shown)


amirbey (approved):
LGTM

PR #3354, opened 2021-05-12
Refactor usage of site serializer, authorizer and site querying

### LINKED ISSUE #3353: Do not require github connection to view sites within an org. Also potentially refactor site serializer.

davemcorwin:
Resolves #3353 

This is NOT a full refactor of all site serialization but achieves most, if not all, of the pressing goals. There is definitely duplicate querying when using the authorizor, but I don't think this is required to address immediately. 

There will still probably be cases this is missing.


apburnes:
This breaks the data structure for the frontend's query for sites' organization. Looking at it now

api/controllers/site.js
```diff
@@ -27,41 +29,45 @@ const stripCredentials = ({ username, password }) => {
...
+    const { user } = req;
+
+    const sites = await Site.forUser(user).findAll();
 
     if (!sites) {
       return res.notFound();
     }
 
-    const siteJSON = await siteSerializer.serialize(sites);
+    const siteJSON = siteSerializer.serializeMany(sites);
```
apburnes, line 40: We can update `serializeMany` to use `serializeObject` instead of `serializeNew` because `serializeObject` runs `serializeNew` and then cleans up the included `User` and `Organization` related keys.

apburnes: Probably should change the name to something like `serializeIncluded`. I'll work on it and get you alls take.

davemcorwin:
this looks pretty good to me with default data

apburnes (approved):

PR #3355, opened 2021-05-12
Update serializeMany with serializeObject

apburnes:
Use's the `serializeObject` function in 'serializeMany`.


davemcorwin (approved):
I'm guessing the reason I didn't do this in the first place was opinions around including the Users and Organizations in the response to begin with...

PR #3429, opened 2021-06-02
Add a helpful Makefile

davemcorwin:
OK! Thoughts on this? We can def expand on this, but I added the commands that I commonly use. If we like, we can also update the readme to reference

Ex. Running `make start` will fire up the servers and `make install` will install npm deps for all apps.

Running `make` or `make help` outputs the possible commands.

Screen Shot 2021-06-02 at 3 17 37 PM


amirbey (approved):
👍

PR #3497, opened 2021-07-16
Update the live URL link to documentation

apburnes:
Fix the "live URL" link to the documentation when setting up a custom domain for a site.


davemcorwin (approved):
💯

PR #3498, opened 2021-07-16
Update form-action and frame-ancestors csp directives

### LINKED ISSUE #3496: Close SAR finding for CSP: Wildcard Directive
#### User Story

Address this finding:

>The following directives either allow wildcard sources (or ancestors), are not defined, or are overly broadly defined:
> * frame-ancestors, form-action
>
>The directive(s): frame-ancestors, form-action are among the directives that do not fallback to default-src, missing/excluding them is the same as allowing anything.

For this evidence:

URL | https://federalistapp-staging.18f.gov
-- | --
Method | GET
Parameter | Content-Security-Policy
Evidence | script-src  'self' www.googletagmanager.com dap.digitalgov.gov  www.google-analytics.com  'nonce-f53ec67d715a724082877073ec2a51a2';connect-src 'self'  www.google-analytics.com;default-src 'self';base-uri  'self';block-all-mixed-content;font-src 'self' https:  data:;frame-ancestors 'self';img-src 'self' data:;object-src  'none';script-src-attr 'none';style-src 'self' https:  'unsafe-inline';upgrade-insecure-requests

URL | https://federalistapp-staging.18f.gov/images/logos
-- | --
Method | GET
Parameter | Content-Security-Policy
Evidence | default-src 'none'

It looks like the first can be addressed in `helmet.js`, I don't know if this is a false positive for logos etc.

#### Acceptance Criteria

- Clean ZAP scan

- [ ] <TODO>

davemcorwin:
Resolves #3496


config/helmet.js
```diff
@@ -18,6 +18,12 @@ module.exports = {
         "'self'",
         'www.google-analytics.com',
       ],
+      'frame-ancestors': [
```
pburkholder, line 21: I don't think we need `frame-ancestors`. The finding is that either `frame-ancestors` OR `form-action` is missing. But we seem to have `frame-ancestors` covered by the defaults (per https://helmetjs.github.io/) and scan result looks ok for the pages that _have_ included helmet:

image

But I don't know why some other number of pages don't seem to picked up any of this helmet config (e.g. https://federalistapp-staging.18f.gov/images/logos).

davemcorwin: @pburkholder it's not a big to narrow `frame-ancestors` down, we don't want to IFramed so changing it to `'none'` is fine.

amirbey (approved):
LGTM

PR #3500, opened 2021-07-19
do not run mailer if not configured

davemcorwin:
Prevents the mailer from running in the worker application when the SMTP HOST is set to 'NA'. Currently, this fails the entire worker app because this is not configured for production AND even with configuration this will fail because the `production` space does not have access to postfix.


onelittlebecca (approved):
LGTM

PR #3504, opened 2021-07-19
Release 7-19-2021

davemcorwin:
- Use default of npm 7 with node 16
- fix crashing worker app in production Federalist by not instantiating Mailer if configured to be 'NA'


onelittlebecca (approved):
LGTM

PR #3510, opened 2021-07-26
Dc/fun refactoring

davemcorwin:
## Background
Soooo, in admin staging I noticed a bunch of JSON parsing errors for routes that don't exist and wondered to myself why we were even trying to parse the JSON for these. Well, it is because middleware in express is executed in the order in which it is added and we were adding this BEFORE the router and also before authentication. This got me started in looking into the which middleware we are using where to see if it makes sense.

### Remove `method-override` middleware
This is only necessary if the api is accepting requests from clients that don't support PUT, DELETE, etc... which would only be the case if we were handling HTML form requests, which we are not, so, bye bye.

### Remove `body-parser` `urlencoded` middleware
See previous explanation.

### Replace `body-parser` `json` middleware with `express.json`
This is now included with `express` so there is no need to pull in an additional package.

### Move usage of `express.json` middleware (Admin only)
Per the background above, we shouldn't parse the json until we actually need to, and for most routes (GETs) we don't even need to try. At least for the admin routes, this is added to the middleware for each route where necessary. I did NOT make this change yet in client api as its a larger change and wanted to make sure there aren't issues and this makes sense to everyone.

### Additional refactoring/reorganization
Just because.


apburnes (approved):
Like the re-org of the routes to sub directories and being more explicit with the middleware.

PR #3538, opened 2021-08-05
Timeout builds script for circle cli

amirbey:
- Specify staging and main to run timeout builds against staging and production respectively
- add a timeout script in package.json and in scripts


davemcorwin:
@amirbey I had already pushed up this PR yesterday... https://github.com/18F/federalist/pull/3537

PR #3539, opened 2021-08-06
separate prod and staging restage into different jobs

amirbey:
restage in staging is failing b/c of invalid app names


amirbey:
https://github.com/18F/federalist/pull/3542

PR #3540, opened 2021-08-09
Add better error message for 401

dylanirlbeck:
If we can't modify the Federalist OAuth to be less restrictive, the next best solution, in my view, is to give folks a better error message. Specifically, I'd like for it to be more clear where you need to go to get access to Federalist services. 

Open to suggestions on how to word this!


dylanirlbeck:
@davemcorwin Can you take a look at this PR?

amirbey (approved):

PR #3575, opened 2021-08-25
Add origin restriction and csrf protection to admin api, refactoring

### LINKED ISSUE #3531: Ensure CSRF is used everywhere it should be, including admin.
#### Background (Optional)

Checking the `Origin` header along with cookies and CORS  may be sufficient for the admin.

https://medium.com/tresorit-engineering/modern-csrf-mitigation-in-single-page-applications-695bcb538eec

#### Acceptance Criteria

- [ ] <TODO>
- [ ] Change made live via <TODO>.

- [ ] <TODO>

davemcorwin:
Resolves #3531 

- add `ensureOrigin` middleware to admin api routes. While the CORS policy restricts cross-origin requests to the specified domain, the `ensureOrigin` middleware prevents same origin requests which we also want to prevent.
- add csrf middleware to admin api routes. This is probably overkill with the above and CORS but 🤷 . This is implemented by returning the expected csrf token with the response for the user endpoint at which point the client is already authenticated and is coming from an acceptable origin.
- refactor all middlewares to the same place and naming convention


amirbey (approved):
LGTM

Issue #3553, opened 2021-08-13
error out a build in the admin


amirbey:
#### User Story

<TODO>

#### Background (Optional)

An admin can error out a stuck build in the admin ui

#### Acceptance Criteria

- [x] add cancellation button to the builds table in the admin ui
- [x] build status become "error" after cancellation button exists
- [x] Change made live via deploy of Federalist

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3588, opened 2021-08-31
error out build from admin client

amirbey:
Allow admins to error out a stuck build from the admin client.

Reference: https://github.com/18F/federalist/issues/3553


davemcorwin (approved):
LGTM!

PR #3620, opened 2021-09-22
Run migrations after deployment

### LINKED ISSUE #3545: run migrations at deployment
#### User Story

<TODO>

#### Background (Optional)

Currently, migrations are not running during deployment and require an operator to run them manually post deployment.  Ideally, the deployment process should run all migrations and not require manual processing.

#### Acceptance Criteria

- [ ] identify a means to run backwards compatible migrations during deployment
- [ ] implement running of backwards compatible migrations at time of deployment
- [ ] Change made live via Federalist

- [ ] <TODO>

davemcorwin:
Resolves #3545 

Run migrations after deployment in staging and production Federalist. This is only implemented for Federalist in Circle bc we don't want them to run twice for the same database in case any migration is not idempotent. It's possible this could lead some issues if Pages deploys first or not at all, but hey, that's the fun with running 2 CI systems simultaneously.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,7 +161,6 @@ jobs:
             # Attempt to submit a report, but don't fail the build if this fails (`|| true`)
             ./cc-test-reporter upload-coverage || true
 
-
   deploy:
     docker:
       - image: circleci/node:16.4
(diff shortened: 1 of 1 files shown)


apburnes (approved):
🏃

PR #3621, opened 2021-09-22
Remove SOCKET_HOST env var and use APP_HOSTNAME

davemcorwin:
These are the same, no need for both and we should reduce the amount of data that the server renders for the React App so these can eventually be decoupled.


apburnes (approved):
✂️

Issue #3625, opened 2021-09-27
update federalist app robots.txt to deny-all


amirbey:
#### Background (Optional)

There are currently no previews at the app /preview; therefore, the robots txt should remove previews.

#### Acceptance Criteria

- [x] update robots.txt to deny-all
- [x] Change made live via deploy of Federalist

_After evaluating, edit this part:_

PR #3635, opened 2021-09-29
robots deny all

amirbey:
There is no preview path for this application.  Robots should be denied for all paths.

For reference: https://github.com/18F/federalist/issues/3625


davemcorwin (approved):
We could prob just move this to a static file to make it easier, but this does the trick! Thanks!

PR #3637, opened 2021-09-29
Update issue templates

amirbey:


amirbey:
will be inherited by org settings when repo is migrated

PR #3639, opened 2021-09-29
prevent helmetjs from overriding apex hsts

amirbey:
helmet js is overiding the federlistapp.18f.gov hsts headers set for 18f.gov  ... disable helmetjs hsts to stop this


davemcorwin (approved):
cloud.gov router adding hsts?

amirbey:
> cloud.gov router adding hsts?

proxy server is adding it for cloud.gov

Issue #3627, opened 2021-09-27
disable OPTIONS Method in the API router


amirbey:
#### Background (Optional)

The Federalist app should not provide a server error (404) when receiving a options method request.

#### Acceptance Criteria

- [x] a 404 is returned when making an OPTIONS request to the web app
- [ ] Change made live via deploy of Federalist

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3640, opened 2021-09-30
return 404 for options requests

amirbey:
HTTP OPTIONS request return a 404 instead of displaying options

Reference: https://github.com/18F/federalist/issues/3627


davemcorwin (approved):
I don't think this should interfere with routes that actually need it. I'm confused as to why this doesn't already happen though...

Issue #2882, opened 2020-08-31
Add default value for `repoLastVerified`


davemcorwin:
#### User Story

The create date for a site make sense for the default value of `repoLastVerified` and reduces the need for clients to introduce extra logic

#### Background (Optional)

<TODO>

#### Acceptance Criteria

- [ ] <TODO>
- [ ] Change made live via <TODO>.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3643, opened 2021-10-04
site.repoLastVerified has default value

amirbey:
provide default value for a site repoLastVerified

### DIFF
--- a/frontend/components/siteList/repoLastVerified.jsx
+++ b/frontend/components/siteList/repoLastVerified.jsx
@@ -1,22 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+import moment from 'moment';
 import { dateAndTime } from '../../util/datetime';
 
 const getRepoLastVerified = (site) => {
-  let msg = 'Repository not found';
-  if (site.repoLastVerified) {
-    const formattedBuildTime = dateAndTime(site.repoLastVerified);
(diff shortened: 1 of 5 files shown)


davemcorwin (commented):
Was there an issue/ bug?

amirbey:
Issue: https://github.com/18F/federalist/issues/2882

frontend/components/siteList/repoLastVerified.jsx
```diff
@@ -1,22 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+import moment from 'moment';
```
davemcorwin, line 3: we don't use `moment` anywhere else in the client and this is a rather large dependency, can we accomplish this without it either by computing it another way or doing it in the backend?

davemcorwin (commented):
Thanks @amirbey ! I left a comment about using `moment` below. Also, if you add something like `Resolves #<issue>` it will link the PR to the issue, but only specific words work. (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

frontend/components/siteList/repoLastVerified.jsx
amirbey: moment js addition reverted 👍🏿

davemcorwin (approved):

Issue #3671, opened 2021-10-21
admin allow setting buildStatus attribute for a site


amirbey:
#### User Story

An admin user should be able to set the buildStatus of a build

#### Background (Optional)

site.buildStatus = {'active','inactive'}

#### Acceptance Criteria

- [ ] buildStatus can be set in admin interface
- [ ] Change made live via deploy of Federalist

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3676, opened 2021-11-02
deactivate a site

amirbey:
- rename site.buildstatus to site.isactive
- prevent users from updating/rebuilding a site
- display the site is inactive in ui
- allow updating site status in the admin ui

resolves https://github.com/18F/federalist/issues/3671

### DIFF
--- a/api/services/Webhooks.js
+++ b/api/services/Webhooks.js
@@ -17,7 +17,7 @@ const findSiteForWebhookRequest = (payload) => {
   });
 };
 
-const shouldBuildForSite = site => site && site.isActive
+const shouldBuildForSite = site => site?.isActive
   && (!site.Organization || site.Organization.isActive);
 
 const organizationWebhookRequest = async (payload) => {
--- a/api/services/Webhooks.js
+++ b/api/services/Webhooks.js
@@ -17,7 +17,7 @@ const findSiteForWebhookRequest = (payload) => {
   });
 };
 
-const shouldBuildForSite = site => site && site.buildStatus !== 'inactive'
+const shouldBuildForSite = site => site && site.isActive
   && (!site.Organization || site.Organization.isActive);
 
 const organizationWebhookRequest = async (payload) => {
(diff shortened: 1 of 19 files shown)


davemcorwin:
@amirbey what is the advantage to renaming the `buildState` column and changing it to a boolean?

amirbey:
> @amirbey what is the advantage to renaming the `buildState` column and changing it to a boolean?

 the current prod implementation only ignores webhooks ... the intent is to not only not build the site from push webhooks  but also make the site inaccessible in the app same as organizations that are deactivated.
the field wasn't being used at all so there was not real risk in changing it ... seems more accurate

davemcorwin:
> the current prod implementation only ignores webhooks ... the intent is to not only not build the site from push webhooks but also make the site inaccessible in the app same as organizations that are deactivated. 

renaming/changing to boolean doesn't affect our ability to do this.

> the field wasn't being used at all so there was not real risk in changing it ... seems more accurate

yes, but how does this help?

amirbey:
> > the current prod implementation only ignores webhooks ... the intent is to not only not build the site from push webhooks but also make the site inaccessible in the app same as organizations that are deactivated.
> 
> renaming/changing to boolean doesn't affect our ability to do this.
> 
> > the field wasn't being used at all so there was not real risk in changing it ... seems more accurate
> 
> yes, but how does this help?

i agree ... it does not affect ability to do this .. 👍🏿 
while going forward,  the field would be doing more than deactivating builds and mimicing the behavior if organizations.isActive, is there detriment to renaming it to to something not limited to builds?

api/services/Webhooks.js
```diff
@@ -17,7 +17,7 @@ const findSiteForWebhookRequest = (payload) => {
   });
 };
 
-const shouldBuildForSite = site => site && site.buildStatus !== 'inactive'
+const shouldBuildForSite = site => site && site.isActive
```
davemcorwin, line 20:
```suggestion
const shouldBuildForSite = site => site?.isActive
```

davemcorwin (approved):
One fancy js suggestion below, otherwise 💯

Issue #3678, opened 2021-11-03
admin-ui: site build configuration configuration radio button


amirbey:
#### User Story

An admin should be be able to toggle between build configuration settings for  a site in the admin.

#### Background (Optional)

<TODO>

#### Acceptance Criteria

- [x] able to toggle between build configuration settings for  a site in the admin
- [ ] Change made live via Federalist

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3693, opened 2021-11-09
fix sticky container name radio button

amirbey:
resolves https://github.com/18F/federalist/issues/3678


davemcorwin (commented):
Can we just update this form to use the new components and value binding introduced in the last PR instead?

apburnes (approved):

PR #3698, opened 2021-11-11
update site form stuff

davemcorwin:
Some suggestions on refactoring to use the same approach as in other areas. The one thing that is still awkward is when saving the data, the whole page rerenders, closing the accordion and whatnot...

### DIFF
--- a/admin-client/src/components/RadioInput.svelte
+++ b/admin-client/src/components/RadioInput.svelte
@@ -1,10 +1,10 @@
 <script>
   export let error;
   export let hint;
-  export let label;
   export let name;
   export let options;
   export let value;
+  export let tile = false;
 
   $: errorMessageId = `${name}-error-message`;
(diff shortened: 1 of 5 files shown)


admin-client/src/components/RadioInput.svelte
```diff
@@ -1,26 +1,26 @@
...
+  export let tile = false;
 
   $: errorMessageId = `${name}-error-message`;
   $: if (!options.map((option) => option.value ?? option).includes(value)) {
     value = '';
   }
 
   function getOptionLabel(option) {
-    return option.label || option;
+    return option.label ?? option;
```
davemcorwin, line 15: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

amirbey (approved):

Issue #3689, opened 2021-11-09
Flesh out domain administration


amirbey:
#### User Story

An admin user can fully administer (CRUD) domains.

#### Background (Optional)
- deprovision domain from UI
- destroy domain from UI
- ~edit domain from UI~
- search domains by state in UI
- do not paginate sites in domain creation form site list

#### Acceptance Criteria

- [x] can deprovision domain from admin UI
- [x] can destroy domain from admin UI
- [x] ~edit domain from admin UI~
- [x] ~add domain to legacy site domain configuration after provisioning and rebuild~
- [x] add links to site page (in admin) from domain index/show pages
- [x] can search domains by state in UI
- [x] sites are not paginated in domain creation form site list

- [x] Change made live via Federalist

svenaas:
Deprovisioning works as of PRs #3720, #3723, and #3726. Added a separate task, #3727, to improve the UX.

svenaas:
PR #3731 took care of "can destroy domain from admin UI"
PR #3732 took care of "add links to site page (in admin) from domain index/show pages"
and "add domain to legacy site domain configuration after provisioning and rebuild" has been extracted into issue #3733. 

And with that everything here is done except "Change made live via Federalist" so I'm moving this to "Done?"

PR #3699, opened 2021-11-15
3689 domain admin sites and search

davemcorwin:
Partially addresses #3689 

- search domains by state in UI
- do not paginate sites in domain creation form site list

Regarding the solution for the second item, this could have been done just by passing a large number to the existing site query, but felt that the use case was so different that it really deserved it's own action. This will also be much more performant if we do have a lot of sites.


amirbey (approved):
LGTM

Issue #3703, opened 2021-11-17
audit site deactivation and reactivation


amirbey:
#### Background (Optional)

User invitations should be tracked in the audit table.

#### Acceptance Criteria

- [x] an audit record is created when an site is reactivated and shows by whom
- [x] an audit record is created when an site is deactivated and shows by whom
- [x] Change made live via Federalist

Issue #3702, opened 2021-11-16
audit organization creation and deactivation


amirbey:
#### Background (Optional)

User invitations should be tracked in the audit table.

#### Acceptance Criteria

- [x] an audit record is created when an org is created and shows by whom
- [x] an audit record is created when an org is deleted and shows by whom
- [x] an audit record is created when an org is deactivated and shows by whom
- [ ] Change made live via Federalist

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3706, opened 2021-11-22
audit user activity - orgs, sites, orgRoles

amirbey:
Create events to audit who is performing CRUD on organizations and membership, sites, and organizationRoles.

Resolves:
https://github.com/18F/federalist/issues/3703
https://github.com/18F/federalist/issues/3702

### DIFF
--- a/api/models/event.js
+++ b/api/models/event.js
@@ -13,6 +13,7 @@ module.exports = (sequelize, DataTypes) => {
     SITE_USER: 'site-user',
     REQUEST_HANDLER: 'request-handler',
     USER_ACTION: 'user-action',
+    ADMIN_ACTION: 'admin-action',
   };
 
   function isValidType(value) {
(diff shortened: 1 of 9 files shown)


api/models/event.js
```diff
@@ -12,6 +12,7 @@ module.exports = (sequelize, DataTypes) => {
     SOCKET_IO: 'socket.io',
     SITE_USER: 'site-user',
     REQUEST_HANDLER: 'request-handler',
+    USER_ACTION: 'user-action',
```
davemcorwin, line 15: should we differentiate between a user and an operator?

davemcorwin (approved):
See comment below, otherwise gtg

api/models/event.js
amirbey: great suggestion 👍🏿  ... i've added ADMIN_ACTION: 'admin-action'

davemcorwin (approved):

Issue #3712, opened 2021-11-26
add migrate up to install


amirbey:
#### User Story

A user following setup instructions should be able to create-dev-data on a current database

#### Background (Optional)

Either add migrate up to create-dev-data package.json script or add as a step to instructions

#### Acceptance Criteria

- [ ] create-dev-data runs successfully after docker-compose build
- [ ] Change made live via <TODO>.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

PR #3719, opened 2021-12-03
Updates to README.md

svenaas:
Multiple changes to README.md:
- Adds the missing migrate:up setup step, fixing #3712.
- Adds the missing admin-client dependency setup step.
- Updates the GitHub OAuth app setup instruction to indicate the correct "github2" URL path.
- Updates the OAuth configuration setup step to match the initial state of `config/local.sample.js`.
- Adds specific call for the _v7_ version of the Cloud Foundry CLI.
- Adds a note of the admin client's local port.
- Adds mention that certain pieces of functionality aren't expected to work in the development environment (at least with regard to the seed data).

## security considerations
There are no changes to the code, but this does add explicit mention of admin-client to README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Note that `npm run update-local-config` will need to be re-run with some frequen
 
 1. Run `docker-compose build`.
 1. Run `docker-compose run app yarn` to install dependencies.
+1. Run `docker-compose run admin-client yarn` to install dependencies.
 1. Run `docker-compose run app yarn create-dev-data` and answer its prompts to create some fake development data for your local database.
 1. Run `docker-compose up` to start the development environment.
 
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ In our Docker Compose environment, `app` is the name of the container where the 
 For example:
 
 - Use `docker-compose run app yarn test` to run local testing on the app.
-- Use `docker-compose run app yarn lint:diff` to check that your local changes meet our linting standards.
+- Use `docker-compose run app yarn lint` to check that your local changes meet our linting standards.
 
 Similarly you can run any command in the context of the database container `db` by running `docker-compose run db <THE COMMAND>`.
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ Note that when using `docker-compose run`, the docker network will not be expose
 
 The `db` container is exposed on port `5433` of your host computer to make it easier to run commands on. For instance, you can open a `psql` session to it by running `psql -h localhost -p 5433 -d federalist -U postgres`.
 
+The admin client is running on port `3000` of hour host computer.
+
 #### Front end application
 
 If you are working on the front end of the application, the things you need to know are:
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ The `db` container is exposed on port `5433` of your host computer to make it ea
 
 The admin client is running on port `3000` of hour host computer.
 
-Some aspects of the system aren't expected to be fully functional in a development environment. For example: the "View site", "View logs", and "Uploaded files" links associated with sites in the seed data do not reach working URLs.
+Some aspects of the system aren't expected to be fully functional in a development environment. For example: the "View site" and "Uploaded files" links associated with sites in the seed data do not reach working URLs.
 
 #### Front end application
 
--- a/README.md
+++ b/README.md
@@ -48,15 +48,10 @@ This will be the file that holds your S3 and SQS configurations.
 
 1. Once you have created the application, you'll see a `Client ID` and `Client Secret`. Open the `config/local.js` file in your text or code editor and update it with these values:
     ```js
-    passport: {
-      github: {
-        options: {
-          clientID: 'VALUE FROM GITHUB',
-          clientSecret: 'VALUE FROM GITHUB',
-          callbackURL: 'http://localhost:1337/auth/github/callback'
-        }
-      }
-    }
+    const githubOptions = {
+      clientID: 'VALUE FROM GITHUB',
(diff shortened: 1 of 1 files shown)


svenaas:
I strongly suspect that there are multiple other things in README.md that are out of date. These are just the ones I encountered getting the app running on my local dev environment for the first time. Reviewing this PR might present a good opportunity to catch other necessary changes. 

There's one more specific thing that I ran into but don't yet know enough to properly correct the README: the `update-local-config` step in "For 18F/TTS developers" blew up when I tried to run it locally. @apburnes questioned whether that step is really used anymore, then noted that it looks like @davemcorwin has kept the script up to date. Dave and Andrew, can you help determine whether that step should stay in the README?

README.md
```diff
@@ -44,19 +44,14 @@ _Note: some terminal commands may take a while to process, without offering feed
 
 1. Make a copy of `config/local.sample.js` and name it `local.js` and place it in the `config` folder. You can do this by running `cp config/local{.sample,}.js`
 This will be the file that holds your S3 and SQS configurations.
-1. [Register a new OAuth application on GitHub](https://github.com/settings/applications/new). Give your app a name and "Homepage URL" (`http://localhost:1337`), and use `http://localhost:1337/auth/github/callback` as the "Authorization callback url".
+1. [Register a new OAuth application on GitHub](https://github.com/settings/applications/new). Give your app a name and "Homepage URL" (`http://localhost:1337`), and use `http://localhost:1337/auth/github2/callback` as the "Authorization callback url".
```
davemcorwin, line 47: This should be both urls is possible: `http://localhost:1337/auth/github/callback`, `http://localhost:1337/auth/github2/callback`

README.md
```diff
@@ -81,6 +76,8 @@ Note that `npm run update-local-config` will need to be re-run with some frequen
 
 1. Run `docker-compose build`.
 1. Run `docker-compose run app yarn` to install dependencies.
+1. Run `docker-compose run admin-client yarn` to install dependencies.
```
davemcorwin, line 79: There actually is a new `Makefile` that makes some of these steps clearer, maybe we should update the readme to point to it.

README.md
```diff
@@ -106,6 +103,10 @@ Note that when using `docker-compose run`, the docker network will not be expose
 
 The `db` container is exposed on port `5433` of your host computer to make it easier to run commands on. For instance, you can open a `psql` session to it by running `psql -h localhost -p 5433 -d federalist -U postgres`.
 
+The admin client is running on port `3000` of hour host computer.
+
+Some aspects of the system aren't expected to be fully functional in a development environment. For example: the "View site", "View logs", and "Uploaded files" links associated with sites in the seed data do not reach working URLs.
```
davemcorwin, line 108: hmmm, `View Logs` should generally work...

davemcorwin (commented):
woohoo first PR! See below for a few minor thoughts

README.md
svenaas: I think I grouped those three things together overly broadly in the notes I was working from here. "View logs" *is* a working URL — it 200s — but reports "Not found" and "This build does not have any build logs" for the links I've tried.

README.md
svenaas: Are you saying we should add both URLs to one application, or add two applications? The GitHub "Register a new OAuth application" only has a field for one callback URL. When I entered `http://localhost:1337/auth/github/callback` I was unable to authenticate in local Federalist using GitHub. When I changed it to `ttp://localhost:1337/auth/github2/callback` it fixed the problem.

README.md
davemcorwin: That just means that no logs were seeded for those builds.

README.md
davemcorwin: aaaahhhhhh, ok, maybe 2 applications makes more sense:
- localhost:1337/auth/github2/callback
- localhost:1338/auth/github/callback

README.md
svenaas: That would be great but you may be the best person to document this option.

README.md
svenaas: I went in to adjust the README with this and realized I don't know how it's supposed to be set up. It's easy enough to make two GitHub applications, but how is a developer supposed to edit `githuboptions` in `config/local.js` to represent both of them? `config/local.sample.js` only has keys `clientID` and `clientSecret`. Can you provide an example of a `local.js` configuration that binds to two GitHub OAuth apps?

davemcorwin: `local.sample.js` does this. there are `options`, `authorizationOptions`, and `externalOptions`, those are used in different Github `passport` "strategies" in the various `passport.js` configurations.

This is confusing and should be cleaned up post-SCR.

svenaas: All right. `passport.js` associates `authorizationOptions` with "github2" and `options` with "github". I suppose `authorizationOptions` is used at :1337 and `options` is used at :1338? Do you recommend I update `local.sample.js` to standardize two separate configurations, or should I leave README referring only to the business happening at :1337?

davemcorwin: just update the README for now, the work to clean this up will happen after the SCR.

svenaas: In that case I think my change to the "github2" callback for the single OAuth application is probably sufficient as it stands. Developers won't need to set the others unless they're working at :1338, and that port isn't documented at all in the README and can perhaps wait until after the SCR as well.

README.md
svenaas: I have removed my assertion about "View logs".

davemcorwin (approved):

svenaas:
Thanks! Even for a README and staging I don't think I'll hit merge as I walk out the door on a Friday. Next week!

PR #3725, opened 2021-12-08
actually return the value...

### LINKED ISSUE #3724: Investigate why return value of Bull jobs is not reflected in bull board

davemcorwin:
Resolves #3724 

- Return the job value

## security considerations
None


apburnes (approved):
Switches!

Issue #3729, opened 2021-12-13
Confirm before deleting/deprovisioning domain in admin client


svenaas:
#### User Story

As an admin I need a chance to confirm that I haven't clicked Deprovision too hastily or in error.

#### Background (Optional)

As implemented the Deprovision button immediately begins the deprovisioning process. If clicked when it shouldn't be it could take some time to restore things; admins should be asked to confirm that this action is intended.

#### Acceptance Criteria

- [ ] A confirmation interface is presented allowing the admin user to either cancel or proceed with deprovisioning.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

svenaas:
It's not _pretty_ but `window.confirm()` is currently supported cross-browser (so long as it's not in an `<iframe>`):

1. https://caniuse.com/mdn-api_window_confirm
2. https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm

svenaas:
Merged into staging with #3734. Just waiting for deployment to production.

PR #3731, opened 2021-12-14
Add domain Delete button in admin client

svenaas:
- Once DNS status is loaded the admin client Show page now provides a Delete button when appropriate (when the domain is in the "Pending" state). 
- Clicking the Delete button triggers a confirmation dialog per #3729 before anything happens. If the admin user confirms then the domain is deleted. 
- On success the user is returned to the domains index page. On failure, any error is displayed as a notification. 

This pull request represents work toward #3689 and should be sufficient to satisfy #3729.

## security considerations
None


davemcorwin (approved):
LGTM

PR #3734, opened 2021-12-16
Add confirmation dialog before deprovisioning domain

svenaas:
- Adds a confirmation dialog before deprovisioning a domain through the admin client

This uses the same UX as the recently-added confirmation before deleting. It addresses #3729.

## security considerations
None

### DIFF
--- a/admin-client/src/pages/domain/Show.svelte
+++ b/admin-client/src/pages/domain/Show.svelte
@@ -32,7 +32,10 @@
   }
 
   function deprovision() {
-    domainPromise = deprovisionDomain(id);
+    // eslint-disable-next-line no-alert
+    if (window.confirm('Are you sure you want to destroy this domain?')) {
+      domainPromise = deprovisionDomain(id);
+    }
   }
 
   async function destroy() {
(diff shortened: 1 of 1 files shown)


admin-client/src/pages/domain/Show.svelte
```diff
@@ -32,7 +32,10 @@
   }
 
   function deprovision() {
-    domainPromise = deprovisionDomain(id);
+    // eslint-disable-next-line no-alert
+    if (window.confirm('Are you sure you want to destroy this domain?')) {
```
davemcorwin, line 36:
```suggestion
    if (window.confirm('Are you sure you want to deprovision this domain?')) {
```

davemcorwin (approved):
One suggestion, otherwise GTG.

admin-client/src/pages/domain/Show.svelte
svenaas: Oh for heaven's sake. I was Ctrl-Z-ing back and forth on my way to this as I resolved a different lint thing and I obviously missed one Cmd-Ctrl-Z and lost this. Thanks for catching it, Dave.

Issue #3733, opened 2021-12-15
Integrate new domain management with legacy site settings


davemcorwin:
#### User Story

User facing site settings should be in sync with new domain management

#### Acceptance Criteria

- [x] After a domain is provisioned and the corresponding live site/demo url DOES NOT exist, populate the corresponding live site/demo url in the site table with the first domain name from `domain.names` and rebuild
- [x] After a domain is provisioned and the corresponding live site/demo url DOES exist, do nothing 
- [x] When a domain is deprovisioned, depopulate the corresponding live site/demo url in the site table and rebuild ONLY if it corresponds to one of the names domain being deprovisioned
- [x] When a live/site demo URL is unset or when a domain exists for a live site/demo and one of its names matches the site url, disable the live site/demo url field in the UI and ignore it in backend parameters
- [x] Change made live via deploy Federalist application.

svenaas:
Functionally complete and merged with #3739. Not yet inspected in staging (or deployed to production).

PR #3739, opened 2021-12-23
Domain management site integration

svenaas:
- Update site live/demo URL as appropriate when de/provisioning a domain from the admin interface:
  - After a domain is provisioned and the corresponding live site/demo url _does not_ exist, populate the corresponding live site/demo url in the site table with the first domain name from `domain.names`.
  - After a domain is provisioned and the corresponding live site/demo url _does_ exist, do nothing
  - When a domain is deprovisioned, depopulate the corresponding live site/demo url in the site table _only_ if it corresponds to one of the names domain being deprovisioned
- When a site URL has been modified following provision/deprovision based on the rules above, enqueue a site rebuild
- A site's URL is considered to be managed by domains if the URL is null _or_ the the URL is non-null and matches the URL of an associated domain with corresponding context (site/demo).
- When a site URL is managed by domains the URL field is disabled in the UI and ignored in backend parameters.

This work will address #3733 upon production deployment.

## security considerations
None

### DIFF
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -172,8 +172,16 @@ module.exports = wrapHandlers({
     };
 
     // Include domain URL(s) in update only if they're managed by an associated Domain
-    const canEditLiveUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'site');
-    const canEditDemoUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'demo');
+    const canEditLiveUrl = !DomainService.isSiteUrlManagedByDomain(
+      site,
+      site.Domains,
+      Domain.Contexts.Site
+    );
+    const canEditDemoUrl = !DomainService.isSiteUrlManagedByDomain(
+      site,
+      site.Domains,
+      Domain.Contexts.Demo
+    );
     if (canEditLiveUrl) {
       updateParams.domain = params.domain;
     }
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -6,6 +6,7 @@ const SiteMembershipCreator = require('../services/SiteMembershipCreator');
 const UserActionCreator = require('../services/UserActionCreator');
 const EventCreator = require('../services/EventCreator');
 const siteSerializer = require('../serializers/site');
+const DomainService = require('../services/Domain');
 const {
   Build, Organization, Site, User, Event, Domain,
 } = require('../models');
@@ -152,7 +153,7 @@ module.exports = wrapHandlers({
   async update(req, res) {
     const { user, params: { id: siteId }, body } = req;
 
-    const site = await fetchModelById(siteId, Site.forUser(user));
+    const site = await fetchModelById(siteId, Site.forUser(user), { include: [Domain] });
 
     if (!site) {
       return res.notFound();
@@ -171,6 +172,13 @@ module.exports = wrapHandlers({
       domain: params.domain,
       engine: params.engine,
     };
+
+    // Exclude domain URL(s) from update if they're managed by an associated Domain
+    const canEditLiveUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'site');
+    const canEditDemoUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'demo');
+    if (!canEditLiveUrl) { delete updateParams.domain; }
+    if (!canEditDemoUrl) { delete updateParams.demoDomain; }
+
     await site.update(updateParams);
     EventCreator.audit(req.user, Event.labels.USER_ACTION, 'Site Updated', {
       site: { ...updateParams, id: site.id },
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -164,20 +164,22 @@ module.exports = wrapHandlers({
     const params = Object.assign(site, body);
     const updateParams = {
       demoBranch: params.demoBranch,
-      demoDomain: params.demoDomain,
       defaultConfig: params.defaultConfig,
       previewConfig: params.previewConfig,
       demoConfig: params.demoConfig,
       defaultBranch: params.defaultBranch,
-      domain: params.domain,
       engine: params.engine,
     };
 
-    // Exclude domain URL(s) from update if they're managed by an associated Domain
+    // Include domain URL(s) in update only if they're managed by an associated Domain
     const canEditLiveUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'site');
     const canEditDemoUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'demo');
-    if (!canEditLiveUrl) { delete updateParams.domain; }
-    if (!canEditDemoUrl) { delete updateParams.demoDomain; }
+    if (canEditLiveUrl) {
+      updateParams.domain = params.domain;
+    }
+    if (canEditDemoUrl) {
+      updateParams.demoDomain = params.demoDomain;
+    }
 
     await site.update(updateParams);
     EventCreator.audit(req.user, Event.labels.USER_ACTION, 'Site Updated', {
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -7,7 +7,7 @@ const UserActionCreator = require('../services/UserActionCreator');
 const EventCreator = require('../services/EventCreator');
 const siteSerializer = require('../serializers/site');
 const {
-  Build, Organization, Site, User, Event,
+  Build, Organization, Site, User, Event, Domain,
 } = require('../models');
 const siteErrors = require('../responses/siteErrors');
 const {
@@ -30,7 +30,7 @@ module.exports = wrapHandlers({
   async findAllForUser(req, res) {
     const { user } = req;
 
-    const sites = await Site.forUser(user).findAll();
+    const sites = await Site.forUser(user).findAll({ include: [Domain] });
 
     if (!sites) {
       return res.notFound();
--- a/api/services/Domain.js
+++ b/api/services/Domain.js
@@ -124,7 +124,7 @@ function isSiteUrlManagedByDomain(site, domains, context) {
   let siteUrlIsManagedByDomain = false;
   domains.forEach((domain) => {
     if (domain.context === context) {
-      if (domain.names.split(',').find(name => `https://${name}` === site[siteDomain])) {
+      if (domain.namesArray().find(name => `https://${name}` === site[siteDomain])) {
         siteUrlIsManagedByDomain = true;
       }
     }
@@ -154,7 +154,7 @@ async function rebuildAssociatedSite(domain) {
 async function updateSiteForProvisionedDomain(domain) {
   // Populate appropriate site URL if it isn't already set
   const site = await domain.getSite();
-  const firstDomain = `https://${domain.names.split(',')[0]}`;
+  const firstDomain = `https://${domain.namesArray()[0]}`;
   const siteDomain = Site.domainFromContext(domain.context);
   if (isSiteUrlManagedByDomain(site, [domain], domain.context) && site[siteDomain] === null) {
     await site.update({ [siteDomain]: firstDomain });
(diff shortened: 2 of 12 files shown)


svenaas:
That last commit factored out a few common lines in an attempt to make CodeClimate happy. Now that I know that it did I'm going to refactor the corresponding tests.

test/api/unit/serializers/site.test.js
```diff
@@ -1,8 +1,9 @@
-const { expect } = require('chai');
+const { expect, assert } = require('chai');
```
davemcorwin, line 1: was this an automagically added `assert` ? Doesn't look like it is used.

test/api/unit/serializers/site.test.js
```diff
@@ -1,8 +1,9 @@
...
 const validateJSONSchema = require('jsonschema').validate;
 
 const siteSchema = require('../../../../public/swagger/Site.json');
 const factory = require('../../support/factory');
+const { Domain, Site } = require('../../../../api/models');
```
davemcorwin, line 6: `Site` doesn't like it is used.

test/api/unit/serializers/site.test.js
```diff
@@ -64,6 +65,19 @@ describe('SiteSerializer', () => {
       expect(object.basicAuth.password).to.eq('**********');
       expect(object.organizationId).to.equal(org.id);
     });
+
+    it('includes URL editability when associated to domain', async () => {
+      const site = await factory.site({ domain: 'https://www.agency.gov', demoDomain: null });
+      const domain = await factory.domain.create({ siteId: site.id, names: 'www.agency.gov', context: "site" });
```
davemcorwin, line 71:
```suggestion
      const domain = await factory.domain.create({ siteId: site.id, names: 'www.agency.gov', context: 'site' });
```

test/api/requests/site.test.js
```diff
@@ -115,6 +116,45 @@ describe('Site API', () => {
         })
         .catch(done);
     });
+
```
davemcorwin, line 119: These tests seem reasonable. I would prefer them to be written in the newer style, there are examples to which I could point, but it's not a big deal to leave them as is.

api/services/Domain.js
```diff
@@ -111,6 +111,73 @@ function checkAcmeChallengeDnsRecord(domain) {
...
+      username: 'Federalist Operators',
+    }).then(b => b.enqueue());
+  }
+}
+
+/**
+ * @param {DomainModel} domain
+ * @returns {Promise<DomainModel>}
+ */
+async function updateSiteForProvisionedDomain(domain) {
```
davemcorwin, line 154: does this need to requery for the site, or can we just pass in both the site and domain as arguments since the only use case right now already has them.

davemcorwin (commented):
woot! thanks @svenaas 

1. Instead of hardcoding `'site'` and `'demo'` where we have access to the `Domain` model, I believe we can use `Domain.Contexts.Site` and `Domain.Contexts.Demo`.
2. We now have some repeated logic that we might be better served putting somewhere. In particular we should look at adding:
  - `Site.domainFromContext = context => context === 'site' ? 'domain' : 'demoDomain'`
  - `Domain.prototype.namesArray = function() { return this.names.split(','); }`
  - `Domain.prototype.firstName = function() { return this.namesArray()[0]; }`
3. See a few additional comments below.

api/controllers/site.js
```diff
@@ -171,6 +172,13 @@ module.exports = wrapHandlers({
       domain: params.domain,
       engine: params.engine,
     };
+
+    // Exclude domain URL(s) from update if they're managed by an associated Domain
+    const canEditLiveUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'site');
+    const canEditDemoUrl = !DomainService.isSiteUrlManagedByDomain(site, site.Domains, 'demo');
+    if (!canEditLiveUrl) { delete updateParams.domain; }
+    if (!canEditDemoUrl) { delete updateParams.demoDomain; }
+
```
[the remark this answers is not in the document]
svenaas: Yup, I considered this too, and appreciate the vote in favor.

api/services/Domain.js
```diff
@@ -111,6 +111,73 @@ function checkAcmeChallengeDnsRecord(domain) {
...
+  // Populate appropriate site URL if it isn't already set
+  const site = await domain.getSite();
+  const firstDomain = `https://${domain.names.split(',')[0]}`;
+  const siteDomain = domain.context === 'site' ? 'domain' : 'demoDomain';
+  if (site[siteDomain] === null) {
```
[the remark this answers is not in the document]
svenaas: The reason I didn't do that is that in cases where the domain URL is _non_-null but matches the `firstDomain` the function `isSiteUrlManagedByDomain` will return true, resulting in an unnecessary update and rebuild — we only need to update and rebuild when the site's domain URL is null and will be set as a result of provisioning. We could still avoid the unnecessary actions if we checked `isSiteUrlManagedByDomain` *and* the nullness of the domain URL, of course. Do you think that's better, @davemcorwin?

api/services/Domain.js
```diff
@@ -111,6 +111,73 @@ function checkAcmeChallengeDnsRecord(domain) {
...
+  if (site[siteDomain] === null) { return true; }
+  if (domains.length === 0) { return false; }
+  let siteUrlIsManagedByDomain = false;
+  domains.forEach((domain) => {
+    if (domain.context === context) {
+      if (domain.names.split(',').find(name => `https://${name}` === site[siteDomain])) {
+        siteUrlIsManagedByDomain = true;
+      }
+    }
+  });
```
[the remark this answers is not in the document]
svenaas: Those are _differently_ gnarly, but I'm sure they are more idiomatic. I'll sit with them and reflect and pick one. Thanks!

api/services/Domain.js
```diff
@@ -111,6 +111,73 @@ function checkAcmeChallengeDnsRecord(domain) {
...
+
+/**
+ * @param {DomainModel} domain
+ */
+async function rebuildAssociatedSite(domain) {
+  const site = await domain.getSite();
+  const branch = site[domain.context === 'site' ? 'defaultBranch' : 'demoBranch'];
+  if (branch) {
+    await Build.create({
+      user: null,
```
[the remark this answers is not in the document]
svenaas: Oh, good catch. Thank you!

api/services/Domain.js
svenaas: There's no reason I couldn't do it the way you suggest.

test/api/unit/serializers/site.test.js
svenaas: Definitely automagic.

api/services/Domain.js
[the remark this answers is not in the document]
davemcorwin: let's check both

api/services/Domain.js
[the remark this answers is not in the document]
davemcorwin: 😄

test/api/requests/site.test.js
svenaas: I mean, I almost just added my single expectation
```
expect(responseSite).to.include.keys('canEditLiveUrl','canEditDemoUrl');
```
to the `'should not render any sites not associated with the user'` test above, but decided to copy that test wholesale to make this new concern more explicit. 

If the tests should be rewritten in the newer style then they should be, but I vote that that's outside the scope of this PR, and for the moment it would be weird for this one new test to be in a different style than the one it's derived from. Maybe we should create an issue to update the test style? Or maybe we have one?

api/services/Domain.js
svenaas: On closer inspection the only place this is called right now is in `DomainService`'s `checkProvisionStatus()`, which does _not_ already have the site. I vote for leaving the site query where it is in `updateSiteForProvisionedDomain()`.

The same goes for `updateSiteForDeprovisionedDomain()` and `checkDeprovisionStatus()`. 

What do you say, @davemcorwin?

test/api/unit/serializers/site.test.js
svenaas: True. Probably an artifact of an uncommitted early state of my work.

api/services/Domain.js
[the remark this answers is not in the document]
svenaas: Done.

test/api/unit/serializers/site.test.js
svenaas: Whoops. Weird that eslint didn't complain about that.

svenaas:
I'm now using Domain.Contexts.Site and Domain.Contexts.Demo in this branch's/PR's new code, where Domain is already available. 

There are a few more places where I could use it if I required the domain model, but I hesitate to add that require just for this. 

There are also plenty of places _beyond_ this branch and PR where this change could probably be made, but some of them are dicier and I think that's outside of the scope of this PR and associated issue.

api/services/Domain.js
svenaas: For the record this suggestion was about `updateSiteForProvisionedDomain`'s call of `rebuildAssociatedSite`, which is a place where the change makes sense. Taking care of it now.

api/services/Domain.js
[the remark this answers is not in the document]
svenaas: Both work. I chose the latter. eslint made me put the closing paren on the line above, though.

api/services/Domain.js
```diff
@@ -12,6 +12,7 @@ const { States } = Domain;
 
 /**
  * @typedef {object} DomainModel
+ * @typedef {object} SiteModel
```
davemcorwin, line 15: The `SiteModel` declaration should be in it's own comment block, I believe here it is going to mess with the definition of `DomainModel` which is supposed to have a property `state`.

api/services/Domain.js
```diff
@@ -111,6 +112,68 @@ function checkAcmeChallengeDnsRecord(domain) {
...
+
+/**
+ * @param {DomainModel} domain
+ * @returns {Promise<DomainModel>}
+ */
+async function updateSiteForDeprovisionedDomain(domain) {
+  // Clear appropriate site URL if it matches a domain name
+  const site = await domain.getSite();
+  if (isSiteUrlManagedByDomain(site, [domain], domain.context)) {
+    const siteDomain = domain.context === Domain.Contexts.Site ? 'domain' : 'demoDomain';
```
davemcorwin, line 170:
```suggestion
    const siteDomain = Site.domainFromContext(domain.context);
```

api/services/Domain.js
```diff
@@ -149,7 +212,7 @@ async function provision(domain, dnsResults) {
 
   const site = await domain.getSite();
 
-  const [firstDomainName] = domain.names.split(',');
+  const [firstDomainName] = domain.namesArray();
```
davemcorwin, line 215:
```suggestion
  const firstDomainName = domain.firstName();
```

test/api/unit/services/Domain.test.js
```diff
@@ -1,8 +1,10 @@
-const { expect } = require('chai');
+const { expect, assert } = require('chai');
```
davemcorwin, line 1:
```suggestion
const { expect } = require('chai');
```

test/api/unit/services/Domain.test.js
```diff
@@ -1,8 +1,10 @@
...
 const sinon = require('sinon');
 
-const { Domain } = require('../../../../api/models');
+const { Domain, Build } = require('../../../../api/models');
+const { Site } = require('../../../../api/models');
```
davemcorwin, line 5:
```suggestion
const { Domain, Build, Site } = require('../../../../api/models');
```

test/api/unit/services/Domain.test.js
```diff
@@ -1,8 +1,10 @@
...
 const { domain: DomainFactory } = require('../../support/factory');
+const { site: SiteFactory } = require('../../support/factory');
```
davemcorwin, line 7:
```suggestion
const { domain: DomainFactory, site: SiteFactory } = require('../../support/factory');
```

davemcorwin (approved):
Minor suggestions below, otherwise GTG!

test/api/unit/services/Domain.test.js
svenaas: I thought I already took care of this. Perhaps it showed up again.

api/services/Domain.js
svenaas: Missed a few spots. Thanks for catching them!

svenaas:
I kind of wish linting would help with require style consistency.

davemcorwin:
> I kind of wish linting would help with require style consistency.
It can, you may not have it configured properly or there may be things we could find/implement.

Issue #3744, opened 2022-01-04
Remove UAA war from staging environment


svenaas:
The file `/uaa/cloudfoundry-identity-uaa-4.19.0.war` — which is only needed in local development but which also ends up in a Concourse container— contains potentially vulnerable code and needs to be removed. 

- [x] **High priority**: Remove the war file while still allowing tests to succeed
- [x] **High priority**: Deploy war file removal to staging
- [x] Find a way to _optionally_ include the UAA image/container/war dependency in local development
- [x] .gitignore the war file to prevent this cycle from beginning anew
- [x] ~~Find a way for developers to get the necessary war file onto their workstations without maintaining it in the repo~~ (Extracted to #3749)
- [x] Update the documentation
- [x] ~~Eventually move to a newer war that no longer presents the vulnerability~~

svenaas:
#3745 will take care of the first and third checkboxes, and merging #3745 will take care of the second.

svenaas:
Merged. More tomorrow.

svenaas:
#3746 took care of the `.gitignore` and updated the documentation. We're not sure about a good source of the war, though. We may explore building it (per Cloud Foundry docs) during docker setup.

PR #3746, opened 2022-01-05
Git ignore cloudfoundry-identity-uaa-4.19.0.war and update README

svenaas:
- Adds `uaa/cloudfoundry-identity-uaa-4.19.0.war` to `.gitignore`
- Updates readme to note the potential need to install `cloudfoundry-identity-uaa-4.19.0.war` locally in `uaa/` and to document appropriate usage of the second optional uaa yml file with `docker-compose` 

This is work towards #3744 

## security considerations
The `.gitignore` change is intended to reduce the likelihood of `cloudfoundry-identity-uaa-4.19.0.war` being inadvertently recommitted and pushed, where, if it found its way to staging, it could create a security concern.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ current-sites.csv
 credentials*.json
 tmp
 ci/vars/.*
+uaa/cloudfoundry-identity-uaa-4.19.0.war
--- a/Makefile
+++ b/Makefile
@@ -19,29 +19,29 @@ db: ## Connect to the database - it must already be running ie with `make start`
 	psql postgresql://postgres:password@localhost:5433/federalist
 
 build-client: ## Build client app - typically only done to test production artifacts
-	docker-compose -f ./docker-compose.yml -f ./docker-compose.uaa.yml run -rm app yarn build
+	docker-compose run -rm app yarn build
 
 install: ## Install npm deps
-	docker-compose -f ./docker-compose.yml -f ./docker-compose.uaa.yml run --rm app yarn
-	docker-compose -f ./docker-compose.yml -f ./docker-compose.uaa.yml run --rm admin-client yarn
+	docker-compose run --rm app yarn
+	docker-compose run --rm admin-client yarn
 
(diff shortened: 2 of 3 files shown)


svenaas:
I have not found an authoritative source for `cloudfoundry-identity-uaa-4.19.0.war`. [Cloud Foundry documentation describes building your own war file locally](https://docs.cloudfoundry.org/concepts/architecture/uaa.html) but I'm not sure whether we want to encourage that. (I _have_ found sites hosting `https://docs.cloudfoundry.org/concepts/architecture/uaa.html` but not sites I'd be comfortable linking to.) Any suggestions?

davemcorwin:
The instructions are there, looks like we could include the building of UAA (maybe war) in our dockerfile or as part of the local setup. It does look different from what we have now.

svenaas:
That would be a good solution if it's workable. But maybe it's worth getting the `.gitignore` into the staging branch now and working on that subsequently?

davemcorwin:
for sure, can you make the changes to the Makefile as well?

svenaas:
Sure. Do you want two execution options, or do you want the Makefile to always assume a local war and use both dc yml files?

davemcorwin:
this is all only used locally

svenaas:
Yes, I know that. I just didn't know whether you wanted the option to run locally without it. Here's a version that includes the two-yml approach only:

davemcorwin:
We don't need multiple options, just include it where we need it, which is only when running the application ie (`docker compose up`)

svenaas:
OK... 

So we **don't** need it for 
- yarn (dependency installation of app or admin-client)
- db:migrate
- create-dev-data
- lint
- test (of app or admin-client)

But we **do** need it for
- docker-compose build
- docker-compose up

Is that correct? If so I'll update both the README and the Makefile.

davemcorwin:
think so

svenaas:
So it has been done.

davemcorwin (approved):
👍

Issue #3749, opened 2022-01-11
Source or automate build of UAA war


svenaas:
This repo no longer includes `/uaa/cloudfoundry-identity-uaa-4.19.0.war` due to an included vulnerability that we don't want running on the staging environment (the UAA container only needs to be run in local development).

In #3744, from which this issue is extracted, we removed the war file from the repo and established a way for developers to use their own local copy. What we do not have is place for developers to _get_ that local copy. 

If we can find a reliable, canonical (or at least reputable) source online, whether of `cloudfoundry-identity-uaa-4.19.0.war` specifically or of a version sufficiently compatible with our codebase, we can simply reference that in the README. 

Failing that, we can try building the war locally ([doing this is covered in the Cloud Foundry documentation](https://docs.cloudfoundry.org/concepts/architecture/uaa.html)). If that works it may be practical to automate such a build when Docker prepares the UAA container.

svenaas:
Mind you, the Cloud Foundry docs tell you to 
```
git clone git://github.com/cloudfoundry/uaa.git
```
instead of the current and correct
```
git clone git@github.com:cloudfoundry/uaa.git
```
so they _may_ not be reliable for our purposes...

svenaas:
After installing Java:
```
brew tap homebrew/cask-versions
brew update
brew tap homebrew/cask
brew install java
sudo ln -sfn /usr/local/opt/openjdk/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk.jdk
```
I tried Cloud Foundry's
```
./gradlew :cloudfoundry-identity-uaa:war
```
and got errors that may be attributable to Java version issues — the steps above installed Java 17.0.1. Trying again with Java8 specifically: 
```
brew install adoptopenjdk8
export JAVA_HOME=$(/usr/libexec/java_home -v1.8)
./gradlew :cloudfoundry-identity-uaa:war
```
This time Gradle told me it requires Java11 at least. _Thanks_, Gradle. 

```
brew install openjdk@11 
brew uninstall openjdk@17
sudo ln -sfn /usr/local/opt/openjdk@11/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk-11.jdk
export JAVA_HOME=$(/usr/libexec/java_home -v11)
./gradlew :cloudfoundry-identity-uaa:war
```
and I got
```
BUILD SUCCESSFUL in 1m 2s
```
but so far I haven't figured out where it _put_ the war file.

[...]

PR #3750, opened 2022-01-12
Source UAA War from S&W GitHub

svenaas:
- Sources UAA war online instead of locally, and uses a newer version.
  - Instead of looking for `uaa/cloudfoundry-identity-uaa-4.19.0.war` — which must be sourced elsewhere and placed there by the developer — then loading that file into a UAA Docker, `Dockerfile-uaa` now downloads a `75.13.0` version of the UAA war [from Stark & Wayne](https://github.com/starkandwayne/uaa-war-releases/releases), verifies that its sha512 hash hasn't changed since the one I observed today, and uses that instead.
  - This required updating from Java 8 to Java 11 and from Tomcat 8 to Tomcat 9. I went with Java 11.0.13 and Tomcat 9.0.45, specifically. I found these to work locally but others might as well, and we could explore it further if anyone feels it's worth doing.
- Update the README to remove the need to have your own 4.19.0 war before running with the UAA container enabled. As of this writing I do still have the UAA container in a optional separate dc yml file. We could consider abandoning that additional complexity, but it does have the positive effect of preventing any UAA stuff from happening on staging.

This work should satisfy #3749 (which was split out of #3744). 

All I know to do to test this is to
1. Remove any local UAA docker containers and images.
2. Rebuild and restart: 
```
docker-compose -f ./docker-compose.yml -f ./docker-compose.uaa.yml build
docker-compose -f ./docker-compose.yml -f ./docker-compose.uaa.yml up
```
3. Go to localhost:1337 and make sure I can log in via UAA. 

**If there are other details to local UAA integration or utilization, _please_ let me know so that I can follow up.**

## security considerations
Some of us have started to get warnings about having the 4.19.0 war file on our workstations. If we're satisfied that this change works, we won't have to do that any more.

### DIFF
--- a/Dockerfile-uaa
+++ b/Dockerfile-uaa
@@ -1,22 +1,23 @@
-FROM openjdk:8u141-jre
+FROM openjdk:11.0.13-jre
 
 ENV UAA_CONFIG_PATH /uaa
 ENV CATALINA_HOME /tomcat
 
 COPY ./uaa/uaa.yml /uaa/uaa.yml
 
-RUN wget -q https://archive.apache.org/dist/tomcat/tomcat-8/v8.0.28/bin/apache-tomcat-8.0.28.tar.gz
-RUN wget -qO- https://archive.apache.org/dist/tomcat/tomcat-8/v8.0.28/bin/apache-tomcat-8.0.28.tar.gz.md5 | md5sum -c -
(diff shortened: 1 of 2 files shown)


davemcorwin (approved):
LGTM

PR #3756, opened 2022-01-25
Parameterize app name in Federalist web interface

svenaas:
-
-
-

## security considerations
None


svenaas:
I want to do a round of spot-checking on this locally in the browser before I ask for a code review.

davemcorwin (approved):
Assuming nothing breaks, this looks ok to me. We can play around in the staging envs as well.

It's not worth changing here, but it's worth evaluating whether or not we need the app name at all in a lot of the client-facing messaging, we probably could have just dropped it in a bunch of places. 🤷

Issue #3753, opened 2022-01-25
Correctly set site engine when adding a site


davemcorwin:
#### User Story

As a user, when I specify the site engine when adding a site to Federalist, it should be used.

#### Background (Optional)

Currently, it seems as if this is ignored.

#### Acceptance Criteria

- [ ] Specifying the site engine when adding a site works
- [ ] Change made live via deploy Federalist.

svenaas:
The front end is definitely POSTing the engine, e.g.:

```
federalist_1    | info: HTTP POST /v0/site
```
```
... "body":{"owner":"svenaas","repository":"ncd-test","engine":"node.js","organizationId":null}} ...
```
```
... "body":{"owner":"svenaas","repository":"ncd-test","engine":"static","organizationId":null}} ...
```

or 

```
app_1           | info: HTTP POST /v0/site 
```
```
... "body":{"owner":"svenaas","repository":"ncd-test","engine":"hugo","organizationId":null}} ...
```

svenaas:
But by the time `api/services/SiteCreator`'s `paramsForNewSite` gets called the `params` it gets passed do not include that `engine`...

Oh, it's the controller. 🙄

svenaas:
Merged into staging, but not yet into production.

svenaas:
I just noticed that this wasn't closed yet. The commit associated with this fix has been in `main` for a long time.

PR #3761, opened 2022-01-27
Correctly set site engine when adding a site

svenaas:
- Actually passes the `engine` parameter through when creating new sites.

This is intended to address #3753

## security considerations
None

### DIFF
--- a/api/controllers/site.js
+++ b/api/controllers/site.js
@@ -126,7 +126,7 @@ module.exports = wrapHandlers({
   async create(req, res) {
     const {
       body: {
-        owner, template, organizationId, repository,
+        owner, template, organizationId, repository, engine,
       },
       user,
     } = req;
@@ -136,6 +136,7 @@ module.exports = wrapHandlers({
       template,
       organizationId: toInt(organizationId),
       repository,
+      engine,
       sharedBucket: false,
     };
 


svenaas:
Weird. Failed on Concourse but not on local. It's one of these: 
```
  1) Domain Service
       .updateSiteForProvisionedDomain()
         sets the domain name on the associated site if not previously set:
     Error: Timeout of 2500ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/test/api/unit/services/Domain.test.js)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)
```

In the hopes that it is _actually_ a timeout because Concourse is being slow, I am trying it again.

svenaas:
Well, shoot. Same timeout.

The tests still pass on my local machine (where, like Concourse, `PRODUCT` is set to `'Pages'`) as well as on CircleCI (where `PRODUCT` is set to `'Federalist'`). 

The timeout happens in the DomainService unit test during a call to `DomainService.updateSiteForProvisionedDomain` which should have zero connection to the change this branch introduces in the site controller. 

Any clever ideas, @18F/federalist-admins?

davemcorwin (approved):
looks like something I probably borked....

Issue #3758, opened 2022-01-25
In the admin UI, fix bug in new domain creation form to be able to also select the Context of "demo"


apburnes:
The frontend looks for `demoBranch` in the `sites` returned from the api which is never present because the api is not returning it. Add `demoBranch` to the site data return from the api admin/controllers/site#listRaw.

PR #3766, opened 2022-02-01
Add `demoBranch` field to sites returned when creating domain to enable selecting `demo` context when present

davemcorwin:
- Add `demoBranch` field to sites returned when creating domain to enable selecting `demo` context when present #3758 

## security considerations
None


apburnes (approved):
If `demoBranch`... let there be `demo` 🧙

Issue #3728, opened 2021-12-13
Change org Delete button to Deactivate in admin client


svenaas:
#### User Story

As an admin using the admin client I need to be able to deactivate organizations.

#### Background (Optional)

The Orgs index page currently presents a Delete button in the Actions column. This button currently does nothing. It should be changed to read "Deactivate" and trigger that functionality.

#### Acceptance Criteria

- [ ] The deactivate button is presented instead of the delete button
- [ ] Pressing the deactivate button results in the deactivation of an org.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

svenaas:
Deactivation functionality is not yet implemented for organizations. I can implement this, but @davemcorwin, what business rules should determine whether an organization can or cannot be deactivated? Can organizations be deactivated if they have associated sites? If they have associated users? Does it depend on the status of those sites or users?

svenaas:
Should we add the deactivation event to the audit log?

svenaas:
Is org deactivation a serious enough action that we should require admin confirmation before carrying out the action, as we do for domain deprovisioning and deletion?

davemcorwin:
> Deactivation functionality is not yet implemented for organizations. I can implement this, but @davemcorwin, what business rules should determine whether an organization can or cannot be deactivated? Can organizations be deactivated if they have associated sites? If they have associated users? Does it depend on the status of those sites or users?

The use case here is if an IAA expires, we want to essentially freeze the organization. I believe we already have this in place on a per site basis, so deactivating an organization should have the same effect for all of the sites within an organization as well as preventing access to that organization for its users.

> Should we add the deactivation event to the audit log?
Probably.

> Is org deactivation a serious enough action that we should require admin confirmation before carrying out the action, as we do for domain deprovisioning and deletion?
Probably.

svenaas:
OK. Organization has an isActive property; at the moment my provisional deactivate implementation simply sets that to false. I haven't looked yet to see what, if anything, depends on isActive. It's possible that more things should than do, if we need this to have the effect of preventing users from accessing an org or its sites and that business logic hasn't already been implemented. 

[...]

PR #3770, opened 2022-02-02
Replace organization delete button with de/activate buttons

svenaas:
- Removes the (non-working) delete button on the orgs index page in the admin client
- Replaces it with deactivate and activate buttons
- Disables* the isActive (organization status) field set on the org update page. 

These changes addresses #3728

*While I disabled the isActive field, I did _not_ add logic to reject or ignore this field when updates. It didn't seem important under the circumstances, but I am open to opposing viewpoints.

## security considerations
None

### DIFF
--- a/admin-client/src/pages/organization/Index.svelte
+++ b/admin-client/src/pages/organization/Index.svelte
@@ -1,6 +1,6 @@
 <script>
   import page from 'page';
-  import { fetchOrganizations, deactivateOrganization } from '../../lib/api';
+  import { fetchOrganizations, deactivateOrganization, activateOrganization } from '../../lib/api';
   import { formatDateTime } from '../../helpers/formatter';
   import { DataTable, PaginatedQueryPage } from '../../components';
 
@@ -10,6 +10,11 @@
       deactivateOrganization(id);
       page('/organizations');
     }
+  }
+
+  function activate(id) {
+    activateOrganization(id);
+    page('/organizations');
   }
 </script>
 
@@ -66,6 +71,12 @@
           class="usa-button usa-button--secondary"
           on:click={deactivate(org.id)}>
           Deactivate
+        </button>
+        {:else}
+        <button
+          class="usa-button"
+          on:click={activate(org.id)}>
+          Activate
         </button>
         {/if}
       </td>
--- a/admin-client/src/pages/organization/Index.svelte
+++ b/admin-client/src/pages/organization/Index.svelte
@@ -1,7 +1,16 @@
 <script>
-  import { fetchOrganizations } from '../../lib/api';
+  import page from 'page';
+  import { fetchOrganizations, deactivateOrganization } from '../../lib/api';
   import { formatDateTime } from '../../helpers/formatter';
   import { DataTable, PaginatedQueryPage } from '../../components';
+
+  function deactivate(id) {
+    // eslint-disable-next-line no-alert
+    if (window.confirm('Are you sure you want to deactivate this organization?')) {
+      deactivateOrganization(id);
+      page('/organizations');
+    }
+  }
 </script>
 
 <PaginatedQueryPage path="organizations" query={fetchOrganizations} addAction let:data>
@@ -52,7 +61,13 @@
         </ul>
       </td>
       <td>
-        <button class="usa-button usa-button--secondary">Delete</button>
+        {#if org.isActive}
+        <button
+          class="usa-button usa-button--secondary"
+          on:click={deactivate(org.id)}>
+          Deactivate
+        </button>
+        {/if}
       </td>
     </tr>
   </DataTable>
--- a/api/services/organization/Organization.js
+++ b/api/services/organization/Organization.js
@@ -217,4 +217,13 @@ module.exports = {
     await organization.update({ isActive: false });
     return organization;
   },
+
+  /**
+     * @param {OrganizationModel} organization The organization
+     * @returns {Promise<OrgType>}
+     */
+  async activateOrganization(organization) {
+    await organization.update({ isActive: true });
+    return organization;
+  },
 };
--- a/api/services/organization/Organization.js
+++ b/api/services/organization/Organization.js
@@ -208,4 +208,13 @@ module.exports = {
 
     return [org, uaaUserAttributes];
   },
+
+  /**
+   * @param {OrganizationModel} organization The organization
+   * @returns {Promise<OrgType>}
+   */
+  async deactivateOrganization(organization) {
+    await organization.update({ isActive: false });
+    return organization;
+  },
 };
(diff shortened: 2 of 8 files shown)


svenaas:
CodeClimate is being annoying: 
> **codeclimate/total-coverage** — 90% (-0.1% change)
Otherwise this is ready for review.

api/services/organization/Organization.js
```diff
@@ -208,4 +208,22 @@ module.exports = {
...
   },
+
+  /**
+   * @param {OrganizationModel} organization The organization
+   * @returns {Promise<OrgType>}
+   */
+  async deactivateOrganization(organization) {
+    await organization.update({ isActive: false });
+    return organization;
+  },
```
davemcorwin, line 219:
```suggestion
  deactivateOrganization(organization) {
    return organization.update({ isActive: false });
  },
```

api/services/organization/Organization.js
```diff
@@ -208,4 +208,22 @@ module.exports = {
...
+
+  /**
+     * @param {OrganizationModel} organization The organization
+     * @returns {Promise<OrgType>}
+     */
+  async activateOrganization(organization) {
+    await organization.update({ isActive: true });
+    return organization;
+  },
```
davemcorwin, line 228:
```suggestion
  activateOrganization(organization) {
    return organization.update({ isActive: true });
  },
```

test/api/unit/services/organization.test.js
```diff
@@ -487,4 +487,24 @@ describe('OrganizationService', () => {
       });
     });
   });
+
+  describe('.deactivate', () => {
+    it('deactivates the organization', async () => {
+      const org = await orgFactory.create({ isActive: true });
+      expect(org.isActive).to.be.true;
+      await OrganizationService.deactivateOrganization(org);
```
davemcorwin, line 495: May want to also test that the returned `org` has the appropriate value for `isActive` since that is what is used in the controller.

test/api/unit/services/organization.test.js
```diff
@@ -487,4 +487,24 @@ describe('OrganizationService', () => {
...
+      await org.reload();
+      expect(org.isActive).to.be.false;
+    });
+  });
+
+  describe('.activate', () => {
+    it('activates the organization', async () => {
+      const org = await orgFactory.create({ isActive: false });
+      expect(org.isActive).to.be.false;
+      await OrganizationService.activateOrganization(org);
```
davemcorwin, line 505: May want to also test that the returned `org` has the appropriate value for `isActive` since that is what is used in the controller.

admin-client/src/pages/organization/Index.svelte
```diff
@@ -52,7 +66,19 @@
         </ul>
       </td>
       <td>
-        <button class="usa-button usa-button--secondary">Delete</button>
+        {#if org.isActive}
+        <button
+          class="usa-button usa-button--secondary"
+          on:click={deactivate(org.id)}>
```
davemcorwin, line 72: Does this work? Shouldn't it be
```suggestion
          on:click={() => deactivate(org.id)}>
```

admin-client/src/pages/organization/Index.svelte
```diff
@@ -52,7 +66,19 @@
...
+          Deactivate
+        </button>
+        {:else}
+        <button
+          class="usa-button"
+          on:click={activate(org.id)}>
```
davemcorwin, line 78:
```suggestion
          on:click={() => activate(org.id)}>
```

admin-client/src/pages/organization/Index.svelte
```diff
@@ -1,7 +1,21 @@
...
   import { formatDateTime } from '../../helpers/formatter';
   import { DataTable, PaginatedQueryPage } from '../../components';
+
+  function deactivate(id) {
+    // eslint-disable-next-line no-alert
+    if (window.confirm('Are you sure you want to deactivate this organization?')) {
+      deactivateOrganization(id);
+      page('/organizations');
+    }
+  }
```
davemcorwin, line 13: We probably want to wait for the deactivate before navigating. Also, what happens if it fails?

```suggestion
  async function deactivate(id) {
    // eslint-disable-next-line no-alert
    if (window.confirm('Are you sure you want to deactivate this organization?')) {
      await deactivateOrganization(id);
      page('/organizations');
    }
  }
```

admin-client/src/pages/organization/Index.svelte
```diff
@@ -1,7 +1,21 @@
...
+
+  function activate(id) {
+    activateOrganization(id);
+    page('/organizations');
+  }
```
davemcorwin, line 18:
```suggestion
  async function activate(id) {
    await activateOrganization(id);
    page('/organizations');
  }
```

davemcorwin (changes requested):
Thanks @svenaas See comments below!

api/services/organization/Organization.js
svenaas: Oh hey, yeah! I can _do_ that!

admin-client/src/pages/organization/Index.svelte
davemcorwin: it's not about being idiomatic, they are different. The original code SHOULD be calling the function immediately and passing it's result as the click handler, while the second is passing the function itself. We want the latter. The first shouldn't work and makes me wonder what Svelte is doing.

svenaas: I see what you mean – and now that I see it I also wonder what Svelte is doing. I'm going to try it again to confirm that I was seeing what I thought I was seeing, record that result, and then try changing it.

svenaas: Yes, it does work as you see it, and it doesn't activate/deactivate until you click it.

svenaas: :jinx:

svenaas: Yes, with that fix (which I'm linting and testing before pushing) this one looks like it'll be OK. I'll just be a few minutes.

admin-client/src/pages/organization/Index.svelte
svenaas: I went further, just pushed: 

```javascript
  async function deactivate(id) {
    // eslint-disable-next-line no-alert
    if (!window.confirm('Are you sure you want to deactivate this organization?')) { return null; }
    try {
      await deactivateOrganization(id);
      page('/organizations');
      return notification.setSuccess(`Organization ${id} deactivated successfully!`);
    } catch (error) {
      return notification.setError(`Unable to deactivate organization ${id}: ${error.message}`);
    }
  }
```

svenaas:
`codeclimate/total-coverage` you are not acting like my friend today.

test/api/unit/services/organization.test.js
svenaas: Thanks for this really good catch. The version I just push checks this.

davemcorwin (approved):
LGTM

Issue #3727, opened 2021-12-09
Improve de/provision UX in admin client


svenaas:
#### User Story

As an admin administering a domain from its show page I want to provision or deprovision that domain and see the appropriate action button update below the domain details. 

#### Background (Optional)

Currently, when viewing a domain in the "pending" state, if an admin presses the "Provision" button the domain moves into the "provisioning" state, and the client begins periodically checking to update the status information displayed. But it does **not** correctly update the button — the "Provision" button goes away, which is a good start, but once the domain is in the "provisioning" state the "Provision" button should be replaced with a "Deprovision" button, which is indeed displayed if the user fully refreshes the page. 

#### Acceptance Criteria

- [x] Correctly update the correct action button (Provision or Deprovision) and its state (enabled/disabled) after the button has been pressed, without requiring the admin user to fully reload the page.

#### Level of effort - <low/medium/high>

low

#### Implementation outline (if higher than "low" effort):

- [x] The domain state is already updated after the Provision/Deprovision button is pressed; this code should be expanded to also update the button.
- [x] Deployed

svenaas:
I think this may be as simple as calling `refreshDnsResults();` after making the `provisionDomain()` or `deprovisionDomain()` call.

svenaas:
PR created, but not yet reviewed/approved/merged into staging.

svenaas:
#3772 approved and merged. We will need to test it on provision and deprovision in staging to see whether it works or whether there are issues — the thing I'm uncertain about is timing.

svenaas:
This, unfortunately, does _not_ do what I hoped it would in staging.

svenaas:
Trying again with #3781.

svenaas:
This one's looking better on staging, but it takes a while to test because I have to wait for de/provisions to complete. I just successfully started a provision, saw the refresh work, immediately deprovisioned, and saw that work as well. I'm waiting for that to complete and will then try the two actions individually to be sure before I move this one to Done.

svenaas:
Works great on provision. The delete button goes away and the deprovision button appears.

svenaas:
Works great on deprovision too.

PR #3772, opened 2022-02-04
Refresh DNS table (and de/provision button) after domain de/provision

svenaas:
- In admin client domain administration, after pressing the provision or deprovision button, refresh the domain table. 
 
This is intended to ensure that the available button is correctly displayed once the de/provision process has begun, which would address the concern noted in #3727. 

What I don't know is 
1. Whether adding the refresh where I am is _too soon_ after beginning the de/provision process, and
2. Whether there is any way for me to give this a workout _without_ deploying it on staging.

## security considerations
None

### DIFF
--- a/admin-client/src/pages/domain/Show.svelte
+++ b/admin-client/src/pages/domain/Show.svelte
@@ -29,12 +29,14 @@
 
   function provision() {
     domainPromise = provisionDomain(id);
+    refreshDnsResults();
   }
 
   function deprovision() {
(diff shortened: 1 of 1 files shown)


apburnes (approved):

svenaas:
This is going to need review in staging where de/provisioning is meaningfully possible, to make sure the timing is OK. Once we have staging. But I'm merging now to get ahead of that.

PR #3774, opened 2022-02-09
Only configure UAA auth if feature flag is set

davemcorwin:
- Only configure UAA if feature flag is set

This enables Federalist to run without UAA creds since they are not required.

## security considerations
None


svenaas (approved):
I think this looks OK

svenaas:
Except for that build-test-deploy failure, of course...

davemcorwin:
admin Merging as Concourse/pages deployment will be addressed later.

Issue #3771, opened 2022-02-03
Fix event audit logging 


svenaas:
#### User Story

As an operator I want to be able to rely on audit logging working where we expect it to.

Currently of 33 calls to `Event.audit` I believe only 15 are using the correct parameter list, and 18 are using a list which does not work.

#### Background (Optional)

Most of our event logging calls `EventCreator.audit`, which takes these parameters: 
```js
 * @param {string} label The Event label
 * @param {object} model An instance of the target Sequelize model
 * @param {string} message A brief description of the event
 * @param {object} [body] Additional attributes
```

Here is an example call from the admin Site controller, which appears to be valid:
```js
EventCreator.audit(Event.labels.ADMIN_ACTION, req.user, 'Site Activated', { site });
```
Activating a site via the admin UI (which I _believe_ should call this method) raises no errors.

But we also have many calls in the codebase which appear to be invalid, like this example, from the [non-admin] Site controller: 
```js
EventCreator.audit(req.user, Event.labels.USER_ACTION, 'Site Updated', {
  site: { ...updateParams, id: site.id },
});
```
Updating a site via the front end (I specified a demo branch and saved) raises this error:
```
SequelizeValidationError: string violation: label cannot be an array or an object,
Validation error: Invalid event label: [object SequelizeInstance:User],
Validation error: Invalid event model: String
    at InstanceValidator._validate (/app/node_modules/sequelize/lib/instance-validator.js:78:13)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async InstanceValidator._validateAndRunHooks (/app/node_modules/sequelize/lib/instance-validator.js:111:7)
    at async InstanceValidator.validate (/app/node_modules/sequelize/lib/instance-validator.js:93:12)
    at async model.save (/app/node_modules/sequelize/lib/model.js:3877:7)
    at async Function.create (/app/node_modules/sequelize/lib/model.js:2207:12) {"service":"federalistapp-development"}
```

The error in this case is that the user parameter is being passed before the event label instead of the other way around. Perhaps this method signature changed at some point in the past. In any event I'm guessing it hasn't been noticed because a) it does not prevent the site update from being saved, and b) it doesn't blow up in testing, since this audit call is being made from the controller, which does not currently have its own tests.

Changing the call to 
```js
EventCreator.audit(Event.labels.USER_ACTION, req.user, 'Site Updated', {
  site: { ...updateParams, id: site.id },
});
```
[...]

PR #3775, opened 2022-02-11
Fix .audit() calls in controllers.

svenaas:
- Corrects the syntax of `EventCreator.audit` calls in controllers

As documented in #3771, about half of these were wrong. Because the calls are in controllers this did not cause failures in testing, and in general I believe they were just silently breaking in practice. Well, they would cause 'Failed to create Event' errors in the application container and the events wouldn't got logged. For a full list of the events that weren't logging correctly, see the lines changed in this PR.

## security considerations
None


davemcorwin (approved):
LGTM

PR #3776, opened 2022-02-11
Add script to create a Federalist environment

davemcorwin:
- Adds a script to setup an entire Federalist environment

## security considerations
Manually verified all gitleaks findings, no leaks. An issue has been created for preventing false positives when using environment variable.

### DIFF
--- a/scripts/create-instance.sh
+++ b/scripts/create-instance.sh
@@ -1,0 +1,391 @@
+#!/bin/bash
+
+set -e
+
+add_ext_if_not_prod () {
+  local prefix=$1
+  local env=$2
+
+  if [[ $env != "production" ]]; then
+    prefix+="-${env}"
+  fi
+
+  echo $prefix
+}
+
+## Before Run
+# - Create Github OAuth with callback url https://$domain and set env vars
+# - Create required domain
+# - Create required DNS records for the external domain service
+
+echo
+
+show_usage="false"
+
+POSITIONAL=()
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    -e|--env_type) env_type="$2"; shift; shift ;;
+    -d|--domain) domain="$2"; shift; shift ;;
+    -h|--help) show_usage="true"; shift; shift ;;
+    -o|--org) cf_org="$2"; shift; shift ;;
+    -p|--product) product="$2"; shift; shift ;;
+    -s|--subdomain) subdomain="$2"; shift; shift ;;
+    *) POSITIONAL+=("$1"); shift ;;
+  esac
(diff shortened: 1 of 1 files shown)


svenaas:
@davemcorwin this is awesome.

I don't know of any way to test it, though, and I think @apburnes would be the Most Valuable Reviewer here.

scripts/create-instance.sh
```diff
@@ -0,0 +1,391 @@
...
+EOF
+)"
+
+# Grant permissions in the sites space
+cf set-space-role "${api_user_username}" "${cf_org}" "${cf_sites_space}" "SpaceDeveloper"
+
+# Temporarily create a service key to get the bucket name
+tmp_s3_service_key_name="${service_name_s3}-key-tmp"
+cf create-service-key "${service_name_s3}" "${tmp_s3_service_key_name}"
+shared_bucket=$(cf service-key "${service_name_s3}" "${tmp_s3_service_key_name}" | tail -n +2 | jq -r .bucket)
```
apburnes, line 243: Not sure we need a shared bucket in new space creation.

apburnes: Just legacy to deploy the proxy?

scripts/create-instance.sh
```diff
@@ -0,0 +1,391 @@
...
+echo
+echo "-  In the Concourse jumpbox, please run the following:"
+echo "     credhub set -n /concourse/pages/${cf_space}-cf-password -t value -v ${ci_user_password}"
+echo "     credhub set -n /concourse/pages/${cf_space}-cf-username -t value -v ${ci_user_username}"
+echo
+echo "-  Add any additional humans:"
+echo "     cf set-space-role <username@gsa.gov> ${cf_org} ${cf_space} SpaceDeveloper"
+echo "     cf set-space-role <username@gsa.gov> ${cf_org} ${cf_sites_space} SpaceDeveloper"
+echo
+echo "You may now deploy the pages and applications:"
```
apburnes, line 384: Be more explicit about kicking off a build in CI?

scripts/create-instance.sh
davemcorwin: yeah, we still have a few stranglers, we can remove the shared bucket as a separate issue.

scripts/create-instance.sh
davemcorwin: What do you mean?

apburnes: If we want any more documentation on using Concourse or Circle to kick off a deploy. But it's not like we're going to use this that offten.

apburnes (approved):

Issue #3780, opened 2022-02-14
Enable Github authentication for queues ui


davemcorwin:
#### User Story

As a Federalist Operator I want to be able to see/manage background jobs.

#### Acceptance Criteria

- [ ] Queues UI is available and requires `federalist-admin` membership in Github to access
- [ ] Change made live via deploy Queues UI.

PR #3768, opened 2022-02-01
Release - 2021-02-17

svenaas:
## Changes proposed in this pull request:
- Parameterize app name #3709
- Refresh DNS table (and de/provision button) after admin domain de/provision #3727
- Replace admin organization delete button with de/activate buttons (Pages) #3728
- Upgrade node, remove DNS workaround #3737
- Allow selection of demo context when creating domains through admin UI #3758
- Correctly set site engine when adding a site #3761
- Update .codeclimate.yml #3762
- Remove nsp from federalist codeclimate pipeline #3764
- Fix .audit() calls in controllers #3771
- Only configure UAA if feature flag is set #3774 
- Add script to create a Federalist environment #3776
- Add GitHub auth for queues UI, completely separate bull board from the rest of the application, update config #3780

## Post deployment
- Record these changes in the next sprint architecture review following their deployment.

## security considerations
- CodeClimate change adds more static code analysis
- Environment creation script presents gitleaks findings but these have been manually verified and represent no real leaks. An issue has been created for preventing false positives when using environment variable.
- Allowing Github authentication for queues UI based on membership in `federalist-admins` GH org, same as admin interface.

PR #3782, opened 2022-02-15
Add Github auth for queues ui

davemcorwin:
- Configures Github authentication for the Queues UI #3780 
- Some refactoring to partly disentangle the queues ui from the rest of the app, at least some configuration
- use `FederalistLocal` GH org and OAuth apps for local development

## security considerations
Allowing Github authentication for queues UI based on membership in `federalist-admins` GH org, same as admin interface.

## Other thoughts
For the config that has been updated, the architecture reflects where I think we should be going with configuration.
1. The environment (`process.env`) should be source of truth so we can easily configure on CG, Docker, and local.
2. IFF the app is running on CG, map CG json variables (`VCAP_APPLICATION` and VCAP_SERVICES`) to environment variables. Typically this would be with a `.profile` file, but since the Queues UI isn't deployed all by it's lonesome quite yet, we use `env.js` to do this.  
3. If necessary, flesh out (and maybe validate) the configuration from the environment into js (`config.js`) for easy consumption.


davemcorwin:
must fix tests....

svenaas (commented):
I've looked this over and don't have any suggestions, but this is my first time looking at the `bull-board` code. @apburnes, maybe you should have final approval on this one?

Note that if we do approve it we'll want to decide whether it gets merged before #3768 goes into `main`.

apburnes:
> 👍

Tests aren't running because we aren't pulling in redis creds

api/bull-board/config.js
```diff
@@ -1,38 +1,24 @@
-if (process.env.NODE_ENV === 'production') {
-  const cfenv = require('cfenv'); // eslint-disable-line global-require
-  const appEnv = cfenv.getAppEnv();
+const {
```
apburnes, line 1: I can hydrate these when running the tests locally. We use the `config/test.js` convention for other parts of the app  but bull-board is a bit different since it may also be broken out into its own entity through something like the monorepo.

davemcorwin: yeah, i started going down a bit of a yak-shave in my head for how we should handle env vars and config...

apburnes (approved):

Issue #3757, opened 2022-01-25
In the admin UI, create link from a site page to the new domain page with pre-populated site org/repo in the form.


apburnes:
## Acceptance Criteria

- [x] Add link in site page to create new domain
- [x] Add site site ID as a querystring to new domain url ie `/domain/new?siteId=<x>`
- [x] Take the querystring and populate the form field for org repo
- [x] Also look at improving the `select` by either sorting the sites

apburnes:
@davemcorwin are you working on this?

PR #3786, opened 2022-02-18
Add create domain from admin site page

apburnes:
- Take a look at the issue #3757 

## security considerations
None


svenaas (approved):
Looks good to me. 

I do like that you can simply pair `/domains/new?siteId=${id}` to `$: siteId = $router.query.siteId` here.

PR #3787, opened 2022-02-18
Use Redis-backed session and do not cache reponses

davemcorwin:
- #3780 
- Use Redis-backed session instead of non-recommended memory store
- do not cache responses
- resolves potential redirect loop
- makes cookie name unique

## security considerations
Prevents caching of responses that may include cookies


svenaas (approved):
Looks good to me.

Issue #3793, opened 2022-02-23
Investigate build log archiving issues


davemcorwin:
#### User Story

As a Federalist operator I would not like our database to run out of space.

#### Background (Optional)

There is a daily background job that moves build logs from the database to S3, it appears this is not running (or failing)

#### Acceptance Criteria

- [x] Daily build log archiver runs successfully
- [x] all old logs are archived and verify database free space
- [x] Change made live via <TODO>.

_After evaluating, edit this part:_

#### Implementation outline (if higher than "low" effort):

- [ ] <TODO>

davemcorwin:
It looks like the node process is running out of memory and crashing...

davemcorwin:
Apparently `timeout` is not implemented, butour job is still timing out...https://github.com/taskforcesh/bullmq/issues/136

davemcorwin:
Build log archiving is now succeeding, however, misleading errors are reported AFTER success, see https://github.com/OptimalBits/bull/issues/1591 for details on how to configure the queue.

davemcorwin:
this is all good except for the misleading error messages.

PR #3796, opened 2022-02-25
Increase worker memory and moves logs to job

davemcorwin:
It looks like the worker is running out of memory when processing this job, let's try upping the memory a bit. We can keep increasing it to a limit, if that doesn't work, we have a couple options:
1. Move the job to a "sandboxed" processor which may prevent the job from "stalling"
2. Revisit how we archive since we're trying to move all the logs in one request.

- Increase worker memory to 2GB
- Move logs to job so they are visible in Bull Board
- #3793 

## security considerations
None

### DIFF
--- a/.cloudgov/manifest.yml
+++ b/.cloudgov/manifest.yml
@@ -14,7 +14,7 @@ applications:
         command: yarn start-workers
         disk_quota: 2G
         instances: 1
-        memory: ((memory))
+        memory: 2G
     services:
       - app-((env))-uaa-client
       - federalist-((env))-rds
--- a/api/workers/jobProcessors/archiveBuildLogsDaily.js
+++ b/api/workers/jobProcessors/archiveBuildLogsDaily.js
@@ -3,13 +3,27 @@ const { Op } = require('sequelize');
 const PromisePool = require('@supercharge/promise-pool');
 const BuildLogs = require('../../services/build-logs');
 const { Build } = require('../../models');
-const { logger } = require('../../../winston');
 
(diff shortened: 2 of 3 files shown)


.cloudgov/manifest.yml
```diff
@@ -14,7 +14,7 @@ applications:
         command: yarn start-workers
         disk_quota: 2G
         instances: 1
-        memory: ((memory))
```
svenaas, line 17: Any idea what it was getting here before?

svenaas (approved):
This looks good to me.

.cloudgov/manifest.yml
apburnes: Any double parentheses `((...))` in a CF manifest file references a variable.  See [here](https://github.com/18F/federalist/blob/staging/.cloudgov/vars/production.yml) for our production deploys and [here](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#variable-substitution) for the CF docs.

davemcorwin:
Security COnsiderations succeeded, not sure why it is not reflecting that in the checks...

.cloudgov/manifest.yml
svenaas: Yes, I could tell it as going to be a variable. I was curious what the value was before if not 2GB. (And for that matter couldn't we also increase it by editing the variable instead of the code?)

svenaas: (I did find the answer to my production environment value in the production deploys.

Issue #3769, opened 2022-02-01
Remove adding of robots.txt to S3 bucket


davemcorwin:
#### User Story

As a user I don't want a default robots.txt file

#### Background (Optional)

https://github.com/18F/federalist-proxy/issues/111

#### Acceptance Criteria

- [ ] There is no default robots.txt file
- [ ] Change made live via deploy Federalist.

#### Level of effort - low

https://github.com/18F/federalist/blob/staging/api/services/SiteBuildQueue.js#L111

PR #3800, opened 2022-03-02
Remove robots.txt

hursey013:
Related to https://github.com/18F/federalist/issues/3769

- Removes static robots.txt in S3 bucket in favor of `X-Robots-Tag` header (https://github.com/18F/federalist-proxy/pull/145)

## security considerations
None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,6 +153,10 @@ jobs:
           environment:
             APP_HOSTNAME: http://localhost:1337
             PRODUCT: federalist
+            CLOUD_FOUNDRY_API_HOST: https://api.example.com
+            CLOUD_FOUNDRY_OAUTH_TOKEN_URL: https://login.example.com/oauth/token
+            CF_API_USERNAME: deploy_user
+            CF_API_PASSWORD: deploy_pass
 
       - run:
           name: CodeClimate combine and upload coverage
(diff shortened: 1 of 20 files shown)


davemcorwin (changes requested):
Hi @hursey013 
1. The serving of the robots.txt file here is for the Federalist web app, not customer sites, so for the purposes of this issue, we want to leave that alone. This refers to the changes in api/controllers/main.js, api/routers/main.js and associated test.
2. Re: api/services/S3SiteRemover.js, unless we are planning on manually removing all the existing platform-provided robots.txt files, we don't want to remove this code quite yet as all the buckets have the robots.txt in them right now and we need to remove it to delete the bucket. Instead, we should ensure this function is resilient to situations where the file doesn't exist.

hursey013:
@davemcorwin Got excited to rip stuff out without maybe fully understanding the ask, apologies for not looking at this more closely - will revisit.

hursey013:
@davemcorwin removed the extraneous changes. Also looked into S3's `deleteObjects` and it seems like it's safe to run as is even if the file does not exist:

> For each key, Amazon S3 performs a delete action and returns the result of that delete, success, or failure, in the response. Note that if the object specified in the request is not found, Amazon S3 returns the result as deleted.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html

davemcorwin:
> @davemcorwin do we also want the ability to block robots.txt from deploying to sites in the Garden Build under certain conditions? Blocking it _in general_ (as an overrideable default) would be compatible with what I'm working on for [18F/federalist-garden-build#343](https://github.com/18F/federalist-garden-build/issues/343) but I haven't considered making it conditional other than through `federalist.json` configuration.

We don't want to block this by default or have any special logic for it. We're just removing the platform default for previews, users need to be able to supply their own for custom domains. With the changes to the proxy, the existing robots.txt files at the root of the bucket are just not utilized.

davemcorwin (approved):

PR #3801, opened 2022-03-03
Refactor config.js out of CF clients and use env vars

davemcorwin:
- Refactor CF clients to be externally configurable and not rely on config.js
- Uses `.profile` to move CF username and password to env vars
-

## security considerations
None


svenaas (approved):
This looks good!

PR #3802, opened 2022-03-04
Refactor cf client round 2

davemcorwin:
- Move functions in `utils` that were only used by the CF client into the CF client
- Remove transient references to `config`

No the CF Api Client only depends on the CF Auth client and the Http Client, and CF Auth Client only depends on the Https client. Tests are another matter to be addressed later.

## security considerations
None


svenaas (approved):
Refactoring ❤️

Issue #3806, opened 2022-03-10
Add basic platform memory metrics


davemcorwin:
#### User Story

As an operator I want to know the status of the platform memory quota

#### Background (Optional)

An incident where the platform ran out of memory due to phantom tasks

#### Acceptance Criteria

- [ ] Actionable metrics are collected in New Relic
- [ ] Actionable alert conditions and notifications are configured
- [ ] Change made live via deploy of Metrics application and New Relic Configuration.

PR #3807, opened 2022-03-10
Metrics

davemcorwin:
- Add Metrics application #3806 
- Reports to New Relic

This application demonstrates most of what a monorepo approach would look like. With the exception of being in the same repository, it is an entirely self-contained application with it's own manifest and ci pipeline.

Until `pnpm` is implemented, code that should end up in separate packages is in `lib`, and some has been duplicated and enhanced from the main app.

Alerts in New Relic have also been configured (subject to improvements) and a dashboard added.

## security considerations
None - this pushes metrics to New Relic, has no endpoints


svenaas (approved):
This is a cool piece of work, @davemcorwin.

giphy

svenaas (approved):
Still good. Better, even!

Issue #3803, opened 2022-03-08
Migrate GH access token in concourse from davemcorwin personal token to bot token. Cg-ci-bot or should we have one for pages?


apburnes:
Create a Github account for cloud-gov-pages-operations to create access token for CI pipelines

# Acceptance Criteria
- [x] Create a google group for cloud-gov-pages-operations
- [x] Create a GH user for cloud-gov-pages-operations
- [x] Create a access token for 1 year
- [x] Add credentials to credhub
- [x] Add a calendar reminder a week before token expiration
- [x] Update doc on creation, update, and use
- [x] Update CI pipelines with new access token

PR #3808, opened 2022-03-10
Update github access token name from credhub

apburnes:
Related to the work of #3803

- Update credhub credentials used in CI

## security considerations
Updates credentials


davemcorwin (commented):
Why not just update the value (and add pages bot to all our repos...) instead of changing the name?

apburnes:
Because I like a hard life

PR #3809, opened 2022-03-11
Refactor spies to stubs and resolve

davemcorwin:
- Refactors spies to stubs that return promises that resolve so there aren't unresolved promises.

## security considerations
None


svenaas (approved):
Yeah, those were all mine. Your "Make a stub not a spy silly" on #3807 makes me feel a little better that I didn't know better back when I introduced all of these way back in #3739 (oh wow I've been here long enough to describe one of my commits with "way back"). Anyway this looks good.

davemcorwin:
> Yeah, those were all mine. Your "Make a stub not a spy silly" on #3807 makes me feel a little better that I didn't know better back when I introduced all of these way back in #3739 (oh wow I've been here long enough to describe one of my commits with "way back"). Anyway this looks good.

It depends what you're going for. A spy may be what you want, if you want the function to do what it normally does, but be able to track the calls. In this case, we don't care about testing the effects of that function so might as well just stub and return a promise that resolves.

Issue #3873, opened 2022-06-01
Skip `federalist-users` check on login via Github to Pages


davemcorwin:
#### User Story

As a user I want to be able to initially authenticate to Pages with Github so I can upgrade to UAA auth.

#### Background (Optional)

Pages does not have access to view membership in `federalist-users` so will rely on current Federalist access to determine access to the platform. 

#### Acceptance Criteria

- [x] A user can authenticate to Pages with Github
- [x] Change made live via deploy Federalist web.

PR #3874, opened 2022-06-01
3873 skip federalist users check for pages

davemcorwin:
- Addresses #3873 

## security considerations
Temporarily could allow access to Pages to a former user, if the webhook for the removal was not handled.


apburnes (approved):
I can't wait to tear out all of this limbo logic. 

_You, you, you oughta know._

PR #3922, opened 2022-09-07
Deploy 9/7/2022

drewbo:
- Fixes uaa client to include `scim.write` and only grabs the token with the `scim.write` scope when adding a user to the 'pages.user' group.
- Fixes the scoped query for a user's sites on the `Site.forUsers` method to properly account for a user's sites in and outside of an org.

## security considerations
Update's uaa client authority to add `scim.write` scope but is only used on authenticated requests when adding a user to the `pages.user` uaa group. The token is not saved or available to the end user.

### DIFF
--- a/api/models/site.js
+++ b/api/models/site.js
@@ -84,15 +84,24 @@ const associate = ({
   }));
   Site.addScope('forUser', user => ({
     where: {
-      [Op.or]: [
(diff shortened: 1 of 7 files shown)


apburnes (approved):
🛫

Issue #3915, opened 2022-08-30
Remove SQS logic, dependencies, and config from pages-core


apburnes:
With the move to the Redis queue, we need to remove the AWS SQS related dependencies, configuration, and logic from the app and its deployments.

## Acceptance Criteria
* [x] Remove the user provided services supply now deprecated SQS credentials
* [x] Remove config for SQS
* [x] Remove logic within app for SQS

--- 

## Security considerations
Needed for SCR

apburnes:
See https://github.com/cloud-gov/pages-core/pull/3928

drewbo:
Deployed and services removed

PR #3928, opened 2022-09-09
chore: remove sqs logic config

drewbo:
- Removes SQS queue logic, config, tests from the application. Towards #3915  

## security considerations
None


apburnes (approved):
✂️

PR #3932, opened 2022-09-13
feat: Disable Federalist site creation

drewbo:
- Remove the add site (and github connect) button for Federalist

## security considerations
None


apburnes (commented):
Works well between the two different deployments. Can we add an alert on the Federalist deployment on the `/sites` page telling users the can log into Pages if they want to create a new site?

drewbo:
I updated the main transition banner with:
- A link to the new app
- Updated link to cloud.gov/pages migration info
- A call to head there to create new sites

This might be a little heavy in terms of content in one banner and also driving extra traffic? But it keeps us from having multiple banners.

apburnes (approved):
Looks good and works well for the initial migration.

PR #3933, opened 2022-09-14
feat: block already migrated users from using Federalist

drewbo:
- blocks users from Github authentication (the only option) on Federalist if they've already migrated (i.e. have a UAA identify)
- fixes the logout button for UAA identities in Federalists (though this should now be unreachable)

## security considerations
None


hursey013 (approved):
Looks good!

PR #3937, opened 2022-09-21
fix: add more rest-like route for removal of user from organization

### LINKED ISSUE #3936: Unable to remove user from Organization
When a user is removed from an Organization a 404 is returned on the request delete/organization-role. 

## Acceptance Criteria

- [ ] Determine next steps and acceptance criteria

drewbo:
- fix: add more rest-like route for removal of user from organization
- Close #3936

## security considerations
None 

## Notes
there is some non-ideal organization of routes now because this new route is `/organizations/...` while the others are `/organization-role` but I'm okay doing this little by little

### DIFF
--- a/api/routers/organization-role.js
+++ b/api/routers/organization-role.js
@@ -6,7 +6,10 @@ router.use(sessionAuth);
 router.use(csrfProtection);
 
 router.get('/organization-role', OrganizationRoleController.findAllForUser);
-router.delete('/organization-role', OrganizationRoleController.destroy);
+// for reasons that aren't totally clear, this method doesn't receive a request
+// body when running normally but the test requests do pass the body properly
+// router.delete('/organization-role', OrganizationRoleController.destroy);
+router.delete('/organization/:org_id/user/:user_id', OrganizationRoleController.destroy);
 router.put('/organization-role', OrganizationRoleController.update);
 
 module.exports = router;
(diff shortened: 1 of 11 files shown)


api/routers/organization-role.js
```diff
@@ -6,7 +6,10 @@ router.use(sessionAuth);
 router.use(csrfProtection);
 
 router.get('/organization-role', OrganizationRoleController.findAllForUser);
-router.delete('/organization-role', OrganizationRoleController.destroy);
+// for reasons that aren't totally clear, this method doesn't receive a request
+// body when running normally but the test requests do pass the body properly
+// router.delete('/organization-role', OrganizationRoleController.destroy);
+router.delete('/organization/:org_id/user/:user_id', OrganizationRoleController.destroy);
```
apburnes, line 12: Would it make sense to move this to the `api/routers/organization`?  Your call.

_* Note: I'm not sure if any of the `organization-role` routes make much sense but not worth refactoring everything at this moment._

apburnes (commented):
Had a comment on just on semantics of the route. When we get the test passing, I think we are good to go.

drewbo:
@apburnes do you know why the admin tests are failing when the code there wasn't changed and they don't share controllers? My assumption would be that these tests would still pass (since they are unchanged; even if the functionality isn't fixed on the admin side)

test/api/support/validateAgainstJSONSchema.js
```diff
@@ -12,6 +12,7 @@ const validateAgainstJSONSchema = (action, path, statusCode, response) => {
   const actionLower = action.toLowerCase();
   const pathLower = path.toLowerCase();
   const statusCodeInt = parseInt(statusCode, 10);
+  console.log(schema.paths[pathLower], actionLower)
```
apburnes, line 15:
```suggestion
```

apburnes (dismissed):
Good to go 👍 . I put in one suggestion to remove a console.log statement.

apburnes (approved):

PR #3949, opened 2022-10-07
add route level authorization

drewbo:
- Add route level authorization based on uaa groups
- Close https://github.com/cloud-gov/product/issues/2164

## Notes
- Tests are only partial but show that the `authorize` function works as expected across multiple endpoints
- Need to validate that [these](https://docs.google.com/spreadsheets/d/1SJ_qW2jivasrPYkfzu3a_wPbh0OqZth8nXJcdb86owY/edit#gid=624931696) are the permissions we should enforce and they match the current route structure.

## security considerations
Admin API will now differentiate between `pages.admin` and `pages.support` roles, should be stricter than current security requirements but implementation error could create risks


api/admin/passport.js
```diff
@@ -49,14 +49,19 @@ if (Features.enabled(Features.Flags.FEATURE_AUTH_UAA)) {
...
+    passReqToCallback: true,
   };
 
-  const verify = async (accessToken, refreshToken, profile, callback) => {
+  const verify = async (req, accessToken, refreshToken, profile, callback) => {
     try {
-      const user = await verifyUAAUser(accessToken, refreshToken, profile, ['pages.admin']);
-
+      let user = await verifyUAAUser(accessToken, refreshToken, profile, ['pages.support']);
+      if (!user) {
```
apburnes, line 58: This block is a bit confusing when using the `verifyUAAUser` function and it will always set the `req.session.role` to some value even if they are not an admin or support user. Maybe we should create a new function in [`uaaStrategy.js`](https://github.com/cloud-gov/pages-core/blob/staging/api/services/uaaStrategy.js) that can simplify this logic. We could create a function to verify and return an object with the user and their role.

apburnes (approved):
🚢

PR #3950, opened 2022-10-07
chore: update concourse links to contain variables

drewbo:
- chore: update concourse links to contain variables

## security considerations
- None


apburnes (commented):
Awesome! I can't believe those where broken for so long 😬 . With the [`ci/federalist-pipeline.yml`](https://github.com/cloud-gov/pages-core/pull/3950/files#diff-c1607936811d7f64943fb07754a8a018b857efa3ec7c9cd82d1dbe8b58c2a2ca) there is no need for the instance variable queries in the url since that pipeline does not use them.

PR #3951, opened 2022-10-11
Remove references to shared s3 bucket

hursey013:
- Removes conditional logic and unit tests associated with the (legacy) shared s3 bucket

## security considerations
Removes unused dependency


hursey013:
@drewbo there is a lot going on in these tests that I don't fully understand, but it's clear the tests were written with the assumption they would be using a shared S3 bucket. There was one test for using a private bucket however, so I was able to pull that logic out and now use it in the setup of the tests that require it.

hursey013:
@drewbo Not super DRY, but hopefully easier to grok. Main change is to mock the private s3 bucket creation before each test to make it more reflective of how the process works today (as opposed to testing both private and shared buckets).

Issue #3953, opened 2022-10-12
Refresh Github token for non-migrated users who are invited to an organization


drewbo:
Current federalist users who are invited to an organization but haven't migrated will have a GH token that won't have allowed third-party access from the Pages Production access. We need to remove their current token so they can re-authorize with Github using the provided button. h/t @Ephraim-G for identifying this potential issue earlier today

## Acceptance Criteria

- [ ] Determine next steps and acceptance criteria

drewbo:
Improved functionality in https://github.com/cloud-gov/pages-core/pull/3958

PR #3955, opened 2022-10-14
Feat: add github auth reset button to user settings

drewbo:
- add Github auth reset button to user settings
- Towards #3953
- Note that this does not require removing the `githubAccessToken` from the database (because the token itself doesn't change) but trying to authorize again will prompt accepting the OAuth app again

## security considerations
None


apburnes (approved):
👍

PR #3958, opened 2022-10-20
Github OAuth App Token Revocation

drewbo:
- Adds a new route/service for revoking a user token grant from our Github OAuth app
- Uses this new route from the frontend `/settings` page for token resets

## Background
The token reset implemented in #3955 will only prompt re-authorizing the "Pages {env}" Github OAuth application for all organizations if the old token is first revoked. This implements the revocation flow prior to re-auth.

## security considerations
None.

### DIFF
--- a/frontend/components/GithubAuthButton.jsx
+++ b/frontend/components/GithubAuthButton.jsx
@@ -3,7 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import globals from '../globals';
-
+import api from '../util/federalistApi';
 import { IconGitHub } from './icons';
 
 const apiUrl = globals.APP_HOSTNAME;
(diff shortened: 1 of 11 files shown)


frontend/components/GithubAuthButton.jsx
```diff
@@ -16,8 +16,8 @@ const calcWindow = () => ({
   top: window.screen.height / 2 - 800 / 2,
 });
 
-function authorize() {
-  return new Promise((resolve, reject) => {
+function authorize(revokeFirst) {
```
apburnes, line 19: Is there anytime we wouldn't revoke first?

apburnes (commented):
Awesome! What do you think about putting the GH token reset component first to make it more prominent? 

ie
user-settings

frontend/components/GithubAuthButton.jsx
drewbo: On the main page, if they aren't connected to Github already, we don't really need to revoke first (although doing so won't cause issues as we catch the GH 404 in our own API). Fine to implement either way

apburnes (approved):

Issue #3957, opened 2022-10-17
Disable checking of pages user status for external auth


drewbo:
Our external github auth endpoint checks for a user's cloud.gov Pages membership. This shouldn't be necessary for using Netlify CMS as a content editor as long as the user has access to the account. That check is currently performed by [Netlify CMS's auth code](https://github.com/netlify/netlify-cms/blob/dbf2920254fb3682e12463a6df8ded4b94b55be0/packages/netlify-cms-backend-github/src/implementation.tsx#L330-L333)

## Notes
- [Discussion](https://gsa-tts.slack.com/archives/G7EFTFW03/p1666034747701459)

## Acceptance Criteria

- [ ] Write up an ADR on how this could work
- [ ] Confirm that this authentication structure is acceptable per the current cloud.gov Pages SSP
- [ ] Users without a cloud.gov Pages account can use Netlify CMS to make content edits

apburnes:
Closing stale.

PR #3960, opened 2022-10-24
Feat disable user check 3957

drewbo:
- Disable checking cloud.gov Pages/Federalist membership for external auth
- Enabled checking site referrer information to see that the Netlify CMS app is running at a cloud.gov registered domain
- Note: An Events API call was added to log an audit even for each external authentication request. It is currently only logging the `referer` and also uses a dummy `Site` model to ensure the method doesn't error.
- Towards #3957

## security considerations
External users are no longer checked to see that they use cloud.gov Pages/Federalist but they are instead checked to see if the site `Referer` is a `cloud.gov` domain (generated url) or registered domain (by querying the `Sites` list). There is an additional (existing) check implemented by `netlify-cms` directly which ensures the OAuth'ed GitHub user has write access to the site they'd like to edit.

### DIFF
--- a/api/external-auth/index.js
+++ b/api/external-auth/index.js
@@ -1,5 +1,7 @@
 const express = require('express');
 const passport = require('./passport');
+const { Site, Event } = require('../models');
+const EventCreator = require('../services/EventCreator');
 
 const script = (nonce, status, content) => `
   <script nonce="${nonce}">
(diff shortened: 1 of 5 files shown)


apburnes:
@drewbo should we close this out?

drewbo:
@apburnes I think we didn't get a clear answer ever on how this would affect our various security/compliance boundaries; are we still interested in pursuing this change?

apburnes:
I put https://github.com/cloud-gov/pages-core/issues/3957 into this IP sprint so we can review it again.

svenaas:
I just noticed this again. What's its status?

drewbo:
I think we need a "product + security" review of this before proceeding. Basically:
- Do we want non-users to be able to make changes to the repo and kickoff builds (this is possible either way via the Github UI and somewhat dependent upon what we do with access tokens)?
- What are the security ramifications of making that change? 
- What are the security ramifications of having an open endpoint for Github authentication
- General code review:
  - Is our current check for "is this request coming from one our sites" sufficient? (probably no)
  - How should we handle event logging for this?

drewbo:
Not implementing

PR #3962, opened 2022-10-28
Update build log access

drewbo:
- Match build log access with ability to access sites

## security considerations
None

### DIFF
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -1,4 +1,4 @@
-const { Op, Sequelize } = require('sequelize');
+const { Op } = require('sequelize');
 const crypto = require('crypto');
 const URLSafeBase64 = require('urlsafe-base64');
 const SiteBuildQueue = require('../services/SiteBuildQueue');
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -54,14 +54,8 @@ const associate = ({
   }));
   Build.addScope('forSiteUser', user => ({
     include: [{
-      model: Site,
+      model: Site.forUser(user),
       required: true,
-      include: [{
-        model: User,
-        where: {
-          id: user.id,
-        },
-      }],
     }],
   }));
 };
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -1,3 +1,4 @@
+const { Op, Sequelize } = require('sequelize');
 const crypto = require('crypto');
 const URLSafeBase64 = require('urlsafe-base64');
 const SiteBuildQueue = require('../services/SiteBuildQueue');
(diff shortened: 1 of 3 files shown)


drewbo:
**Intent**: A user should be able to "see" a site if either:
  1. They added the site directly to Pages. We have previously named this being a "collaborator" and this implies they have GitHub write access to the repo.
  2. OR they are a member of an organization to which the site belongs. This is, potentially, the more common method of having visibility to a given site going forward.

From a technical standpoint, checking both of these conditions can be a complex SQL query (and often a more difficult `sequelize` query...) but was implemented recently for site querying directly [here](https://github.com/cloud-gov/pages-core/commit/b42e655b0f79807afb72d6d13add512b7ba3b451#diff-57ecef382be4c067db6767be24b321fc476220bfc2afb3ebb854cd5613e30287L84-L103) (via the `Site` model as a scope `forUser`). This scope is used for listing sites on the main application page and seems to be working correctly. 

However, this PR needs to update very similar logic for access to the build logs (which first involves a check on whether the user has access to the `Site` and then `Build`). When attempting to use the `Site.forUser` scope in the `Build.forSiteUser` scope include statement, the SQL query is malformed (many portions left out here):

```
SELECT Build.id, Build.branch, ..., Site.id, Site.demoBranch, ...,  "Site->Users"."id" AS "Site.Users.id"...
FROM "build" AS "Build"
INNER JOIN 
  "site" AS "Site" ON "Build"."site" = "Site"."id" AND
    ("Site"."deletedAt" IS NULL AND
      (
        ("Users"."id" IS NOT NULL OR "Site"."organizationId" IS NOT NULL) AND
        ("Users"."id" = 2 OR "Organization->OrganizationRoles"."userId" = 2)
       )
     )
     ...
    WHERE "Build"."id" = '7';
```

Shown above: Because of the nested scope in the include statement, the condition is written as `Users.id` in the join clause whereas it's `Site->Users.id` in the SELECT clause. This results in the error `uncaughtException: missing FROM-clause entry for table \"Users\"`. I see four potential paths forward, in rough order of level of effort:
- Writing out the logic again inside the `Build.forSiteUser` scope (likely to work, very non-DRY)
- Trying upgrading to `sequelize v7` to see if that changes the nesting naming logic (there is [some indication](https://sequelize.org/docs/v7/other-topics/upgrade/#attribute-names-cannot-use-syntax-reserved-by-sequelize) that it might)
- Researching or finding a `sequelize` expert who knows in-library ways to resolve this
- Replacing `sequelize` with another ORM/query-builder (longer term, much more difficult to implement quickly)

hursey013 (approved):
Awesome, nice work!

PR #4002, opened 2022-12-15
convert app to functional component, update test strategy

drewbo:
- Convert app to functional component, update tests
- The test strategy isn't great at this point, but (1) it's working in the application* and (2) I'd like to replace the test library and strategy at the very end of this 
- The one asterisk above is that I can't confirm the prior notification strategy is working easily, open to suggestions around that

## security considerations
None


frontend/components/app.jsx
```diff
@@ -1,61 +1,46 @@
...
+      shouldClearAlert(alert);
     }
-  }
-
-  render() {
-    const {
-      children,
-      notifications,
-    } = this.props;
+  }, [location.key]);
```
apburnes, line 36: Do we need to also add the `alert` prop into the array of the `useEffect` argument so the effect will fire when the `alert` prop is updated?

drewbo: So I'm not positive about the logic here but I purposefully removed `alert` from the dependencies to simplify the logic. By my understanding, the previous implementation and intent was:
1. fire on every props change
2. if the alert is stale, clear it
3. if the location key matches, early return (no action)
4. if there is an existing message during a location key change, we should mark it as stale for removal on the next props update

The new logic should _mostly_ retain that because it will only fire on location.key changes (skips the early return) and will mark the message as stale + clear stale messages, the one difference is that it requires a location change to clear the message (which wasn't true before).

I wasn't sure of a great way to test/simulate this behavior though

apburnes: Thanks for laying out your thought process. It makes sense to me and let's see how it goes.

apburnes (approved):
🪝 🪝

Issue #3987, opened 2022-12-06
Add event logging to authentication


apburnes:
In order to improve our logging and understand our users' authentication experience, add event logging to different steps in the authentication process.

## Acceptance Criteria

- [x] Use the EventCreator service to log events
- [x] Add log audit event to specify the authentication for Netlify CMS, Github, and UAA
- [x] Add log error events during the UAA authentication process

apburnes:
Closed by https://github.com/cloud-gov/pages-core/pull/4010

PR #4003, opened 2022-12-19
Add additional event logging around authentication

apburnes:

- Work to close the #3987 issue

## security considerations
none


api/services/passport.js
```diff
@@ -150,7 +155,13 @@ if (Features.enabled(Features.Flags.FEATURE_AUTH_UAA)) {
         ['pages.user', 'pages.support', 'pages.admin']
       );
 
-      if (!user) return callback(null, false, flashMessage);
```
drewbo, line 153: (Optional): I think there are two separate ways that this can fail: if they don't have a UAA profile sent back or if it isn't in the `pages.user` group. Is it worth logging these separately or will we see it in the `profile`?

drewbo (approved):
Looks great. One question added but fine to ship either way

api/services/passport.js
apburnes: We can add additional logging in the UAA strategy and the verify UAA user. Should we set them as audit or error types for the events?

drewbo: I'd say audit since it we'll throw for failure later on

apburnes: Moved the event auditing into the `uaaStrategy` service to log if:
- UAA user is not in the verified group
- If no UAAIdentity exists in our DB that matches the UAA user
- If no User is associated to the UAAIdentity in our DB

drewbo (approved):

PR #4004, opened 2022-12-19
Add most/many components rendered under /site as functional components

drewbo:
- The test framework was updated to:
  - Centralize test configuration into one file (`_mochaSetup.js`). It's still a bit messy but this should be easier to port to a new framework eventually
  - Use a new (and unfortunately slower) rendering method `mountRouter` to ensure that hooks for location or redux state don't error.
  - Reduce pollution of `global`
- Various components were updated to become functional components rather than class components
- Component tests were updated in response

## Pieces I'm worried about breaking:
- Expandable areas [previously had an optional `render` prop](https://github.com/cloud-gov/pages-core/blob/staging/frontend/components/ExpandableArea.jsx#L47) which allowed for resetting the area + toggle state. However it seems like the toggle state was being double controlled by the React + USWDS (this was very confusing when refactoring, h/t to @hursey013 for the assist on finding it). So I've removed React control of the toggle to avoid having to partially remove or otherwise deactivate the USWDS control. This makes it so there isn't a great way to programmatically toggle
- Moving away from class components has at least one drawback and that's redefining function on each new set of props (rather than having a method with a bound class and access to props). React provides [linting rules to prevent this](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) which I've disabled. If rendering performance becomes an issue here, we should consider memoizing these functions (perhaps also via hooks)

## security considerations
None


frontend/components/ExpandableArea.jsx
```diff
@@ -5,62 +5,38 @@ import shortid from 'shortid';
...
-          {title}
-        </button>
-        <div
-          id={this.id}
-          className="usa-accordion-content"
-          aria-hidden={!isExpanded}
-        >
-          {isExpanded && ((render && render(this.toggle)) || children)}
-        </div>
+function ExpandableArea(props) {
```
apburnes, line 8: Was unable to show the `EnvironmentVariableForm` when testing locally.  Looks like the `render` prop from the `ExpandableArea` component is still being used in the `EnvironmentVariables` component.

apburnes (commented):
Added the comment about updating how the `EnvironmentVariables` component shows the `EnvironmentVariableForm` with the `ExpandableArea` area. Everything else seems to work as expected. Thanks!

frontend/components/ExpandableArea.jsx
drewbo: Updated to fix the rendering here. Note that (1) I still plan to update this component to functional in a later PR and (2) this does break the previous functionality of closing the expanded area after submission but it still seems like okay UX in this state.

apburnes (approved):
Looks good to go to me!

Issue #4007, opened 2022-12-21
Fix query to list site environment variables for site org users


apburnes:
An org member should be able to add, delete, and see a site's environment variables when the site is in the same organization as the member.

## Notes

- Current scope only allows users who have added the site through the `/sites/new` functionality

## Acceptance Criteria

- [ ] Update query to allow both site collaborators and site org members.

apburnes:
Closed by https://github.com/cloud-gov/pages-core/pull/4010

PR #4010, opened 2022-12-21
fix: Removes scope from UEV model in favor of Site scope forUser

apburnes:
Addresses Issue #4007 

- Removes `forSiteUser` scope from UserEnvironmentVariable model
- Get, Create, Delete routes use the Site model's `forUser` scope to determine if the user can take an action on the site's environment variable
- Adds tests for an org member's ability to take actions

## security considerations
- none


drewbo (approved):
Nice, this is an elegant solution (and avoids more scopes!)

PR #4013, opened 2022-12-23
Add remaining components rendered under /site as functional components

drewbo:
- Various components were updated to become functional components rather than class components
- Component tests were updated in response

## Pieces I'm worried about breaking:
- `SitePublishedFilesTable` was keeping previously fetched files in memory, now it's only storing the final key of each page. This will have a minor performance hit when using the "Previous" button
- I'm intentionally leaving `EnvironmentVariables` as a class component because it actually uses `mapDispatchToProps` with the dispatch variable and it doesn't rely on any router props (which would need to be replaced with `useParams` with the React router upgrade)

## security considerations
None

### DIFF
--- a/frontend/components/CreateBuildLink.jsx
+++ b/frontend/components/CreateBuildLink.jsx
@@ -1,34 +1,24 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-class CreateBuildLink extends React.Component {
-  constructor(props) {
-    super(props);
-
-    this.handleClick = this.handleClick.bind(this);
-  }
-
-  handleClick(event) {
+function CreateBuildLink(props) {
+  const {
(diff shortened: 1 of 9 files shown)


drewbo:
I cannot for the life of me figure out why [the function spy on mount test](https://github.com/cloud-gov/pages-core/pull/4013/files#diff-53b20a11dfbbae14d09a9bb0e465032675957b4c97127c88f58a9e4185e0e0afR40) isn't working when [seemingly identical tests](https://github.com/cloud-gov/pages-core/blob/chore-no-class-components-more-hooks/test/frontend/components/site/SitePublishedBranchesTable.test.jsx#L73) pass

apburnes (approved):
Looks good to me.

PR #4020, opened 2023-01-06
in non-production environments, use basic-vpc as the s3 plan

drewbo:
- in non-production environments, use basic-vpc as the s3 plan

## security considerations
Restricts S3 access (only in staging for now) to requests that originate from within the cloud.gov VPC


apburnes (dismissed):

api/services/SiteCreator.js
```diff
@@ -111,7 +113,9 @@ function buildInfrastructure(params, s3ServiceName) {
   return apiClient.createSiteBucket(
     s3ServiceName,
     config.env.cfSpaceGuid,
-    config.app.s3ServicePlanId
+    config.app.s3ServicePlanId,
+    null,
```
apburnes, line 117: I think function arguments need to be `undefined` and not null in order to have the default argument value be used . Would it make more sense putting the feature flag in the `createSiteBucket` function?

drewbo: Good call on `undefined`. I can move the check to the `createSiteBucket` function also

test/api/unit/utils/cfApiClient.create.test.js
```diff
@@ -172,6 +172,7 @@ describe('CloudFoundryAPIClient', () => {
         name, config.env.cfSpaceGuid, config.app.s3ServicePlanId, keyIdentifier, planName
       )
         .then((res) => {
+          console.log(res)
```
apburnes, line 175:
```suggestion
```

apburnes (commented):
Put a suggesstion in for a rouge `console.log` in the tests. Should be good to go after that.

apburnes (approved):
🪰

PR #4027, opened 2023-01-12
use basic-vpc as s3 plan in all environments

drewbo:
- Use the new `basic-vpc` s3 plan for bucket creation in all environments
- Note potential to deprecate use of `s3ServicePlanId` across the application

## security considerations
Adds ingress restrictions on S3 buckets


svenaas (commented):
Nice catch on removing the unneeded config require.

apburnes (approved):

PR #4030, opened 2023-01-18
React router v6

drewbo:
Per https://github.com/cloud-gov/pages-core/issues/3372#issuecomment-1352099621, this:
- updates a number of components to functional components to allow use of hooks
- upgrades to `react-router-v6` following many of the instructions here: https://reactrouter.com/en/main/upgrading/reach

## Things I've broken or have concerns about
This branch should undergo substantial checks on staging.
- There is occasionally flickering on first load because not all components wait for the main fetches to complete, I'm unsure of whether this was true previously
- For users without an organization, this branch intentionally turns off the "Add Site" button.
  - This will present a confusing interaction for non-org'd users because the button will instead read "Connect with Github" which won't help resolve their ability to add sites.
- Attempting to delete a site, doesn't wait for success prior to navigation so the alert shows on the next page. I believe this was previously the case (so not a reversion) but there are a number of examples of navigation being non-conditional on the action throughout the app.
- Many components still have the double export pattern (i.e. `export { A }, A`) even though it's no longer needed for testing. However it certain cases when removed, the tests failed.
- Breaks previously mentioned in #4002, #4004, and #4019

## security considerations
None


apburnes (commented):
Had problems loading the site build logs and through a comment in there. Also, I noticed a flashing when loading a user's site. _Sorry for the bad video work_ 



https://user-images.githubusercontent.com/693815/214385312-42149f79-d840-4eae-8748-add0c90abf25.mov

apburnes:
Almost every looks great and works as expected. When going into an organization to manage the users, I am getting `Nothing to see here` 404ish style screen. Is this the router not fetching the correct org info or our API?

image

apburnes (approved):
👏 👏 👏 It looks good to me. Thanks for taking this on!

PR #4040, opened 2023-02-01
[Dummy PR] Chore react 18

drewbo:
Just using this for better Github comparison features, will replace with something better


drewbo:
Incorporated as parts of #4030 and https://github.com/cloud-gov/pages-core/tree/react-18-upgrade

PR #4049, opened 2023-02-08
Change UAA email case sensitivity matching

### LINKED ISSUE #4043: Better handling for UAA email casing
From what I can tell:
- One customer's IdP sends back profile information with `email` cased which I don't think is great practice/common but we should definitely be able to handle this on our end.
- If we can't find a users email in our UAA users database, we show them an error that they aren't a valid user.
- We discovered this problem a few months ago and added a fix here: https://github.com/cloud-gov/pages-core/pull/3984. This lowercases the returned IdP profile value prior to comparing it to the database value
- However, if a user invite is sent with an uppercased email value, that will be directly added to the database, and then our previously updated comparison will fail (by trying to lowercase the profile information).
- For now I've resolved the immediately problem but manually lowercasing the database email values of the affected users
- To properly fix this, we should either:
  - Lowercase all values prior to comparison
  - Migrate this column to a [`citext`](https://www.postgresql.org/docs/current/citext.html) type, so all comparisons on this value are made case-insensitive.

## Acceptance Criteria

- [ ] Users UAA email comparisons are made entirely independently of their casing

drewbo:
- use case insensitive check for UAA email, close #4043

## security considerations
Assumes mail servers are not strictly following RFC 5431 as noted in the referenced issue discussion.

### DIFF
--- a/api/services/uaaStrategy.js
+++ b/api/services/uaaStrategy.js
@@ -1,4 +1,5 @@
 const { Strategy } = require('passport-oauth2');
+const { Sequelize } = require('sequelize');
 const UAAClient = require('../utils/uaaClient');
 const { Event, UAAIdentity, User } = require('../models');
 const EventCreator = require('./EventCreator');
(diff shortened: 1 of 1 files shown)


apburnes (approved):

svenaas:
You **just** beat me to it, Andrew.

PR #4095, opened 2023-04-03
Admin Rebuild Build

### LINKED ISSUE #4083: Admins should be able to retrigger a build from the Admin UI
Right now the rebuild option is only available in the user-facing application

## Acceptance Criteria

- [ ] Admins can retrigger a build from the Admin UI

drewbo:
- add rebuild build option to admin panel
- close #4083

## Notes
- The controller code is almost purely duplicative of the api controller code (that is the case for other functions as well) but I didn't want to break the pattern of accessing a non-admin controller from the admin api. 
- When attempting to test locally (1) I needed to temporarily remove the Github status reporting to get the button press to succeed but (2) the build fails with `There was an error, adding the job to SiteBuildQueue: Error: Unable to find vali...` which is likely another Github related error (no valid build token)`. It's possible that this is all irrelevant on staging.

## security considerations
New admin endpoint for accessing builds


drewbo:
Rethinking this slightly, I think the error is informative:
- if an admin triggers a build, they likely have a github token but it's not necessarily going to fetch the repo and the rebuild will error
- we need to (1) trigger the build with a dummy user with no token, (2) trigger the build with a previous build user with a valid token or (3) rewrite the `GithubBuildHelper` service logic with updated logic to handle admin users

apburnes (approved):

PR #4123, opened 2023-04-25
Admin panel filter fix

drewbo:
- avoid inline styles for the admin panel filters

## security considerations
None


apburnes (dismissed):

drewbo:
The `Svelecte` element is also trying to apply its own inline-styles which are blocked by our CSP policy so the input area is still restricted:

Screen Shot 2023-04-25 at 11 54 53 AM

I'm going to see if I can make any progress with a hash/nonce to allow the style (in this case the variable input width) to go through

apburnes (approved):

Issue #3718, opened 2021-12-03
FederalistUsersHelpers service relies on FEDERALIST_USERS_ADMIN env variable


amirbey:
#### User Story

<TODO>

#### Background (Optional)

FederalistUsersHelpers service relies on FEDERALIST_USERS_ADMIN env variable.  This should dependency should be removed/re-evaluated.  _Note: The need for this service and associated scheduled jobs should go away when organizations are introduced in production._

#### Acceptance Criteria

- [ ] Removes FederalistUsersHelper functions

apburnes:
Closed by https://github.com/cloud-gov/pages-core/pull/4131

PR #4131, opened 2023-05-03
chore: Remove deprecated federalistUsersHelpers function

apburnes:
Related to #3718

- Removes defunct scripts to manage deprecated federalist GitHub org
- Removes tests and functions manages deprecated federalist GitHub org

## security considerations
Removing deprecated service functions


api/authorizers/site.js
```diff
@@ -51,7 +51,7 @@ const authorizeRepositoryAdmin = (user, site) => (
 );
 
 function authorizeFederalistUsersAdmin(user) {
```
drewbo, line 53: Do we still need this function (and the `UsersHelper` service) at all? The only place I see it used is when destroying a site, we check that it's a repo admin first, and then fallback to checking if it's a federalist-users org admin on error. I think we should eliminate it entirely (and potentially replace that second check with "is a Pages admin" in the future)

drewbo (changes requested):
I might be missing something but I vote we remove the UsersHelper service entirely. It looks like it's only used to check federalist-users admin status as a fallback to a few other commands

api/authorizers/site.js
apburnes: Removed `UsersHelper` and the associated admin functionality in the app routes. This update will not affect any of the admin routes.

drewbo (approved):

PR #4138, opened 2023-05-09
feat: Add fail stuck builds queue and breakout timeout builds queue

apburnes:
Related to https://github.com/cloud-gov/pages-core/issues/4026

- Adds queue to fail stuck builds that are in created or tasked for over 10 minutes
- Breaks out the timeout build tasks queue for transparency

## security considerations
None


drewbo:
I imagine this is hard to reproduce on command but I ran a build on dev to make sure everything was normal. Code looks good and concourse test failure looks like 🌞 🔗

Issue #4118, opened 2023-04-20
Add an admin view to see the number of production sites per org


apburnes:
Allow our platform operators the ability to view and export our production sites per organization.

## Acceptance Criteria
- [x] Create admin api endpoints that returns a list of all productions with org, site, and domain info
- [x] Create a new admin page to view all orgs and their production sites
- [x] Enable filtering by organization, following the filter by org example at `/users/`
- [x] Calculate the number of production sites by getting all of the associated site domains per org
- [x] Add a simple CSV download 
- [x] Create a new Reports item in header nav and an initial Reports page where this new report will be the first one listed and linked.

## Security Considerations
none

svenaas:
@apburnes The SQL is straightforward enough but I need to clarify the semantics of "**production** sites" in the first checkbox item above. Is it

1. All sites with a `site.domain`?
2. All sites with a `site.domain` and `active` = true?   
3. All sites with a `site.domain` and a `publishedAt` timestamp?
4. All sites with an associated `domain` record where `state` = ` provisioned`?

Or something even more specific?

svenaas:
@apburnes What else would you like to see in this table?
<img width="1115" alt="image" src="https://github.com/cloud-gov/pages-core/assets/12150/d356a28f-8ad2-4e6e-917c-06099e5a61e3">

apburnes:
I thinks that's a good start. 👍

svenaas:
Note that the displayed count of sites is literally a count of Sites. Is this correct, or should it be a count of the associated Domains (such that a Site with two provisioned Domains counts as two published sites), or, going further, should it be a count of the potentially multiple domains listed in each provisioned Domain's `names`?

svenaas:
Got the download button working, @apburnes. Do you want a demo of how this works at the moment?

Open questions:

1. Per my comment immediately above, am I calculating the number of sites the way you intended?
2. Where should the report page live in the admin client nav? So far I haven't linked it anywhere. 
3. There's one checkbox item up top that I haven't touched: "Create admin api endpoints that returns a list of all productions sites per org with org, site, and domain info". Does this mean just the production sites for a single org?

svenaas:
We decided in demo today that there should be one result row per Domain, so a Site may show up several times if it has multiple domains associated with it. In addition to the query change it will make more sense for this to live on the Site controller instead of the Domain controller. Moving it is going to take a little unwinding and I want to get my in-process admin client changes (I'm working on the filterable table) to a commitable state first.

PR #4145, opened 2023-05-25
Add an admin view to see the number of production sites per org

svenaas:
This pull request addresses #4118 

- Adds an option to PaginatedQueryPage to omit the search form
- Adds API endpoint `/organizations-report` which returns a list of all production sites by organization.
- Adds API endpoint `/organizations-report.csv` which returns the same results in CSV format.
- Adds admin client page `/organizations-report` which presents the list returned from the new API endpoint.
- Implements filtering the admin report view by organization
- Implements downloading the full CSV from the admin report view
- Adds admin client page and nav item `/reports` which establishes a home for this and future admin reports

## security considerations
None. The data presented is only available to authenticated admin/support users, and only represents data already available to those users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "keywords": [],
   "dependencies": {
     "@bull-board/express": "4.11.3",
+    "@json2csv/plainjs": "^7.0.1",
     "@octokit/request": "^6.2.2",
     "@octokit/rest": "^18.5.2",
     "@socket.io/redis-adapter": "^7.0.0",
@@ -37,6 +38,7 @@
     "js-file-download": "^0.4.11",
     "js-yaml": "^4.1.0",
     "json-to-csv": "^1.0.0",
+    "json2csv": "juanjoDiaz/json2csv#v7.0.1",
     "jsonwebtoken": "^8.5.1",
     "moment": "^2.29.2",
     "newrelic": "^8.13.2",
--- a/admin-client/src/Router.svelte
+++ b/admin-client/src/Router.svelte
@@ -66,6 +66,8 @@
   page('/organizations/new', queryString, render(Pages.Organization.New));
   page('/organizations/:id', queryString, render(Pages.Organization.Show));
   page('/organizations/:id/edit', queryString, render(Pages.Organization.Edit));
+  page('/organizations-report', queryString, render(Pages.Organization.Report));
+  page('/reports', queryString, render(Pages.Reports));
   page('*', render(Pages.NotFound));
   page();
 </script>
--- a/admin-client/src/components/Nav.svelte
+++ b/admin-client/src/components/Nav.svelte
@@ -108,6 +108,11 @@
             </a>
           </li>
           <li class="usa-nav__primary-item">
+            <a class="usa-nav__link"  class:usa-current={currentPath === '/reports'} href="/reports">
+              <span>Reports</span>
+            </a>
+          </li>
+          <li class="usa-nav__primary-item">
             <NavButton action={logout}>Logout</NavButton>
           </li>
         {:else}
--- a/admin-client/src/components/PaginatedQueryPage.svelte
+++ b/admin-client/src/components/PaginatedQueryPage.svelte
@@ -10,6 +10,7 @@
   export let path;
   export let query;
   export let addAction = false;
+  export let noSearch = false;
   export let fields = {};
   export let title = null;
   export let expanded = false;
@@ -117,6 +118,7 @@
             </li>
           </ul>
         {/if}
+        {#if ! noSearch}
         <form
           class="usa-search usa-search--small flex-align-center"
           role="search"
@@ -147,6 +149,7 @@
             <span class="usa-sr-only">Search</span>
           </button>
         </form>
+        {/if}
       </nav>
     </div>
     <div class="tag-container usa-nav-container margin-top-1 margin-bottom-2">
--- a/admin-client/src/components/PaginatedQueryPage.svelte
+++ b/admin-client/src/components/PaginatedQueryPage.svelte
@@ -118,7 +118,7 @@
             </li>
           </ul>
         {/if}
-        {#if ! noSearch}
+        {#if !noSearch}
         <form
           class="usa-search usa-search--small flex-align-center"
           role="search"
--- a/admin-client/src/lib/api.js
+++ b/admin-client/src/lib/api.js
@@ -58,6 +58,36 @@ async function _fetch(path, fetchOpts = {}, apiOpts = { handleError: true }) {
   return request;
 }
 
+// eslint-disable-next-line no-underscore-dangle
+async function _fetchAttachedFile(path) {
+  const fetchOptions = { ...defaultOptions };
+  fetchOptions.headers['x-csrf-token'] = session.csrfToken();
+
+  let request = fetch(`${apiUrl}/admin${path}`, fetchOptions)
+    .then(async (r) => {
+      if (r.ok) return r.text();
+      if (r.status === 401) {
+        authLogout();
+        return null;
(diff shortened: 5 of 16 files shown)


Ephraim-G:
The only other item that would be extremely beneficial going forward would be to add a scope to pull in users via email rather than username for readability (+ their roles if possible) so we can periodically check for changes and then update things on the Zendesk side for comms. Other than that this looks fantastic @svenaas , very nice!

svenaas:
@Ephraim-G Do you mean you'd like a column containing user email / role information in the CSV and the onscreen report? Is there a particular format for that data that would be most convenient?

Ephraim-G:
@svenaas Think CSV is fine for now I don't expect many changes if we pull this down 1x per month, the most compatible format is YY/MM/DD or YYYY-MM-DD per zendesk. Also for clarity's sake if implemented users will be grouped by org correct?

svenaas:
@Ephraim-G I'm afraid we might not be understanding each other. As it currently stands the report introduced in this PR looks like this:
image

The downloadable CSV has identical columns:
```
"Organization","Agency","Site","Domain","Engine"
"agency1","GSA","example-node-site","foo.example.gov","node.js"
"agency3","Bureau of Testing","another-example-hugo-site","bar.example.gov","hugo"
"agency4","Demonstration Department","another-example-node-site","qux.example.gov","node.js"
"agency4","Demonstration Department","another-example-node-site","vanity.qux.example.gov","node.js"
,,"an-orgless-node-site","orgless.example.gov","node.js"
```

I thought you were asking suggesting adding an additional column or columns containing user email and role information, and was asking how you'd want the data in that column formatted. But I'm not sure what you're specifying the date format for, and your reference to grouping users by org suggests to me that you might be asking for another new report rather than changes to this one. Am I misunderstanding?

svenaas:
Ephraim and I just checked in on a call; I believe the user/role aspect doesn't need to hold up this initial version of the report. 

For #4146 it might be turn out that we could incorporate the user/role stuff in this report where the list has one row per domain, but it might be more advantageous to create a different report with one row per organization, or one with one row per user. We can address that in a subsequent round.

admin-client/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "date-fns": "^2.15.0",
+    "js-file-download": "^0.4.11",
```
apburnes, line 33: Instead of pulling this in as a dependency, is this simple enough to write ourselves?

package.json
```diff
@@ -36,6 +36,7 @@
     "ioredis": "^4.27.6",
     "js-file-download": "^0.4.11",
     "js-yaml": "^4.1.0",
+    "json2csv": "^3.4.2",
```
apburnes, line 39: This dependency has been abandoned. https://www.npmjs.com/package/json2csv

Can we write this ourselves?

apburnes (commented):
Two questions on the added dependencies and if we should write the functionality ourselves.

package.json
svenaas: We already had this dependency through the similarly-named `json-to-csv`, which is required on the next line. I added `json2csv` here to make it explicit, because I am requiring it directly. `yarn.lock` didn't have to change with this change to `package.json`.

The earlier dependency, `json-to-csv`, is used by `scripts/exportSitesAsCsv.js`. The reason I didn't just use the same approach is that the `json-to-csv` library writes to an output file, whereas for this use case we needed the CSV content as a string to serve through the CSV. The underlying `json2csv` library provided this already to `json-to-csv` so I went ahead and used it.

We can certainly render our own CSV in either or both of our existing use cases, but we have to write our own code and suitable tests for the stuff (quoting, escaping, headers, missing fields) that's already implemented. It's a tradeoff.

svenaas: Honestly, the `json-to-csv` package we were already directly depending to write an output file consists of very little code and could be easily replaced.

But the `json2csv` library [has a lot more going on](https://www.npmjs.com/package/json2csv?activeTab=code) with formatters for different field types and other logic. We don't need most of that in each of our specific uses but if we're going to be adding more reports in the future I think sooner or later we'll end up re-solving a lot of details that a decent library will have already addressed.

Potential approach: we could move to an actively maintained library ([example](https://www.npmjs.com/package/json-2-csv)) for the JSON/CSV logic, and write our own code for the file output stage that's needed by `scripts/exportSitesAsCsv.js`. We just have to decide whether we do part or all of that on this PR.

admin-client/package.json
svenaas: @drewbo is confident that new browser APIs make this approach unnecessary, but uncertain whether we still have to worry about the Safari issue noted in the code I linked immediately above.

package.json
svenaas: @apburnes As of 5c0f55d this PR is dependent on the current (v7.0.1) version of the maintained, GitHub-hosted version of `json2csv`. 

The `exportSitesAsCsv.js` script is still indirectly dependent on the legacy version. I'll be glad to tackle fixing that too but I don't think it belongs on this PR. If it's all right with you I'll make a separate issue for that.

apburnes: I see. Can you update the dependency to the new package names with `@json2csv/transforms` and `@json2csv/plainjs` like the example in https://www.npmjs.com/package/@json2csv/transforms?

admin-client/package.json
apburnes: It looks like all the [relevant browser versions handle the download attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download#browser_compatibility). Do you want to try out writing the functionality so we don't need the added dependency?

svenaas: I'm already in the process of doing that for this PR. 

But note that `js-file-download` is also required in the React app where it's an existing dependency of the `downloadBuildLogsButton` component. If we'd like to do away with it completely I think that's another issue (which I'd be glad to tackle).

svenaas: fd0d58d removes this PR's new admin client dependency on `js-file-download` (and has been tested successfully in Safari 16.5/Mac as well as in Chrome).

package.json
svenaas: I'm not clear on what you're asking for, @apburnes. I'm not using any of the transforms so I'm only requiring the `plainjs` package. If you mean you'd like me to use the import syntax instead of require, I'm afraid I tried that and got `SyntaxError: Cannot use import statement outside a module` which @drewbo concluded was probably because we don’t bundle our API to support imports.

apburnes: If we just need `@json2csv/plainjs` functionality then add that dependency and require it by

```js
const json2csv = require('@json2csv/plainjs');
const parser = new json2csv.Parser(opts);
const csv = parser.parse(data);
```

admin-client/package.json
apburnes: Awesome.

package.json
svenaas: I'm still a little confused because I got that 
```js
{ Parser } from '@json2csv/plainjs'
```
syntax from the docs you linked to.

But I have now changed (d617c4d) the latest code here from 
```js
const { Parser } = require('@json2csv/plainjs');
const parser = new Parser({ fields });
const csv = parser.parse(domains);
```
to
```js
const json2csv = require('@json2csv/plainjs');
const parser = new json2csv.Parser({ fields });
const csv = parser.parse(domains);
```
and it continues to function.

apburnes: `const { Parser } from '@json2csv/plainjs'` is es module syntax. The app currently only supports commonjs syntax like `const { Parser } = require('@json2csv/plainjs');`.

apburnes (approved):
👍

svenaas:
Actually, no re-review necessary. Merging.

PR #4156, opened 2023-06-16
Fix organization scope on Build model.

svenaas:
This addresses #4155 

I based the Domain organization scope in #4145 on the existing one on the Build model and in unit testing I discovered the existing approach does not work as intended. 

When a scope matches on a nested association using this approach:
```
include: [{
  model: AssociatedModel,
  include: [{
    model: NestedAssociatedModel,
    where: { id },
  }],
}],
```
it returns both those instances of the model where the nested associated model matches on `id` _and_ those instances where it does not — for the latter case simply omitting the columns of both the nested and the directly associated model. 

It is not enough simply to add `required: true` on the nested associated model like
```
include: [{
  model: Site,
  include: [{
    model: Organization,
    required: true,
    where: { id },
  }],
}],
```

A `required: true` must be declared on the directly associated model as well:
```
include: [{
  model: Site,
  required: true,
  include: [{
    model: Organization,
    required: true,
    where: { id },
  }],
}],
```

I believe no code calls the Build organization scope at this time — including, prior to this PR, any test code. 

- Add a unit test for the Build organization scope which shows the bug in the absence of the new
- Fix to the organization scope on the Build model

## security considerations
None. This fixes a lurking bug in code that is not yet being called.

### DIFF
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -42,8 +42,10 @@ const associate = ({
   Build.addScope('byOrg', id => ({
     include: [{
       model: Site,
+      required: true,
       include: [{
         model: Organization,
+        required: true,
         where: { id },
       }],
     }],
--- a/test/api/unit/models/build.test.js
+++ b/test/api/unit/models/build.test.js
@@ -350,6 +350,30 @@ describe('Build model', () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


apburnes (approved):
👍

PR #4167, opened 2023-07-03
Remove dependencies on json-to-csv and abandoned json2csv version

### LINKED ISSUE #4159: Remove dependency on abandoned json2csv version and on json-to-csv
The abandoned npm version of `json2csv` is required by the `json-to-csv` package used by `scripts/exportSitesAsCsv.js`. We can write our own file output code to eliminate the need for `json-to-csv` and then switch to the current, maintained version of `json2csv` using the approach utilized in #4145.

## Acceptance Criteria

- [ ] dependency on `json-to-csv` eliminated
- [ ] indirect dependency on npm `json2csv` (3.4.2) eliminated.
- [ ] `yarn export:sites` still works as expected.

svenaas:
Fixes #4159 

- Removed dependency on the `json-to-csv` library that saves JSON data to a local file.
- Replaced the removed library's dependency on abandoned version of `json2csv` with code that utilizes the current `json2csv` used elesewhere in `pages-core`

## Testing
On local dev compare the output of `yarn export:sites` against the `staging` branch and against this branch — they should be identical (`docker-compose run --rm app yarn export:sites` generates `current-sites.csv`) — apart, depending on how you do it, from primary keys.

## security considerations
None. This replaces an unneeded dependency and its outdated underlying dependency with a currently-maintained library.

### DIFF
--- a/scripts/exportSitesAsCsv.js
+++ b/scripts/exportSitesAsCsv.js
@@ -1,8 +1,9 @@
 /* eslint no-console: 0 */
 const path = require('path');
+const fs = require('fs');
 
 const _ = require('underscore');
-const jsonToCSV = require('json-to-csv');
+const json2csv = require('@json2csv/plainjs');
 
 const { User, Site } = require('../api/models');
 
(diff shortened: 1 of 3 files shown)


scripts/exportSitesAsCsv.js
```diff
@@ -40,6 +41,21 @@ function consolidateOnSiteId(sites) {
   });
 }
 
+function writeCSV(sites, fileName) {
+  const fields = ['id', 'github', 'domain', 'users'];
```
drewbo, line 45: To futureproof (and match the functionality of the prior library), can you get the fields dynamically rather than hardcoding

drewbo (changes requested):

scripts/exportSitesAsCsv.js
svenaas: Yes. I had that and removed it for declarative simplicity but I'm happy to go the other way.

svenaas: Dynamic field implementation pushed. 

The `json-to-csv` library [used](https://www.npmjs.com/package/json-to-csv?activeTab=code) the Ramda library, effectively calling `Ramda.keys(sites[0])` to determine the fields. I'm using `Object.keys(sites[0])` to avoid that dependency. And unlike the old library I'm doing it inside the try/catch block in case `[]` or `keys()` raise an error.

drewbo (approved):

PR #4197, opened 2023-07-20
chore: automate changelog creation on release

drewbo:
- Adds a new pipeline which does the following:
  - Checks out `main`
  - Tries to run [commitizen](https://commitizen-tools.github.io/commitizen/), specifically `cz bump` which automatically increases the version, based on the commits since the last version.
  - If this succeeds, creates/updates the `release` branch and a corresponding PR
- This PR only completes the first half of the work described in https://github.com/cloud-gov/product/issues/2563 (creating the release PR), the next step will be having our production deployments trigger only on tags and include a github release

## security considerations
Concourse uses a Github bot (and associated credentials) to make release commits to this repo

### DIFF
--- a/.cz.json
+++ b/.cz.json
@@ -1,0 +1,47 @@
+{
+  "commitizen": {
+    "name": "cz_customize",
+    "version_scheme": "semver",
+    "version_provider": "npm",
+    "update_changelog_on_bump": true,
+    "major_version_zero": false,
+    "bump_message": "chore: release $new_version",
+    "gpg_sign": true,
+    "changelog_incremental": true,
+    "customize": {
+      "message_template": "{{change_type}}:{% if show_message %} {{message}}{% endif %}",
+      "example": "feat: this feature enable customize through config file",
+      "schema": "<type>: <body>",
+      "schema_pattern": "^((build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test){1}(\\([\\w\\-\\.]+\\))?(!)?: ([\\w \\-'])+([\\s\\S]*))|Merge.*",
+      "bump_pattern": "^(.+!|BREAKING CHANGE|chore|docs|feat|fix|perf|refactor|revert|style|test):",
+      "bump_map": {
+          ".+!": "MAJOR",
+          "BREAKING CHANGE": "MAJOR",
+          "feat": "MINOR",
+          "fix": "PATCH",
+          "chore": "PATCH",
+          "docs": "PATCH",
+          "perf": "PATCH",
+          "refactor": "PATCH",
+          "revert": "MINOR",
+          "style": "PATCH",
+          "test": "PATCH"
+      },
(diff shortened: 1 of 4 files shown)


.cz.json
```diff
@@ -0,0 +1,48 @@
...
+          "revert": "MINOR",
+          "style": "PATCH",
+          "test": "PATCH"
+      },
+      "change_type_order": ["BREAKING CHANGE", "feat", "fix", "perf", "docs", "revert", "chore", "refactor", "style", "test"],
+      "info_path": "cz_customize_info.txt",
+      "info": "This is customized info",
+      "commit_parser": "^(?P<change_type>chore|docs|feat|fix|perf|refactor|revert|style|test):\\s(?P<message>.*)?",
+      "changelog_pattern": "^(chore|docs|feat|fix|perf|refactor|revert|style|test)?(!)?",
+      "change_type_map": {
```
apburnes, line 35: Should the `change_type_map` and the other change related key values match the `bump_map` values? Is there a situation where want to use the `hotfix` keyword to affect a bump but not the changelog?

ci/pipeline-release-deploy.yml
```diff
@@ -0,0 +1,59 @@
...
+#  RESOURCES
+
+resources:
+  - name: src
+    type: git
+    icon: github
+    source:
+      uri: git@github.com:cloud-gov/pages-core.git
+      branch: ((git-branch))
+      commit_verification_keys: ((cloud-gov-pages-gpg-keys))
```
apburnes, line 58: Is there any documentation or a runbook on best practices to set or update certificates in credhub?

apburnes (commented):
This is awesome! Great to see it come together. I added a couple of question to help me understand better.

.cz.json
drewbo: This is a good question, I think these differences probably aren't necessary and we should match our schema pattern almost exactly (leaving out `build` to trigger builds without changing version). Breaking changes can be indicated with `!` after the type and hotfixes can be tagged normally (e.g. `fix`).

ci/pipeline-release-deploy.yml
drewbo: I got personal support from @soutenniza but agree that this could use a runbook somewhere

apburnes: Sounds good. I'll make an issue for the platform squad to put together a runbook for it.

.cz.json
drewbo: Actually this isn't exactly true, conventional commits spec says we should major bump on either `!` or `BREAKING CHANGE` in the footer and I'll need to add a custom rule for that, working on it now

drewbo: The updated version should correctly handle breaking changes, indicated either via the footer or with `!` with one small exception: if someone uses the `feat!` notation with no footer, it will end up in the `Added` section rather than `Breaking Changes` but I can live with that for now

apburnes (approved):

.cz.json
apburnes: Sounds good.

Issue #4191, opened 2023-07-17
Fix dev tests to not build assets in version controlled directory


apburnes:
When a developer is running the tests locally assets for the app frontend get built and put into a version controlled directory. Update the asset build process for tests to not use a version controlled directory.

## Acceptance Criteria

- [ ] No longer builds test assets into `public/images/` directory
- [ ] Builds assets and places them in a directory that is .gitignore
- [ ] Running tests against local dev environment does not place any built assets into version controlled directories

svenaas:
Well, boy, this will be nice! I'll be able to do away with
```
alias clean-images="rm -rf public/images && git checkout -- public/images"
```

svenaas:
`yarn build` (which is called by `test:prepare`) puts 95 files in `public/images` which show up with `git status` and could be accidentally committed. 

`test:client` doesn't seem to generate any assets. Neither `test:client` nor `test:server` call `test:prepare`. But `test` does, as does `test:cover`.

The 95 files are a bunch of front end assets:
- `.png   (35)`
- `.svg   (31)`
- `.eot   ( 8)`
- `.ttf   ( 8)`
- `.woff  ( 8)`
- `.woff2 ( 8)`

None of these seem to generate any **CSS** assets that show up with `git status`, however, so I wanted to double-check that detail. @apburnes, are you aware of steps to reproduce CSS generation in particular? If not I can reframe this task a little bit ("when a developer is running the tests locally assets for the app frontend get built" etc.).

apburnes:
That's what we want. Fix dev tests to not add the build assets in version controlled directory. I guess there's no css files but all the others should go somewhere outside of version control.

svenaas:
Yup, I understand that. I just wanted to be sure you weren't aware of some other situation that lead to CSS files. I'll amend the description a little.

svenaas:
I stand corrected. It creates plenty of CSS and JS files too. It's just that _those_ were already gitignored.

svenaas:
Holding off until #4208 is merged because that's going to change the Webpack configuration out from under this.

PR #4208, opened 2023-07-27
chore: release on tag

drewbo:
- This PR removes the `git-branch` and `product` instance variables and explicitly defines three different CI pipelines for this repo:
  - `pipeline-dev.yml`: largely the same as before with modularized task definitions (added to `ci/partials`). The `yarn build-dev` command was removed in favor of using `NODE_ENV` to define the build changes and that same env variable is passed to the admin client build as well. 
  - `pipeline-staging.yml`: the same as before but with modularized task definitions from above. The `deploy-env` is removed from the names of PR tests here because they are now deploy env independent. 
  - `pipeline-production.yml`: removes the testing of PRs (since all PRs to `main` are tested in the prior pipeline). Incorporates the release PR management which was previously added as a separate pipeline

## security considerations
None


rcgottlieb (approved):

DEVELOPMENT.md
```diff
@@ -163,7 +163,7 @@ To deploy to CloudFoundry submit the following:
 `cf push federalistapp --strategy rolling --vars-file "./.cloudgov/vars/${CF_SPACE}.yml" -f ./cloudgov/manifest.yml`
 
 ### Continuous Integration
-We are in the process of migrating from CircleCI to an internal instance of Concourse CI, starting with our staging environment. To use Concourse, one must have appropriate permissions in UAA as administered by the cloud.gov operators. Access to Concourse also requires using the GSA VPN.
+Our continuous integration pipeline is run on Concourse CI. To use Concourse, one must have appropriate permissions in UAA as administered by the cloud.gov operators. Access to Concourse also requires using the GSA VPN.
```
svenaas, line 166: > Access to Concourse also requires using the GSA VPN.

Maybe it's beyond the scope of this documentation, but Zscaler changes this, or is at least in the process of doing so.

DEVELOPMENT.md
```diff
@@ -172,26 +172,29 @@ We are in the process of migrating from CircleCI to an internal instance of Conc
...
+
+> __Core Dev__ deploys the Pages app/api, the admin app, and the queues app when a PR is created into the `main` branch. This uses a unique pipeline file: [./ci/pipeline-dev.yml](./ci/pipeline-dev.yml)
 
 #### Core deployment
 ##### Pipeline instance variables
-Three instances of the pipeline are set for the `pages dev`, `pages staging` and `pages production` environments. Instance variables are used to fill in Concourse pipeline parameter variables bearing the same name as the instance variable. See more on [Concourse vars](https://concourse-ci.org/vars.html). Each instance of the pipeline has three instance variables associated to it: `deploy-env`, `git-branch`. `product`
+Three instances of the pipeline are set for the `pages dev`, `pages staging` and `pages production` environments. Instance variables are used to fill in Concourse pipeline parameter variables bearing the same name as the instance variable. See more on [Concourse vars](https://concourse-ci.org/vars.html). Each instance of the pipeline has one instance variables associated to it: `deploy-env`.
```
svenaas, line 193:
```suggestion
Three instances of the pipeline are set for the `pages dev`, `pages staging` and `pages production` environments. Instance variables are used to fill in Concourse pipeline parameter variables bearing the same name as the instance variable. See more on [Concourse vars](https://concourse-ci.org/vars.html). Each instance of the pipeline has one instance variable associated to it: `deploy-env`.
```

DEVELOPMENT.md
```diff
@@ -202,7 +205,7 @@ Some credentials in this pipeline are "compound" credentials that use the pipeli
 --- | --- | --- |
 |**`((deploy-env))-cf-username`**|The deployment environments CloudFoundry deployer username based on the instanced pipeline|:white_check_mark:|
 |**`((deploy-env))-cf-username`**|The deployment environments CloudFoundry deployer password based on the instanced pipeline|:white_check_mark:|
-|**`((deploy-env))-((product))-domain`**|The deployment envinronment and product(Pages|Federalist) specific domain for the app|:white_check_mark:|
+|**`((deploy-env))-pages-domain`**|The deployment envinronment specific domain for the app|:white_check_mark:|
```
svenaas, line 208:
```suggestion
|**`((deploy-env))-pages-domain`**|The deployment environment specific domain for the app|:white_check_mark:|
```

DEVELOPMENT.md
```diff
@@ -202,7 +205,7 @@ Some credentials in this pipeline are "compound" credentials that use the pipeli
 --- | --- | --- |
 |**`((deploy-env))-cf-username`**|The deployment environments CloudFoundry deployer username based on the instanced pipeline|:white_check_mark:|
```
svenaas, line 206: Either "development environment's" or "development environment".

DEVELOPMENT.md
```diff
@@ -202,7 +205,7 @@ Some credentials in this pipeline are "compound" credentials that use the pipeli
...
 |**`((deploy-env))-cf-username`**|The deployment environments CloudFoundry deployer password based on the instanced pipeline|:white_check_mark:|
```
svenaas, line 207: Either "development environment's" or "development environment".

svenaas (changes requested):
Made some copyediting suggestions.

DEVELOPMENT.md
drewbo: Skipping these because they aren't part of this PR (will catch on a larger docs review)

ci/pipeline-dev.yml
```diff
@@ -386,7 +268,7 @@ resources:
     source:
       repository: ((pages-repository-path))
       access_token: ((gh-access-token))
-      base_branch: ((git-branch))
+      base_branch: main
```
apburnes, line 271: Should we set this to `staging` until we make the switch to a single branch and tag deployment? Or should we update this PR to point to `main`?

ci/pipeline-production.yml
```diff
@@ -0,0 +1,434 @@
...
+              :white_check_mark: SUCCESS: Successfully deployed pages queues UI on pages ((deploy-env))
+              <$ATC_EXTERNAL_URL/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME?vars.deploy-env="((deploy-env))"|View build details>
+            channel: ((slack-channel))
+            username: ((slack-username))
+            icon_url: ((slack-icon-url))
+
+  - name: nightly-tasks-pages-((deploy-env))
+    plan:
+      - get: src
+        resource: src-((deploy-env))
```
apburnes, line 261: Is there a reason we aren't pulling the tagged release?

ci/pipeline-production.yml
```diff
@@ -0,0 +1,434 @@
...
+
+  - name: src-((deploy-env))-tagged
+    type: git
+    icon: github
+    source:
+      uri: ((git-base-url))/((pages-repository-path))
+      branch: main
+      commit_verification_keys: ((cloud-gov-pages-gpg-keys))
+      tag_filter: 0.*.*
+      fetch_tags: true
```
apburnes, line 364: Do we need both resources or can it just be the tagged resource?

ci/pipeline-production.yml
```diff
@@ -0,0 +1,434 @@
...
+      private_key: ((pages-gpg-operations-github-sshkey.private_key))
+
+  - name: src-((deploy-env))-tagged
+    type: git
+    icon: github
+    source:
+      uri: ((git-base-url))/((pages-repository-path))
+      branch: main
+      commit_verification_keys: ((cloud-gov-pages-gpg-keys))
+      tag_filter: 0.*.*
```
apburnes, line 363: Can we update the docs on how to pin the resource in Concourse? See https://concourse-ci.org/resource-versions.html#version-pinning

webpack.production.config.js
apburnes, line 1: Was this update related to CI tasks or clean up you came across?

apburnes (commented):
Looks really good and much easier to read. The partials should work out well. Added a couple of comments and questions for clarification. Overall all a great step in the right direction.

ci/pipeline-dev.yml
drewbo: Yeah let's update this PR to point to main (even if the tests here aren't super meaningful against this code, we'll at least see the dev deploy try to run)

ci/pipeline-production.yml
drewbo: I can't remember my logic for including untagged but in retrospect it seems wrong. I'll update nightly/builds-checks to use tagged

webpack.production.config.js
drewbo: It's related: we previously had two commands `yarn build` and `yarn build-dev` with slightly different webpack configs. This was introduced to support the dev build options like nicer error tracing. To modularize the task, I wanted to share the same command (now encapsulated as a task in `ci/partials/build-api.yml`) and webpack config file but configure the differences with an environment variable (`NODE_ENV`). 

Separately, I think it would be even better practice to combine development and production webpack configs into one file and make it even clearer how environment variables swap between different options but I decided that was out of scope for this PR.

svenaas: This is of immediate interest to me because of #4191. I noticed that `build-dev` exists but isn't utilized in `test:prepare`. Also we have both `webpack.development-build.config.js` (used by `build-dev`) and `webpack.development.config.js` (not referenced at all in `package.json`).

drewbo: @svenaas `build-dev` is specifically for the concourse dev deploy. `webpack.development.config.js` is used during our local build (but is included via middleware, not via `--config` in the CLI)

apburnes (approved):
There's a lot here but the structure solid and dev ci is working as expected.  Thanks for handling this work!

PR #4219, opened 2023-08-04
Build dev/test webpack assets in non-version-controlled directory

### LINKED ISSUE #4191: Fix dev tests to not build assets in version controlled directory

svenaas:
Fixes #4191

- Add webpack test config that puts assets in existing gitignored `/dist/` instead of `/public/`
- Add `build:test` task that uses the test webpack config
- Use `build:test` instead of `build` from `test:prepare`

## security considerations
None. This is a dev/test change to reduce the risk of accidentally committing things that needn't be committed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
   },
   "scripts": {
     "build": "webpack --config webpack.production.config.js",
+    "build:test": "webpack --config webpack.test.config.js",
     "start": "node index.js",
     "start-workers": "node ./api/workers/index.js",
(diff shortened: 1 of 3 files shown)


drewbo (approved):
Worked for me locally; good to go once rebased

PR #4245, opened 2023-08-30
Feat build task models

drewbo:
- Add build task DB models + migrations

## security considerations
None

### DIFF
--- a/api/models/build-task.js
+++ b/api/models/build-task.js
@@ -1,0 +1,56 @@
+const { buildEnum } = require('../utils');
+
+const Statuses = buildEnum([
+  'created',
+  'queued',
+  'tasked',
+  'error',
+  'processing',
+  'skipped', // remove?
+  'success',
+]);
+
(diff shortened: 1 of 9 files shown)


api/models/build-task.js
```diff
@@ -0,0 +1,56 @@
...
+        },
+      },
+    }, {
+      tableName: 'build_task',
+      paranoid: true,
+      indexes: [
+        {
+          name: 'build_task_build_id_type_index',
+          unique: true,
+          fields: ['buildId'],
```
apburnes, line 47: You need to add `buildTaskTypeId` based on the migration SQL.

apburnes (changes requested):
You missed a field in the sequelize model of the build task index.

api/models/build-task.js
drewbo: Excellent catch. My kingdom for not having to write these schemas twice

apburnes (approved):
🏗️

PR #4247, opened 2023-09-07
Add build task router and controller

### LINKED ISSUE #4169: Create options for starting, managing, and collecting information from the build tasks
## Acceptance Criteria
- [x] A draft API for managing the checks
- [x] A proposed method for the checks to store results

drewbo:
- add build task router and controller
- close #4169

This includes some deviations from the initial proposal:
- the read routes don't include `/site/:site_id/` because the build ID can be used to authorize the user to read without it.
- the update route can now update any field (not just status) but includes a token check so that random users can't use it

This PR doesn't handle the full title of #4169 but it hits the acceptance criteria, happy to add more if needed

## security considerations
None

### DIFF
--- a/api/controllers/build-task.js
+++ b/api/controllers/build-task.js
@@ -1,0 +1,63 @@
+const { wrapHandlers } = require('../utils');
+const { Build, BuildTask } = require('../models');
+
+module.exports = wrapHandlers({
+  find: async (req, res) => {
+    const { params, user } = req;
+    const { build_id: buildId, build_task_id: buildTaskId } = params;
+
+    // the build check essentially serves as an authorizer
+    const build = await Build.forSiteUser(user).findByPk(buildId);
+
+    if (!build) {
+      return res.notFound();
+    }
+
+    const task = await BuildTask.findOne({
+      where: { buildId, id: buildTaskId },
+      attributes: { exclude: ['token', 'deletedAt'] },
+    });
+
+    if (!task) {
+      return res.notFound();
+    }
+
+    return res.json(task);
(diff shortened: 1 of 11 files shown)


drewbo:
Removing review for now, going to add some tests

drewbo:
Noting here so I don't forget: the initial proposal used `tasks` rather than the singular `task`. Other portions of the API generally use the singular except for build `logs` which are fairly similar to this. Open to changing to singular if needed

api/controllers/build-task.js
```diff
@@ -0,0 +1,63 @@
...
+  update: async (req, res) => {
+    const { params, body } = req;
+    const { build_task_id: buildTaskId, token } = params;
+
+    const task = await BuildTask.findByPk(buildTaskId);
+
+    if (!task) {
+      return res.notFound();
+    }
+    if (task.token !== token) {
```
apburnes, line 55: Should we be expiring tokens? After status becomes `success`, `skipped`, or `error`, the update endpoint will throw a forbidden response.

apburnes, line 55: Should we be expiring tokens? After status becomes `success`, `skipped`, or `error`, the update endpoint will throw a forbidden response.

public/swagger/index.yml
apburnes, line 1: Should we add the `/tasks/:build_task_id/:token` `PUT` endpoint?

apburnes (changes requested):
Just a couple of questions on the build task status update

apburnes (approved):

PR #4248, opened 2023-09-11
Feat skeleton build task

drewbo:
Extreme WIP (but functional) PR which implements the following items from the acceptance criteria

 - [x] Create a build task queue
 - [x] Run build task when a site build starts **and when site build completes, depending on user/task config**
 - [x] Run task from worker thread
 - [x] Update tasks queue with new status
 - [x]  On task complete, save task artifact and complete tasks queue with finish status.

I've also added a "hidden admin page" at `/tasks` which shows some build task information. The build task does run code, but currently it's fully encapsulated in the worker, rather than running a CF task. I have a number of problems with the current architecture and would love advice on proceeding:
- As we've identified previously, the project has an odd mix of object-oriented vs functional programming throughout. The place it is most obvious here is the faux-instance methods `enqueue` (previously existing) and `startBuildTask` (added).
- As written, the tasks are all implemented as worker code only, rather than CF tasks. They also have implied customs with no way to enforce them (function arguments and artifacts location)
- Because the worker tasks have access to the DB directly, they don't necessary need to use the status endpoint but I've included it as an example. Do we want to have some tasks available as worker code only or should they all be CF tasks?

## security considerations
New non-user enforced endpoint is added for status reporting (uses secret token to validate responses)


api/controllers/build.js
```diff
@@ -134,6 +135,17 @@ module.exports = wrapHandlers({
       EventCreator.error(Event.labels.BUILD_STATUS, err, { buildId: build.id });
     }
 
+    // we start certain build tasks on build completion
+    if (body.status === 'success') {
+      try {
+        const tasks = await SiteBuildTask
```
apburnes, line 141: What's your thoughts on model instance hooks? We could add the `startBuildTasks` on the `afterSave` or `afterUpdate` hook and then have a higher order function that starts the proper build tasks based on the updated `status`. https://sequelize.org/docs/v6/other-topics/hooks/#instance-hooks

Does it simplify starting the build tasks? Is it too coupled to the ORM?

api/models/site-build-task.js
```diff
@@ -8,6 +11,59 @@ const associate = ({ BuildTaskType, Site, SiteBuildTask }) => {
...
+      { model: BuildTaskType, required: true },
+      { model: Build, required: true, include: [{ model: Site, required: true }] },
+    ],
+  });
+
+  BuildTaskQueue.sendTaskMessage(createdBuildTask);
+}
+
+// TODO: should this be converted to sequelize scope?
+async function findTasksToStart({ build, startsWhen }) {
```
apburnes, line 41: Should we nest or call this  method in `startBuildTask` to simplify how we start tasks from other parts of the app?

api/queues/BuildTasksQueue.js
```diff
@@ -0,0 +1,24 @@
...
+  constructor(connection) {
+    super(BuildTasksQueueName, {
+      connection,
+      defaultJobOptions: {
+        attempts: 3,
+        backoff: {
+          type: 'exponential',
+          delay: 3000,
+        },
+        delay: 2 * 60 * 1000, // 2 minutes
```
apburnes, line 15: What are the advantages of adding a default `delay` on job kickoff?

apburnes (commented):
Mostly looking for your thoughts on different options for kicking off the build tasks. Once we test out running jobs on a worker, cf task, or third party service, we'll have a better idea of how we want the job processor interface to function. We may want to lean more on bullmq's API for managing the job lifecycle but this is a great start. Awesome work.

api/controllers/build.js
drewbo: I think this one would be tricky because builds are kicked off via `.enqueue` which always follows `Build.create` in our codebase but the two aren't guaranteed to be linked. Until we re-investigate some of the status/queue logic, I'm a little worried about skipping/double-starting build tasks if we link it to the instance hooks (even though it is probably a "cleaner" method overall

api/queues/BuildTasksQueue.js
drewbo: Oversight, I think I copied this config over from another queue but I'll edit

api/models/build.js
```diff
@@ -189,6 +190,15 @@ async function enqueue() {
     });
   }
 
+  // after the build is queued, check for and initiate relevant build tasks
+  try {
+    const tasks = await SiteBuildTask
+      .findTasksToStart({ build, startsWhen: 'build' });
+    tasks.forEach(sbt => sbt.startBuildTask(build.id));
```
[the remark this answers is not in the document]
drewbo: I think `forEach` is fine as long as we aren't doing anything afterwards ("fire and forget") but I'm going to rewrite this portion anyway to incorporate the feedback below

drewbo:
@apburnes updated in response to your feedback:
- I added the `Promise.allSettled` to the tasks but right now there isn't a return value, seems okay just noting
- it's currently not started the "build complete" build task: still looking into that

api/models/build.js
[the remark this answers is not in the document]
drewbo: Ended up implementing this during the rewrite

api/controllers/build.js
drewbo: I tried this and it seemed to work fine; it was nice to remove the logic from the build controller.

api/models/site-build-task.js
```diff
@@ -8,6 +11,62 @@ const associate = ({ BuildTaskType, Site, SiteBuildTask }) => {
...
+      model: BuildTaskType,
+      where: { metadata: { startsWhen } },
+      required: true,
+    },
+  });
+}
+
+async function startBuildTasks({ build, startsWhen }) {
+  const tasks = await this.findTasksToStart({ build, startsWhen });
+  await Promise.allSettled(tasks.map(sbt => sbt.startBuildTask(build.id)));
```
apburnes, line 67: For tasks that reject, we may want to think about how we handle them in the future. If they were to fail it would be a database or queue connection problem which means its probably a larger event happening.

apburnes:
@drewbo looks good so far.

drewbo:
This is ready to go again, in conjunction with https://github.com/cloud-gov/owasp-zap-task/pull/1. It contains API tests but nothing for the CF API call to the tasks endpoint; happy to add that before final review.

Also because this is a large feature, I'm available to demo or add more docs as needed.

apburnes (approved):
👏

Issue #4258, opened 2023-10-02
Make org/user data available via reports interface


svenaas:
Make org, user, and user role information available via the Pages Admin reports interface to facilitate the synchronization process that will be kicked off by #4246 

## Notes

- Make org, user, and user role information available in CSVs suitable for import to ZenDesk

## Acceptance Criteria

- [x] Org information available via CSV
- [x] User information available via CSV with org/role associations

PR #4262, opened 2023-10-03
Make org/user data available via reports interface

svenaas:
Work on part of #4258, adding new reports with orgs and user/org/role information available as CSVs for import into ZenDesk. These reports are presented in the Admin UI as well (slightly differently), with the CSV downloads available from this interface, or potentially via direct calls to the API.

- Move the existing `/organizations-report` to `/reports/published-sites` 
- Add a new organizations report (that includes all orgs, not just those with published sites) at `/reports/organizations` 
- Add a new users report that includes associated organizations and roles at `/reports/users`

## security considerations
None. The data presented is only available to authenticated admin/support users, and only represents data already available to those users.


admin-client/src/Router.svelte
```diff
@@ -66,7 +66,9 @@
   page('/organizations/new', queryString, render(Pages.Organization.New));
   page('/organizations/:id', queryString, render(Pages.Organization.Show));
   page('/organizations/:id/edit', queryString, render(Pages.Organization.Edit));
-  page('/organizations-report', queryString, render(Pages.Organization.Report));
+  page('/reports/organizations', queryString, render(Pages.Organization.OrgsReport));
```
sknep, line 69: This is just a question -- when we remove routes that have previously been public (possibly used/bookmarked by customers), have we ever (or could we) define redirects to the next closest route?

test/api/unit/models/organization.test.js
```diff
@@ -205,6 +205,20 @@ describe('Organization model', () => {
     });
   });
 
+  describe('.byName()', () => {
+    it('returns organizations ordered by name', async () => {
+      const orgB = Organization.create({ name: 'Org B' });
+      const orgA = Organization.create({ name: 'Org A' });
+      const orgC = Organization.create({ name: 'Org C' });
```
sknep, line 212: Would you consider testing the sorting with strings that also start with numbers? Thinking about where "10x" and 18F" go

admin-client/src/Router.svelte
svenaas: Not sure as a general question. But in _this_ case this is admin-only, only usable by Pages operators, only called by the admin interface, and only in existence since I [added it](https://github.com/cloud-gov/pages-core/pull/4145) in May.

test/api/unit/models/organization.test.js
svenaas: The sorting happens in the underlying database via ORM in the Organization model — I expect it to be ASCIIbetical: 
```js
  Organization.addScope('byName', {
    order: [
      ['name', 'ASC'],
    ],
  });
```

It's the job of this unit test to test the organization model and in this specific case what it's doing is verifying that the sorting by org name is being _applied_ to these results, not verifying the details of the sorting algorithm. I would absolutely consider testing other aspects of sorting if we have a good reason to do that — and you may be seeing one that I haven't! — but my gut tells me that wouldn't come up unless we found ourselves writing code to customize the sort to address a particular need.

Ephraim-G (dismissed):

drewbo (approved):
Redeployed and tested on dev. Added another org to see multiple org users. Everything looks good. The only minor (optional) thing is I'd name the columns "Github email" and "UAA email" for consistency with the users view. Good to go once rebased

svenaas:
> The only minor (optional) thing is I'd name the columns "Github email" and "UAA email" for consistency with the users view. 

I agree with that. But the CSV export just has an "Email" (UAA) column for ZenDesk convenience as of this writing. Does that seem OK to you, @drewbo?

svenaas:
@drewbo, I changed the email column names and order in the users report to match the users administration page — that needs a re-review. But you can't just read that change because I also rebased this down to three discrete commits (you'd probably go with one, and I can do that if you prefer).

svenaas:
Hold off on that, please; test is behaving differently in Concourse because I wasn't careful about ordering. Fixing that now.

drewbo (changes requested):
Can you change the new report commits to `feat(admin):` type? I'm not positive how the changelog will handle that but it better reflects those changes

svenaas:
Yes. Done.

drewbo (changes requested):
One last code piece I should have checked earlier: can you abstract the `downloadCSV` function to accept the csv as input, and put this in a util file? (otherwise it's duplicated three times)

svenaas:
Yeah, sure (sigh) that's a good idea. I should have thought of that as I based two new Svelte files on the first one.

svenaas:
OK, @drewbo, it's in `helpers` now; please take a particular look at the way I'm defining the helper function and passing parameters in. Do you think this refactoring is good, or do you think it would be better to go a step further and make a `CSVDownloadButton` Svelte component? Because I could try that too.

drewbo (approved):

PR #4267, opened 2023-10-10
Add self authorization fields to organizations and published sites reports

### LINKED ISSUE #4266: Add self authorization fields to Published Sites and Organizations reports
Include self authorization data in Published Sites and Organizations reports

## Notes

- The Organization model's self authorization data should be included in the Organizations and Published Sites reports (both in-browser and CSV versions)

## Acceptance Criteria

- [ ] Is Self Authorized column in Organizations report
- [ ] Is Self Authorized column in Organizations report CSV
- [ ] Is Self Authorized column in Published Sites report
- [ ] Is Self Authorized column in Published Sites report CSV

svenaas:
Fixes #4266

- Adds "Self Authorized" column to Organizations and Published Sites reports, containing data from the Organization model.
- Adds self authorization values to one org in the test data.

## security considerations
None. This change makes data already available to admins available in two reports available to admins.

### DIFF
--- a/admin-client/src/pages/organization/OrgsReport.svelte
+++ b/admin-client/src/pages/organization/OrgsReport.svelte
@@ -12,6 +12,7 @@
     <tr slot="header">
       <th scope="col">Organization</th>
       <th scope="col">Agency</th>
+      <th scope="col">Self Authorized</th>
     </tr>
     <tr slot="item" let:item={org}>
       <td>
@@ -19,6 +20,9 @@
       </td>
       <td>
         {org.agency}
+      </td>
+      <td>
+        {org.isSelfAuthorized ? 'Yes' : 'No'}
       </td>
     </tr>
   </DataTable>
--- a/admin-client/src/pages/organization/PublishedSitesReport.svelte
+++ b/admin-client/src/pages/organization/PublishedSitesReport.svelte
@@ -22,6 +22,7 @@
     <tr slot="header">
       <th scope="col">Organization</th>
       <th scope="col">Agency</th>
+      <th scope="col">Self Authorized</th>
       <th scope="col">Site</th>
       <th scope="col">Domains</th>
       <th scope="col">Engine</th>
@@ -35,6 +36,11 @@
       <td>
         {#if domain.Site && domain.Site.Organization}
           {domain.Site.Organization.agency}
+        {/if}
+      </td>
+      <td>
+        {#if domain.Site && domain.Site.Organization}
+          {domain.Site.Organization.isSelfAuthorized ? 'Yes' : 'No'}
         {/if}
       </td>
       <td>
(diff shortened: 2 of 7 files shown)


drewbo (changes requested):
`isSelfAuthorized` is a virtual field, so I think you can skip it in the `.create` method. Good to approve otherwise

admin-client/src/pages/organization/OrgsReport.svelte
```diff
@@ -20,6 +22,13 @@
       <td>
         {org.agency}
       </td>
+      <td>
+        {#if org.isSelfAuthorized }
+          { formatDistanceStrict(new Date(org.selfAuthorizedAt), new Date(), { addSuffix: true, roundingMethod: 'floor' }) }
```
apburnes, line 27: Do we care about a date just for self authorized? Should we just use the `isSelfAuthorized` boolean?

apburnes (commented):
Just a quick question on the self auth data type. Other than that it looks good.

svenaas:
> `isSelfAuthorized` is a virtual field, so I think you can skip it in the `.create` method. Good to approve otherwise

That's what I get for inspecting a model instance instead of the database or the code. Fixing it.

admin-client/src/pages/organization/OrgsReport.svelte
svenaas: Do we care about the date in general? Should this field just be stored as a boolean?

The reason I'm showing it is that I saw that it's shown in the admin orgs index page, so I thought I ought to be consistent and show it here. In the onscreen report I show boolean and date in one column, similar to how it's done in the orgs index page. 

In the CSV export if I were to only include one column in the CSV I would have thought it should be the date one, but right now I've got them in separate columns on the theory it might make downstream sort/filter semantics a small bit simpler.

apburnes: Reviewing it more, I think we only care if an org is or is not self authorized. Using `No` in conjunction with a date since for `Self Authorized` is mixing the types and the current Self Authorized field in the orgs table shows similar ambiguity with the timestamp and an `N/A`. If you want to change the org table's Self Authorized field to boolean too that would be great.

It should simplify the component and a future boolean filter on self authorization is more pertinent in targeting groups.

svenaas: All right, I will only show and export the boolean in these reports. I think changing the underlying field should happen on a separate issue and PR, though. I'll spin that up next.

apburnes (approved):
👍

svenaas:
@drewbo, this is waiting for your approval too since you requested changes. It's not urgent.

drewbo (approved):

PR #4269, opened 2023-10-11
Change self authorized field from timestamp to boolean

### LINKED ISSUE #4268: Change self authorized field from timestamp to boolean.
Self authorization is recorded for Organizations as `selfAuthorizedAt`, and frequently accessed through the computed Boolean field `isSelfAuthorized`. The date and time of self authorization does not need to be tracked here, and could be needlessly complex to source and audit, so it should be simplified, leaving an underlying Boolean with the same name as the computed field.

## Acceptance Criteria

- [ ] The `selfAuthorizedAt` field and column on Organization are replaced by an `isSelfAuthorized` Boolean

svenaas:
Fixes #4268 

Modifies the organization model and table to store only a boolean for `isSelfAuthorized` instead of a timestamp.

- Replaces the `isSelfAuthorized` computed field on Organization with a real `isSelfAuthorized` field (boolean, non null, default false)
- Sets this field to true only for organizations where `selfAuthorizedAt` is not null.
- Removes the `selfAuthorizedAt` field from Organization

## Migration consideration:
The migration in this PR replaces timestamp data with boolean data. This is destructive, but allowable provided we do not need or want this timestamp data. The `down` callback on the migration does retain the key true/false data by setting the timestamp to the current time in place of `true`, which is consistent with the app behavior before this migration.

## security considerations
None. This is a change in the representation of data already available to admins.

### DIFF
--- a/migrations/20231011151001-change-org-self-auth-to-boolean.js
+++ b/migrations/20231011151001-change-org-self-auth-to-boolean.js
@@ -1,0 +1,13 @@
+const TABLE = 'organization';
+
+exports.up = async db => {
+  await db.addColumn(TABLE, 'isSelfAuthorized', { type: 'boolean', notNull: true, defaultValue: false });
+  await db.runSql('update "organization" set "isSelfAuthorized" = TRUE where "selfAuthorizedAt" is not null');
+  await db.removeColumn(TABLE, 'selfAuthorizedAt');
+};
+
+exports.down = async db => {
+  await db.addColumn(TABLE, 'selfAuthorizedAt', { type: 'timestamp', allowNull: true });
+  await db.runSql('update "organization" set "selfAuthorizedAt" = CURRENT_DATE where "isSelfAuthorized" IS TRUE');
+  await db.removeColumn(TABLE, 'isSelfAuthorized');
+};
--- a/admin-client/src/pages/organization/Index.svelte
+++ b/admin-client/src/pages/organization/Index.svelte
@@ -32,7 +32,7 @@
           regular
         {/if}
       </td>
-      <td>{formatDateTime(org.selfAuthorizedAt)}</td>
+      <td>{org.isSelfAuthorized ? 'Yes' : 'No'}</td>
       <td>
         {#if org.isActive}
           active
--- a/api/models/organization.js
+++ b/api/models/organization.js
@@ -119,20 +119,10 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.STRING,
       allowNull: true,
     },
-    selfAuthorizedAt: {
-      type: DataTypes.DATE,
-      allowNull: true,
-    },
     isSelfAuthorized: {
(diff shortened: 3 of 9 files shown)


migrations/20231011151001-change-org-self-auth-to-boolean.js
```diff
@@ -0,0 +1,13 @@
+const TABLE = 'organization';
+
+exports.up = db => Promise.all([
```
apburnes, line 3: The `Promise.all` method runs the promises in the array concurrently and not in succession. Since some of the SQL calls depend on other calls, you'll want to either use the `async/await` syntax to guarantee the calls are called in order.

apburnes (changes requested):
Looks good. We just need to update the way we run the migration and I added a todo comment.

migrations/20231011151001-change-org-self-auth-to-boolean.js
apburnes: Ie. https://github.com/cloud-gov/pages-core/blob/4268-change-self-auth-to-boolean/migrations/20210224210834-init-organizations.js#L78-L91

svenaas: **Great** catch, @apburnes. Order is certainly important here. Thank you!

svenaas:
@apburnes I've changed the migration to control the order of the db calls.

migrations/20231011151001-change-org-self-auth-to-boolean.js
```diff
@@ -0,0 +1,13 @@
...
+
+exports.up = async db => {
+  await db.addColumn(TABLE, 'isSelfAuthorized', { type: 'boolean', notNull: true, defaultValue: false });
+  await db.runSql('update "organization" set "isSelfAuthorized" = TRUE where "selfAuthorizedAt" is not null');
+  await db.removeColumn(TABLE, 'selfAuthorizedAt');
+};
+
+exports.down = async db => {
+  await db.addColumn(TABLE, 'selfAuthorizedAt', { type: 'timestamp', allowNull: true });
+  await db.runSql('update "organization" set "selfAuthorizedAt" = CURRENT_DATE where "isSelfAuthorizedAt" IS TRUE');
```
apburnes, line 11: Need to update the column name to `isSelfAuthorized` not `isSelfAuthorizedAt`

apburnes (changes requested):
Column name mismatch in migration.

migrations/20231011151001-change-org-self-auth-to-boolean.js
svenaas: Now that's just embarrassing.

svenaas:
OK. Fixed a race condition in one of the new tests as well. After that there was an apparent solar flare failure over in `SandboxHelperTest` which I cannot explain, but which was transient.

apburnes (approved):
Good to go. Thanks.

PR #4275, opened 2023-10-19
fix: Fixes domain checkProvisionStatus with new CF api response

apburnes:
- Fixes the `checkProvisionStatus` function in the domain provisioning job to look at the new CF API V3 response structure

## security considerations
None


drewbo (approved):
Are there any other places using the `entity` structure we need to change? (it's a tiny tragedy that there isn't an official CF API mock library)

apburnes:
No other places. This was a test that switched from using mocked requests to method stubs.

PR #4281, opened 2023-10-26
WIP Upgrade to AWS SDK v3

### LINKED ISSUE #3989: Upgrade to aws-sdk v3
https://www.npmjs.com/package/aws-sdk
> We are formalizing our plans to enter AWS SDK for JavaScript v2 into maintenance mode in 2023.

## Acceptance Criteria

- [ ] AWS SDK dependency updated from v2 (2.879) to v3. 
- [ ] Core AWS SDK dependencies function as expected:
  - [ ] Site creation with initial build
  - [ ] Site deletion including the removal of the hosted preview site
  - [ ] Site uploaded files page working normally (note: there's currently a bug on this page in production)

svenaas:
Fixes #3989 

- Update AWS SDK from v2 to v3 (2.879.0 to 3.387.0)
- Switch to `aws-sdk-client-mock` for mocking AWS SDK calls

## security considerations
None. This is an update to an important dependency to keep us on a supported version.


api/services/S3Helper.js
```diff
@@ -43,111 +21,113 @@ class S3Client {
 
   waitForCredentials() {
     const { bucket, client } = this;
-    return retry(() => client.headBucket({ Bucket: bucket }).promise(), { waitTime: 2000 });
+    const command = new HeadBucketCommand({ Bucket: bucket });
+    return retry(async () => client.send(command), { waitTime: 2000 });
   }
 
-  listCommonPrefixes(prefix) {
+  async listCommonPrefixes(prefix) {
```
drewbo, line 28: the three list/paginated functions could go back to using a helper function like before because they all have the same structure

drewbo (changes requested):
Only one small note for DRY sake; otherwise I saw (and approved of) most of this while pairing

api/services/S3Helper.js
svenaas: OK, this is pushed.

svenaas:
Next week when Drew and I are both here we should test this on dev — and subsequently on staging — to give the actual connection to S3 a shakedown.

api/services/S3SiteRemover.js
```diff
@@ -85,24 +49,23 @@ const removeSite = async (site) => {
...
+    for await (const page of paginator) {
+      // Delete all of the objects in the current page
+      const commandInput = {
+        Bucket: s3Client.bucket,
+        Delete: {
+          Objects: page.Contents.map(object => ({ Key: object.Key })),
+        },
+      };
+      const command = new DeleteObjectsCommand(commandInput);
+      await s3Client.client.send(command);
```
apburnes, line 67: Is there a reason we import the aws-client dependency here? If not, can we move the delete objects functionality into the S3Helper as a method?

apburnes (commented):
Awesome. Just a question about the site remover functionality.

api/services/S3SiteRemover.js
svenaas: Honestly I think the reason is simply that I just updated that bit to v3 and simplified without reorganizing it — the deletion was implemented in S3SiteRemover before and I didn't change that; I just tried to simplify it. 

Moving the delete objects functionality into S3Helper wouldn't be enough to eliminate the aws-client dependency in S3SiteRemover, though — it would leave behind the iteration with `paginateListObjectsV2`. We could move that into S3Helper too, I suppose, which would leave behind the credentials logic, error handling, and of course `removeInfrastructure`, which needs to live somewhere but is as much of a CF thing as it is an S3 thing. What do you think would be best?

apburnes: This is a good opportunity to clean up cruft. Can you create a `deleteAllBucketObjects` method in the S3Helper to wrap the pagination, iteration and deletion logic here into that method?

svenaas: I can and now I have. One test is largely duplicated but that's part of why I'm sure I didn't break something when refactoring, so I left it.

apburnes (dismissed):
S3 bucket deletion is not working in dev

svenaas:
@apburnes @drewbo Creation and deletion and the published file page should all be working now. Please come back and review again.

drewbo (approved):
Confirmed all of site creation/deletion, published files, and old build log fetching. There is one _small_ error that I think we can push to a separate issue. If you click published files immediately after the site is created, it will still error but I think that can be handled separately in the front end

PR #4299, opened 2023-11-15
chore: update dependency minor patch bumps for pages core and admin client

Ephraim-G:
- update dependencies across pages core app and admin/client 

## security considerations
N/A


Ephraim-G:
@drewboif it's only node causing the dependency failure then this should be good to go w/o regenerating the .lock file?

svenaas:
Is `concourse/audit-dependencies` an expected failure even on this branch?

drewbo (approved):

Ephraim-G:
@svenaas yup!

Issue #4313, opened 2023-11-27
Implement users report limited to users with a UAA Identity


apburnes:
The current users report contains all non-deleted users in the database. This is an important list, but for Zendesk integration purposes what is relevant is the set of users who can log in — those users with UAA Identities. 

## Acceptance Criteria

- [x] Users report limited to active UAA Identity users implemented
- [x] Update the Pages/Zendesk monthly sync issue template to use this new report instead of the old All Users one.

PR #4317, opened 2023-12-04
Implement users report limited to users with a UAA Identity

svenaas:
Part of #4313

- Introduces a new Active Users report which only includes users who have a UAA Identity
- Renames the existing Users report "All Users". 

## security considerations
None. This adds a report visible by admins to conveniently view information admins already have access to.


drewbo (approved):
Looks good. Do we need to retain the "All users" report for any reason?

PR #4320, opened 2023-12-05
Add global Sequelize hook to make count behavior consistent with find

svenaas:
In ORM queries incorporating association joins, Sequelize's `count` sometimes reports a larger number than the number of rows that Sequelize returns by the same query. This causes inconsistencies between the actual number of returned results and the number of total results reported, a problem which specifically comes up with some of our reports. 

This PR adds a global `beforeCount` hook that changes Sequelize's default behavior to make `count` consistent. It also introduces a test which fails in the absence of this hook. 

In addition to the test case, the misbehavior can be seen in the admin interface at `/reports/users`:

**Without** the hook a current local development instance will show 14 rows of user data:
Screenshot 2023-12-05 at 2 30 05 PM

but an inconsistent "total results: 17": 
image

**With** the hook the total is consistent:
Screenshot 2023-12-05 at 2 30 10 PM

## Background

In working on #4313 it came to my attention that some of our PaginatedQueryPages were displaying inconsistent information. Specifically, the "total results: _n_" in PaginationBanner differed from the number of rows shown. 

@sknep identified this number as coming from the Sequelize's `count` via the `paginate` function in `api/utils/index.js`, and found other people [reporting](https://stackoverflow.com/questions/64354987/findandcountall-count-gets-a-bigger-number-than-the-actual-returned-data) related issues. 

This led me to https://github.com/sequelize/sequelize/issues/9481 and from there to https://github.com/sequelize/sequelize/issues/10557 where various workarounds are discussed. I experimented with simply adding `distinct: true`, but that caused other things to fail. What _did_ work for me was James M. Greene's [beforeCount hook](https://github.com/sequelize/sequelize/issues/10557#issuecomment-481399247), which I opted to place in `api/models/index.js`.

- Adds global Sequelize hook
- Adds unit test for this hook in `test/api/unit/models/sequelize-hook.test.js` — is there a better place for this to live?

## security considerations
None. This change is a bug fix that makes a reported count consistent with the size of the result set it refers to.

### DIFF
--- a/api/models/index.js
+++ b/api/models/index.js
@@ -15,6 +15,21 @@ const sequelize = new Sequelize(database, username, password, {
   retry,
 });
 
+// Add a permanent global hook to prevent unknowingly hitting this Sequelize bug:
+//   https://github.com/sequelize/sequelize/issues/10557#issuecomment-481399247
+sequelize.addHook('beforeCount', function (options) {
+  /* eslint-disable no-underscore-dangle, no-param-reassign */
+  if (this._scope.include && this._scope.include.length > 0) {
+    options.distinct = true;
+    options.col = this._scope.col || options.col || `"${this.options.name.singular}".id`;
+  }
+
+  if (options.include && options.include.length > 0) {
+    options.include = null;
+  }
+  /* eslint-enable no-underscore-dangle, no-param-reassign */
+});
+
 require('./build')(sequelize, DataTypes);
 require('./build-log')(sequelize, DataTypes);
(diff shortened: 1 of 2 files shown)


sknep (dismissed):

drewbo (approved):

svenaas:
Sorry to dismiss your review, @drewbo — I'd forgotten to link directly to the comment where the hook workaround came from.

drewbo (approved):

PR #4322, opened 2023-12-08
Build Task UI

### LINKED ISSUE #4171: Implement build checks UI

drewbo:
- Updates the build history page with new styling, improved contextual information, and links to the new build tasks view
- Adds a new page for viewing all build tasks for a single build
- Other DB/API changes to support the above
- Close #4171 

## security considerations
This solidifies the release of the initial UI associated with Build Tasks. It is gated to not appear in production until an SCR is completed documenting its architecture and any potential security risks.


api/models/build-task-type.js
```diff
@@ -55,6 +55,9 @@ module.exports = (sequelize, DataTypes) => {
           isIn: [StartsWhens.values],
         },
       },
+      url: {
```
apburnes, line 58: What does the URL do?

frontend/components/GitHubLink.jsx
```diff
@@ -15,22 +15,43 @@ const GitHubLink = ({
...
+  }
+  function chooseIcon(iconStr) {
+    switch (iconStr) {
+      case 'branch':
+        return (
+          <IconBranch />
+        );
+      case 'commit':
+      case 'sha':
+        // return (
```
svenaas, line 31: Is there a reason this is being committed?

scripts/create-dev-data.js
```diff
@@ -408,61 +408,237 @@ async function createData() {
...
+    name: 'type',
+    status: 'success',
+    artifact: 'filename-1234.html',
+    message: 'Scan successfully completed. See artifact for details.',
+    count: 42,
+  });
+
+  const taskType2 = await BuildTaskType.create({
+    name: 'WCAG 2.2 AA Accessibility Scan',
+    description: 'This scan detects accessibility issues and provides suggestions for remediation by inspecting focusable elements, HTML tags and attributes, images, data tables, color contrast, document structure, link and button usability, and visually hidden content against the WC3’s WCAG level 2.2 A and AA.',
```
svenaas, line 574:
```suggestion
    description: 'This scan detects accessibility issues and provides suggestions for remediation by inspecting focusable elements, HTML tags and attributes, images, data tables, color contrast, document structure, link and button usability, and visually hidden content against the WC3’s WCAG 2.2 Level A and AA requirements.',
```

drewbo: Actually @sknep do you want this to be the real description also? I can add to the migrations

ci/pipeline-dev.yml
```diff
@@ -35,6 +35,7 @@ jobs:
     plan:
       - get: src
         resource: pr-((deploy-env))
+        params: {integration_tool: checkout}
```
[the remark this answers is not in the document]
drewbo: yes but it looks like all our other single params are passed as inline dictionaries so I went with this for consistency (happy to change them all though)

frontend/components/GitHubLink.jsx
```diff
@@ -15,22 +15,43 @@ const GitHubLink = ({
...
     <a
-      className="repo-link"
+      className={`${icon}-link`}
       href={href}
       title={title}
       target="_blank"
       rel="noopener noreferrer"
     >
+
+      { (chooseIcon(icon))}
```
apburnes, line 53: This is a note for future consideration. We have no conventions or standards in place for how we structure components. Could this be a general icon component that handles the select logic within the component? ie the above could be`<Icon name={icon} />`. If we come to some convention, we could apply it to our common components with just a few props to differentiate them.

frontend/components/site/CommitSummary.jsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+CommitSummary.defaultProps = {
+  build: null,
+};
+
+const mapStateToProps = ({ build }) => ({ build });
+
+export { CommitSummary };
+export default connect(mapStateToProps)(CommitSummary);
```
apburnes, line 79: For future consideration. We need to decide/document how we connect global state to components. Should we keep everything in `frontend/components` as pure components and pass all their props down, should we restructure them into their own directories (ie `frontend/containers`), only connect to the state at the page level, or leave it as is?

frontend/components/GitHubLink.jsx
sknep: I was kind of following the convention here:
https://github.com/cloud-gov/pages-core/blob/348e91b6483f191904f2e78082165530f1f8b924/frontend/components/site/siteBuilds.jsx#L24

I do like the idea of cleaner component structure though.

frontend/components/site/CommitSummary.jsx
sknep: Yeah, I really wondered what the 'right' path for this was, too. It seemed like we connected state and props for better unit testing, but @drewbo has been working on a better way to mount the components for testing, iirc...

drewbo: I think the Redux-era convention was global state, one reducer for each "smart component" which roughly existed at a page-level, and then pure components below that. It seems like React Hooks are moving folks away from even persisting a global state but don't necessarily specify at which level state should be maintained. The current [React docs about state](https://react.dev/learn/managing-state) aren't super prescriptive about this and at times seem to reproduce a lot of Redux ideas, so I'd favor a longer conversation before we look into a bigger refactor here.

svenaas (commented):
The build task type descriptions are duplicated in a couple of places; is there a good way to avoid that? 

Otherwise this looks good to me but I'm going to hold off on clicking Approve until @apburnes has a look since he was part of a few threads here that aren't yet resolved.

svenaas:
> The build task type descriptions are duplicated in a couple of places; is there a good way to avoid that?

I went back and looked again. Duplication is between the migration and `scripts/create-dev-data.js` and there may not be an elegant way to avoid that.

ci/pipeline-dev.yml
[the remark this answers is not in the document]
apburnes: All good. It just stuck out in the diff

frontend/components/GitHubLink.jsx
apburnes: Started tracking the creation of a future standard in https://github.com/cloud-gov/pages-core/issues/4332

frontend/components/site/CommitSummary.jsx
apburnes: Started tracking the creation of a future standard in https://github.com/cloud-gov/pages-core/issues/4332

apburnes (approved):
🏗️ 🥳

svenaas (approved):
Right then. Here's another for good measure.

Issue #4333, opened 2023-12-15
Improve brittle admin user API test


svenaas:
Improve the Admin Users API with the hope of preventing an inconsistent failure.

## Notes

The Admin Users API test is inconsistently and infrequently failing in Concourse with 

```
  1) Admin - Users API
       GET /admin/reports/active-users.csv
         returns all users with UAA identities:
     AssertionError: expected [ …(2) ] to include '70,"user1@example.com","org-16|org-17…'
      at Context.<anonymous> (test/api/admin/requests/user.test.js:204:23)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I have not been able to reproduce locally, but hope to prevent this failure through improvement to the test.

## Acceptance Criteria

- [x] Test improved
- [ ] We experience a decent interval without this test failing unexpectedly

svenaas:
I'm uncertain how long to wait before we consider a decent interval to have elapsed, check that second box, and mark this done and closed.

svenaas:
Well, [_rats_](https://ci.fr.cloud.gov/teams/pages/pipelines/core/jobs/test-and-deploy-api-pages/builds/69?vars.deploy-env=%22staging%22).

The [job passed on re-run with the same inputs](https://ci.fr.cloud.gov/teams/pages/pipelines/core/jobs/test-and-deploy-api-pages/builds/69.1?vars.deploy-env=%22staging%22), so this still isn't deterministic.

svenaas:
The test remains improved, but it also remains evidently flawed. I have not yet identified the problem.

apburnes:
Closing since issue has not occurred.

PR #4334, opened 2023-12-15
fix: Improve brittle admin user API test

svenaas:
Changes which I hope will sufficiently remediate #4333 

- Explicitly truncate UAAIdentity table after test
-  `await` `createUAAIdentity` calls

## security considerations
None. This is an effort to prevent an unexpected and inconsistent test failure.

### DIFF
--- a/test/api/admin/requests/user.test.js
+++ b/test/api/admin/requests/user.test.js
@@ -73,10 +73,10 @@ describe('Admin - Users API', () => {
 
     it('returns all users', async () => {
       const user1 = await factory.user();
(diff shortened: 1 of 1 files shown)


svenaas:
@drewbo This changes two things I spotted that could potentially allow failures due to timing conditions. I think these are both worthwhile, though I can't be certain this will resolve the matter. It had been a full week since the previous version was merged, with plenty of test runs since before you experienced that failure yesterday.

drewbo (approved):

PR #4340, opened 2023-12-19
fix: Check if buildShaLink args are null on CommitSummary component

apburnes:
- Check if `buildShaLink` function args are null on `CommitSummary` component

## security considerations
none


drewbo (commented):
How do we end up with builds without a sha link? Rebuild?

drewbo (approved):

svenaas (approved):

PR #4348, opened 2023-12-21
chore: Remove isActive from User #4338

### LINKED ISSUE #4338: Remove isActive from User
The isActive column in the User table and the corresponding field in the Model have no semantic purpose and should be removed.

Related to #4277

svenaas:
Fixes #4338 

- Removes isActive from user column and model

## security considerations
None. This removes an unused field from a model and corresponding database table.

### DIFF
--- a/admin-client/src/components/UserTable.svelte
+++ b/admin-client/src/components/UserTable.svelte
@@ -4,8 +4,6 @@
 
   export let users = [];
   export let borderless = false;
-
-  const stateColor = (isActive) => (isActive ? 'bg-mint' : 'bg-gray-30');
 </script>
 
 <DataTable data={users} {borderless}>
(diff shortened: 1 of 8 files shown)


svenaas:
@apburnes @drewbo I would appreciate your thoughts on [`api/services/Webhooks.js`](https://github.com/cloud-gov/pages-core/blob/126e6de795defa78e17224f5db993a6c68825a00/api/services/Webhooks.js). The [version in `main`](https://github.com/cloud-gov/pages-core/blob/main/api/services/Webhooks.js) made some use of `User.isActive`, which I removed, leaving behind logic dependent on a local `isActive` (`const isActive = action === 'member_added'`) and want to be sure this is the correct path.

drewbo:
Yeah, the local `isActive` refers to whether they are an active member of the federalist-local Github group. That functionality is deprecated so it should be fine

drewbo (approved):

PR #4389, opened 2024-02-21
ci: remove pipeline duplication, fix production formatting

cloud-gov-pages-operations:
- removes duplication around the database migration `on_failure` step
- fixes formatting on the production pipeline

## security considerations
None


svenaas (commented):
The code looks good. Do we have runs of these changes via manual fly?

apburnes (approved):

drewbo:
We will need to manually set the pipeline once this goes to production because the `set-pipeline` step is currently a no-op

PR #4392, opened 2024-02-22
add hardened playwright, registry-image, general-task

drewbo:
- Use `general-task`, the replacement for `harden-concourse-task`
- Use the hardened `registry-image` in place of the concourse type `registry-image` (identically named)
- Use hardened `playwright` image in place of `node` for end-to-end tests for the convenience/speed of not having to install `playwright` and it's dependencies each run (required separating out `create-test-user` into a separate step)

## security considerations
Replaces images with more updated and/or hardened options.

### DIFF
--- a/ci/pipeline-dev.yml
+++ b/ci/pipeline-dev.yml
@@ -327,8 +327,9 @@ jobs:
         - test-and-deploy-api-pages
         - test-and-deploy-admin-client-pages
         - deploy-queues-ui-pages
-      - get: node
+      - get: playwright
       - get: cf-image
+      - get: node
       - put: src
         resource: pr-((deploy-env))
         params:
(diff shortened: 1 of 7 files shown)


apburnes (dismissed):

ci/pipeline-dev.yml
```diff
@@ -343,13 +344,19 @@ jobs:
           <<: *env-cf
           APP_ENV: ((deploy-env))
           CF_APP_NAME: pages-((deploy-env))
+      - task: create-test-users
+        file: src/ci/partials/create-test-users.yml
+        image: node
+        params:
+          DOMAIN: ((((deploy-env))-pages-domain))
+          APP_ENV: ((deploy-env))
+          PRODUCT: pages
```
apburnes, line 353: For another issue but we should probably remove the `PRODUCT` env from the app and deployment.

ci/tasks/e2e-test.sh
```diff
@@ -2,12 +2,7 @@
...
-yarn install
-yarn playwright install-deps
-yarn playwright install
-
 export GIT_COMMIT=`git rev-parse HEAD`
 mkdir -p playwright-report/$APP_ENV/$GIT_COMMIT
 
-yarn create-test-users
-yarn playwright test
+npm exec playwright test
```
apburnes, line 8: Is yarn not installed in the playwright image? I'd be in favor of dumping yarn for npm across the platform.

drewbo: yarn is in the playwright image, but this works either way. I'm ambivalent about the switch (separate issue)

apburnes (approved):

PR #4399, opened 2024-02-28
Fix: Make invitation URL copyable (#4108)

### LINKED ISSUE #4108: Invitation URL is not copyable from alert popup
When I invite a new user to https://pages.cloud.gov/organizations/33, I get a popup alert that says I should send the invitation URL to the person for them to create an account with cloud.gov. 

Expected results: I can triple click or click-and-drag the URL to select it, and then copy
Actual results: I cannot select the URL text in any way to get it into my clipboard.

## Notes

- browser is Version 112.0.5615.49 (Official Build) (x86_64)
- brief exchange in slack today: https://gsa-tts.slack.com/archives/C04823NMA/p1681304705651299
- previous longer exchange when I was trying to add two people, and had trouble with both: https://gsa-tts.slack.com/archives/C04823NMA/p1677088701848519

## Acceptance Criteria

- [ ] User can select the URL text to copy
- [ ] Alert text is updated to let the user know that the invitee should have received an invitation including the URL, and this is just a fallback in case they don't get that email.

Thanks!

```[tasklist]
### Tasks
- [x] Mock up changes to experience (Design, copy, etc
- [x] Review changes with dev team
- [ ] Implement changes to views
- [ ] Write tests & doublecheck linting
```

svenaas:
Fixes #4108 

- Replaces JS alert popup with an `<AlertBanner>` containing a copyable link.

## security considerations
None. This improves the UX of accessing already-available information.

### DIFF
--- a/frontend/components/organization/Edit.jsx
+++ b/frontend/components/organization/Edit.jsx
@@ -21,14 +21,26 @@ function successNotification(message) {
   });
 }
 
-function showInviteAlert(email, link) {
-  // eslint-disable-next-line no-alert
-  return window.alert(`${email} does not have cloud.gov account yet, please send them the following link to get started:\n\n${link}`);
-}
-
 function showRemoveConfirm(member, org) {
   // eslint-disable-next-line no-alert
   return window.confirm(`Are you sure you want to remove ${member.User.UAAIdentity.email} from ${org.name}?`);
+}
+
+function getInvitationSentMsg(mostRecentlyAddedUser) {
+  return (
+    <span>
+      The new member must create a Pages account to accept the invitation to this organization.
+      They may check their inbox at
+      {' '}
+      {mostRecentlyAddedUser.email}
+      {' '}
+      for an invitation by email or visit
+      {' '}
+      <a href={`${mostRecentlyAddedUser.link}`}>{mostRecentlyAddedUser.link}</a>
+      {' '}
+      to get started.
+    </span>
+  );
 }
 
 function reducer(state, { type, payload }) {
@@ -38,6 +50,11 @@ function reducer(state, { type, payload }) {
         ...state,
         isLoading: false,
         ...payload,
+      };
+    case 'invitationSent':
+      return {
+        ...state,
+        mostRecentlyAddedUser: payload,
       };
     case 'addMember':
       return {
@@ -64,6 +81,7 @@ const initialState = {
   members: [],
   org: null,
   roles: [],
+  mostRecentlyAddedUser: null,
 };
 
 function Edit({ actions }) {
@@ -84,7 +102,7 @@ function Edit({ actions }) {
   }, ['1']);
 
   const {
-    isLoading, members, org, roles,
+    isLoading, members, org, roles, mostRecentlyAddedUser,
   } = state;
 
   if (isLoading) {
@@ -125,6 +143,15 @@ function Edit({ actions }) {
             alertRole={false}
           />
           )}
+        { mostRecentlyAddedUser
+          && (
+          <AlertBanner
+            status="info"
+            header="Sent invitation to create a Pages account"
+            message={getInvitationSentMsg(mostRecentlyAddedUser)}
+            alertRole={false}
+          />
+          )}
         <h3>Members</h3>
         <AddUserForm
           className="well"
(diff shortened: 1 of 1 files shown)


frontend/components/organization/Edit.jsx
```diff
@@ -108,6 +109,23 @@ function Edit({ actions }) {
 
   const getSandboxMsg = days => <span>{sandboxMsg(days)}</span>;
 
+  function getInvitationSentMsg() {
```
apburnes, line 112: Lets move this function out of the component and then pass in `mostRecentlyAddedUser` as an arg.

frontend/components/organization/Edit.jsx
```diff
@@ -108,6 +109,23 @@ function Edit({ actions }) {
...
+    return (
+      <span>
+        The new member must create a Pages account to accept the invitation to this organization.
+        They may check their inbox at
+        {' '}
+        {mostRecentlyAddedUser.email}
+        {' '}
+        for an invitation by email or visit
+        {' '}
+        <a href="{mostRecentlyAddedUser.link}">{mostRecentlyAddedUser.link}</a>
```
apburnes, line 122: To use the template literal syntax you will want 
```
<a href={`${mostRecentlyAddedUser.link}`}>
```

frontend/components/organization/Edit.jsx
```diff
@@ -125,6 +143,15 @@ function Edit({ actions }) {
             alertRole={false}
           />
           )}
+        { mostRecentlyAddedUser
+          && (
+          <AlertBanner
```
apburnes, line 148: Do we want anyway to clear the alert banner? Depending on the user's screen size and position if the form, the alert may be painted out of view. Do we want to do a modal or trap the focus with scroll to it so they see the alert for sure?

frontend/components/organization/Edit.jsx
svenaas: I had it like that originally. I moved it here for parallelism with the similarly-used `getSandboxMsg` right above it.

frontend/components/organization/Edit.jsx
svenaas: Do I? It's being interpolated as currently coded:
image

frontend/components/organization/Edit.jsx
svenaas: @sknep and I talked about that. There's already also the brief alert at upper right which shows up regardless of where the user is scrolled. The point of having this one stay is that users (at least the one user who opened the associated ticket) want to be able to copy the URL, and if it goes away that's going to be challenging.

frontend/components/organization/Edit.jsx
drewbo: The link/href itself is what needs to be changed

frontend/components/organization/Edit.jsx
drewbo: `getSandboxMsg` is still taking an argument (rather than relying on variables outside the function scope)

svenaas: This took an argument too until the linter flagged it as an error because `mostRecentlyAddedUser` was already defined. I could give the argument a different name to fix it but moving it where Andrew suggested is probably clearer.

frontend/components/organization/Edit.jsx
svenaas: _Ohhhhhh_ I see what you both mean. Gotcha; thanks for catching that and explaining!

frontend/components/organization/Edit.jsx
apburnes: Thanks for walking through the thought process. It makes sense.

apburnes (approved):
LGTM

frontend/components/organization/Edit.jsx
svenaas: Sure thing! The whole interaction could still use more work in the future. In the discussion on #4108 Sarah suggested alternate alerts to be shown when a user has been added to the org but no invitation was sent because they were already a Pages user, or when no invitation was sent because the user specified is already a member of the org, but differentiating those cases (or back end errors) will require other changes, and we didn't think it needed to hold this up.

PR #4412, opened 2024-03-07
feat: Update email invites based on user UAA origin #4370

### LINKED ISSUE #4370: Don't create a Cloud IDP registration link if the invited user's email means they'll be authorized by an Agency IDP
During invited user account creation, we build a hashed URL that will allow that email to register for Cloud.gov via Cloud.gov IDP. That registration link is only required for non-agency IDP (only Cloud.gov-provided UAA) users, but right now, it’s generated for every new account. On login, we always refer the user to their agency IDP if available. Although a Cloud.gov IDP registration link is issued to this user, they are _never_ able to accept it, because we will always check for agency IDP first. 

If we check the new user’s email domain against our IDPs when creating an unclaimed cloud.gov user account, we can probably tell whether they actually need a Cloud.gov IDP registration link or not, and only generate and send a link for emails that aren’t covered by an agency’s IDP, rather than everyone.

Screenshot 2024-01-19 at 2 59 00 PM

```[tasklist]
### Tasks
- [ ] Make it possible to tell whether an email address should be authorized through agency IDP or cloud IDP
- [ ] Update the account creation and invitation mail flow to not make links and send email to those accounts
```

apburnes:
Closes #4370 

- Updates the invite email based on the UAA user's origin.
  - If the user origin is from a connected agency IDP, send them a link to login
  - If the user origin is cloud.gov, send them to setup a cloud.gov account
- Removes Github to UAA auth migration functionality since the feature is no longer in use.

## security considerations
None


api/admin/controllers/user.js
```diff
@@ -170,7 +175,7 @@ module.exports = wrapHandlers({
       link,
     });
     if (link) {
-      await Mailer.sendUAAInvite(email, link);
+      await Mailer.sendUAAInvite(email, link, origin, org.name);
```
sknep, line 178: oooh does this mean i could use the origin domain and/or the org name in my invite email???

api/admin/controllers/user.js
```diff
@@ -192,7 +197,7 @@ module.exports = wrapHandlers({
       user,
     } = req;
 
-    const { email, inviteLink: link } = await OrganizationService.resendInvite(
+    const { email, inviteLink: link, origin } = await OrganizationService.resendInvite(
```
sknep, line 200: ok so some of them don't get org.name. If we did pass it on to the mailer, I think it would be lovely to let the user know which org they were invited to. I know it's only for the first org they join, but it would be nice if they could recognize their team/project in the invite email, and most folks are probably in just one org.

api/services/mailer/templates/uaaAgencyInvite.js
```diff
@@ -0,0 +1,48 @@
+const layout = require('./layout');
+
+function uaaInvite({ link, orgName }) {
+  return layout(`
+    <p>Hello!</p>
+    <p>
+      You've been invited to join
+      <a href="https://cloud.gov/pages">cloud.gov Pages </a> <sup>[1]</sup>.
+      as a member of the ${orgName} organization.
```
sknep, line 9: You read my mind <3 thanks!

test/api/unit/services/Mailer.test.js
```diff
@@ -23,16 +23,41 @@ describe('mailer', () => {
...
+        const origin = 'uaa';
+
+        const job = await Mailer.sendUAAInvite(email, link, origin, orgName);
 
         expect(job.name).to.eq('uaa-invite');
         expect(job.data.to).to.deep.eq([email]);
         expect(job.data.html).to.eq(Templates.uaaInvite({ link }));
       });
+
+      it('adds a `uaa-invite` job to the mail queue for agency IDP users', async () => {
```
sknep, line 50: Here you call it agency IDP, but the component name was uaaAgency. Could these match? Either UAA or IDP? (I prefer IDP)

sknep (approved):
+125 −419 

Love it. Small suggestions, not required.

api/admin/controllers/user.js
apburnes: I was pondering this too. The current dilemma is the admin UI is setup to resend invites to the user for the entire platform and it's not org specific. Also, we only need to resend an invite code if the user's origin is cloud.gov. I think we should examine the JTBD for an admin user in this scenario.

test/api/unit/services/Mailer.test.js
apburnes: Updated

sknep (approved):

PR #4422, opened 2024-03-19
update email layouts with new template

sknep:
This closes https://github.com/cloud-gov/pages-core/issues/4373

- Updates the layout.js email template to have a branded header and consistent footer for all Pages-sent emails, with some css style utilities available to emails that use the layout.
- Update the layout of the invite emails
- Slightly change the wording of the sandbox expiration emails
- Adds a few static images to be used in emails.

## security considerations
I can't think of any, but open to being wrong.

## Testing
I couldn't send test emails locally, so we will need to test this on dev to make sure it works as expected.


public/images/transactional-emails/GSA-logo.png
sknep, line 1: Is this the right place to upload images we want to stay online forever?

sknep:
I'm ready for reviews here -- lemme have it

api/services/mailer/templates/layout.js
```diff
@@ -1,16 +1,106 @@
-function layout(content) {
+const { app: { hostname } } = require('../../../../config');
+
+const IMAGE_HOST_DOMAIN = `${hostname}/images/transactional-emails`;
+
+const FOOTER_TEXT = 'cloud.gov is a product of the General Services Administration, 1800 F Street NW, Washington, DC 20405';
+
+// common styles
+const cssForPs = () => 'font-size:16px;line-height:26px;margin:16px 0;';
```
apburnes, line 8: It took me a bit to realize what the `Ps` and `As` meant. For better semantics what if we export an object with the common `p` and `a` css. In the future, it should be straight forward to add other element css styles.

ie
```js
// declare
const css = {
  a: 'font-size:16px;line-height:26px;margin:16px 0;',
  p: 'color:#067df7;text-decoration:none;',
  h1: ...
}

// import and use in other templates
<a style="{layout.css.a}"></a>
```

svenaas: I like this suggestion.

api/services/mailer/templates/alert.js
```diff
@@ -1,11 +1,11 @@
 const layout = require('./layout');
 
 function alert({ errors, reason }) {
-  return layout(`
-    <p><strong>${reason}</strong></p>
-    <p>
+  return layout.render(`
+    <p style="${layout.cssForPs}"><strong>${reason}</strong></p>
```
drewbo, line 5: These styles aren't getting added because the function isn't being called (`layout.cssForPs()`)

drewbo (changes requested):
Real good. One change, one question:

- In general, we prefer (but don't enforce yet) destructuring all imports. So `const { render, cssForPS } = require('./layout')`. It should be a little easier to read through the templates this way. You can totally ignore this if you want because it's currently inconsistent in the code base.
- I think the text is much, much better than previously. Is it worth preserving one link to any sort of documentation? Just in case someone is sent this with less than optimal context from their org manager

api/services/mailer/templates/uaaInvite.js
```diff
@@ -1,23 +1,24 @@
...
       ${link}
     </p>
-    <p style="${layout.cssForPs}">
+    <p style="${css.p}">
       <b>This invitation will expire after 24 hours.</b>
     </p>
-    <p style="${layout.cssForPs}">
-      We&rsquo;re here to help. If you need support, please email us at
+    <p style="${css.p}">
+      We&rsquo;re here to help. If you need support, check out <a href="https://cloud.gov/pages/documentation/#getting-started" style="${css.a}" target="_blank">our Getting Started guide</a> or email us at
```
sknep, line 19: @drewbo link to docs here

drewbo (approved):

PR #4429, opened 2024-03-22
Allow organization managers to resend invitations

### LINKED ISSUE #4371: Allow Org Mgrs to resend invites (since they expire every 24 hrs)
Cloud.gov IDP account registration links currently expire in 24 hours, and the manager cannot then re-invite them. Let's make sure that org managers can resend them if they wish, in cases where the invited user delayed, forgot, ignored, or lost the original invite email.

If the invited user's email domain matches an existing Agency IDP, we should not allow them to be re-invited, because they won't use the invite anyways. see https://github.com/cloud-gov/pages-core/issues/4370 

Make sure to check the invited user's email, not the _inviting_ user's email, because the org manager may be inviting someone who uses an email (and identity) from another agency IDP.

```[tasklist]
### Tasks
- [ ] Make it possible to resend cloud.gov IDP invites
- [x] Make it possible to tell whether an email address should be authorized through agency IDP or cloud IDP
- [ ] Update the UI to use ^^
```

svenaas:
Fixes #4371 

- Adds `isResend` option to the invite() API method
- Adds Resend Invite button to organization edit page, displayed for organization members with `uaa` or `cloud.gov` origins who have not yet signed in.
- Show last signin instead of last update in organization users list.

## security considerations
None. Allows organization managers to resend an invitation that they can already initially send.


drewbo (changes requested):
Broad question: is there a way to refactor this to use the existing (maybe slightly modified) invite method/route? It would keep the codebase smaller/simpler, reduce testing surface, etc

svenaas:
@drewbo I think that would be possible, just at the expense of making the Org controller method more complex. The resend case omits the construction and return of a member object (for updating the users table after invite), doesn't need the roleID parameter, and there might be other details. We could reduce the methods/routes but we'd still need to keep some of the extra testing logic.

frontend/util/federalistApi.js
```diff
@@ -62,6 +62,18 @@ export default {
     });
   },
 
+  resendInviteToOrganization(id, user, uaaEmail, roleId) {
```
drewbo, line 65: You can remove this method and call the above method `inviteToOrganization` with data in this "shape"

svenaas: I can! And I have.

frontend/components/organization/Edit.jsx
```diff
@@ -200,7 +201,7 @@ function Edit({ actions }) {
               <tr key={member.User.id}>
                 <th scope="row" data-title="Email">{member.User.UAAIdentity.email}</th>
                 <td data-title="Role">
-                  <span className="usa-sr-only">No user to remove</span>
```
sknep, line 203: someday, i'm going to do a usability and accessibility audit of Pages...

public/swagger/User.json
```diff
@@ -17,6 +17,9 @@
     },
     "hasGithubAuth": {
       "type": "boolean"
+    },
+    "signedInAt": {
+      "type": "date-time"
```
sknep, line 22: hmm... `signedInLastAt`? `lastSignIn` ?

test/api/requests/organization.test.js
```diff
@@ -154,7 +154,7 @@ describe('Organization API', () => {
       validateAgainstJSONSchema('POST', '/organization/{id}/invite', 400, response.body);
     });
 
-    it('returns the member and the invitation details for cloud.gov origin', async () => {
+    it('returns the member and the invitation details for cloud.gov origin on invite', async () => {
```
sknep, line 157: this is the smallest of nits, but I read this like `invite` is a noun, not a verb. `upon invite` or `upon invitation` or `when invited` or something would help future me know what you meant by returns...details...on invite

public/swagger/User.json
svenaas: That's the name of the existing field, already surfaced through the API for admins. It's just newly available for others.

sknep:
odd question -- what happens if an org manager invites someone twice in a 24 hour span? presumably they can click the button a bunch of times. Do the old links get invalidated? will any work? that's probably ok, i just wanted to know, since we let them do it.

public/swagger/User.json
sknep: those `adminerers` and their awful variable names

drewbo:
@sknep I think any of the invitations will work (just tested on myself but didn't bother setting up MFA)

drewbo (approved):

test/api/requests/organization.test.js
svenaas: OK, changed.

sknep (dismissed):
lookin' good

frontend/components/organization/Edit.jsx
```diff
@@ -239,6 +240,26 @@ function Edit({ actions }) {
                       }
                     }
                   />
+                  { member.User.UAAIdentity.origin === 'uaa' && !member.User.signedInAt && (
```
apburnes, line 243: This should also include `cloud.gov` as an origin to check.

svenaas: All right; will that literally be the string `'cloud.gov'` in origin?

apburnes: Yes

svenaas: Thanks! Done.

apburnes (commented):
Do you want to add stretch task to this PR and change the `Updated` column in the org users table to be `Last Signed In` and use the `member.User.signedInAt` value in place of the `member.updatedAt`? The current info for updated is not very useful compared to last sign in.

svenaas:
Stretch task completed, @apburnes.

frontend/components/organization/Edit.jsx
```diff
@@ -257,6 +279,7 @@ Edit.propTypes = {
     fetchOrganization: PropTypes.func.isRequired,
     fetchRoles: PropTypes.func.isRequired,
     inviteToOrganization: PropTypes.func.isRequired,
+    resendInviteToOrganization: PropTypes.func.isRequired,
```
apburnes, line 282: This should be optional since not every user will get a resend invite.

svenaas: Thanks for catching that, @apburnes. Just pushed a fix.

apburnes (approved):

PR #4444, opened 2024-04-01
fix: Fetch error for old build logs based on build details

apburnes:
- Does not fetch build logs if build completed at is over 180 days

## security considerations
None

### DIFF
--- a/frontend/components/site/siteBuildLogs.jsx
+++ b/frontend/components/site/siteBuildLogs.jsx
@@ -13,26 +13,38 @@ export const REFRESH_INTERVAL = 15 * 1000;
 const BUILD_LOG_RETENTION_LIMIT = 180 * 24 * 60 * 60 * 1000; // 180 days in milliseconds
 
 function buildTooOld(build) {
-  return (new Date() - new Date(build.completedAt)) > BUILD_LOG_RETENTION_LIMIT;
+  return new Date() - new Date(build?.completedAt) > BUILD_LOG_RETENTION_LIMIT;
 }
 
 function getSiteBuildLogTable(buildDetails, logs, state) {
+  if (buildTooOld(buildDetails)) {
+    return (
+      <SiteBuildLogTable
+        buildLogs={[
+          'Builds more than 180 days old are deleted according to platform policy.',
+        ]}
+      />
+    );
+  }
+
   if (logs && logs?.length > 0) {
     return <SiteBuildLogTable buildLogs={logs} buildState={state} />;
   }
-  if (buildTooOld(buildDetails)) {
-    return <SiteBuildLogTable buildLogs={['Builds more than 180 days old are deleted according to platform policy.']} />;
-  }
-  return <SiteBuildLogTable buildLogs={['This build does not have any build logs.']} />;
+
+  return (
+    <SiteBuildLogTable
+      buildLogs={['This build does not have any build logs.']}
+    />
+  );
 }
 
 const SiteBuildLogs = () => {
   const { buildId: buildIdStr } = useParams();
   const buildId = parseInt(buildIdStr, 10);
-  const { logs, state } = useBuildLogs(buildId);
-  const { buildDetails, isLoading } = useBuildDetails(buildId);
+  const { buildDetails, isLoading: isLoadingBuildDetails } = useBuildDetails(buildId);
+  const { logs, state, isLoading: isLoadingBuildLogs } = useBuildLogs(buildId);
 
-  if (isLoading || !logs) {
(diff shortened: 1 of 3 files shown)


frontend/components/site/siteBuildLogs.jsx
```diff
@@ -12,27 +12,46 @@ import CommitSummary from './CommitSummary';
 export const REFRESH_INTERVAL = 15 * 1000;
 const BUILD_LOG_RETENTION_LIMIT = 180 * 24 * 60 * 60 * 1000; // 180 days in milliseconds
 
-function buildTooOld(build) {
-  return (new Date() - new Date(build.completedAt)) > BUILD_LOG_RETENTION_LIMIT;
+function shouldHaveBuildLogs(build) {
```
drewbo, line 15: this function (and the later logic relying on it) are pretty confusing patterns because they treat null and false differently. Is there a way to ensure we don't call this without a build?

frontend/hooks/useBuildLogs.js
```diff
@@ -40,8 +42,14 @@ export const useBuildLogs = (id) => {
...
-    fetchData(results.offset);
-  }, [results]);
+    if (hasBuildLogs) {
+      fetchData(results.offset);
+    }
+
+    if (results.isLoading && !hasBuildLogs) {
+      setResults({ ...initResultsState, isLoading: false });
+    }
+  }, [results, hasBuildLogs, initResultsState]);
```
drewbo, line 52: Does `initResultsState` need to be in here if it's non-reactive?

drewbo (changes requested):

frontend/components/site/siteBuildLogs.jsx
```diff
@@ -13,26 +13,38 @@ export const REFRESH_INTERVAL = 15 * 1000;
...
 const SiteBuildLogs = () => {
   const { buildId: buildIdStr } = useParams();
   const buildId = parseInt(buildIdStr, 10);
-  const { logs, state } = useBuildLogs(buildId);
-  const { buildDetails, isLoading } = useBuildDetails(buildId);
+  const { buildDetails, isLoading: isLoadingBuildDetails } = useBuildDetails(buildId);
+  const { logs, state, isLoading: isLoadingBuildLogs } = useBuildLogs(buildId);
 
-  if (isLoading || !logs) {
+  if (isLoadingBuildDetails || isLoadingBuildLogs || !buildDetails) {
```
drewbo, line 47: What's the case where we are done loading build details but also don't have them?

frontend/hooks/useBuildLogs.js
```diff
@@ -7,6 +7,8 @@ const initResultsState = {
   offset: 0,
   logs: [],
   state: '',
+  isLoading: true,
```
drewbo, line 10: This is good; I thought about this after our review yesterday and we should codify this pattern for data loaders

frontend/hooks/useBuildLogs.js
```diff
@@ -35,8 +39,14 @@ export const useBuildLogs = (id) => {
...
       }
 
-      if (outputCount > 0 && ['success', 'error'].includes(state)) {
-        setResults(updatedResults);
+      if (['success', 'error'].includes(state)) {
+        if (output?.length > 0) {
+          setResults(updatedResults);
+        }
+
+        if (output?.length === 0 && results.totalRequests === 0) {
```
drewbo, line 47: Should this be `results.totalRequests > 0`? My assumption is that we want to still set the results if we didn't get output but have made at least one request

drewbo (changes requested):

frontend/hooks/useBuildLogs.js
apburnes: There are situations where errored builds may or may not have logs so we update the state of just the initial request if there are no logs. If it was `results.totalRequests > 0` it would cause an infinite loop since we fetch the api results recursively.

frontend/components/site/siteBuildLogs.jsx
apburnes: Waiting on `buildDetails` prevents a rendering flash since the `getSiteBuildLogTable` function depends on both `buildDetails` and  the `useBuildLogs` results to determine how the component will be displayed.

drewbo (approved):

frontend/components/site/siteBuildLogs.jsx
drewbo: Covered via voice, can be removed

drewbo (approved):

PR #4478, opened 2024-04-11
Improve build scans display on build history and scans pages (#4446)

### LINKED ISSUE #4446: update frontend scan information "rollups"
Now that more build tasks exist and are running, we need to update the information displayed on the build card:
- For builds with no build tasks, we should show nothing (especially during the beta rollout)
- For builds with any build tasks, even if they aren't complete, the message should be clickable to go through to the `/scans` page
- We should review [this function](https://github.com/cloud-gov/pages-core/blob/main/frontend/components/site/siteBuilds.jsx#L90) to ensure it's providing meaningful information to the user about the status/results of their various tasks

sknep:
Closes #4446

- Refine the summaries for Build status and Scan status
- Reconfigure the Build row so that there's not a separate "results" column for build logs and build scans
- Add a special treatment to a table cell that features the Build scans full-width
- Some new icons added to better reflect these statuses and features. Also the spinner icon now actually spins, with the magic of CSS.

This still only displays Scan-related DOM when the feature flag for scans is enabled, and (where applicable) only for builds with associated scans.

## security considerations
none


frontend/components/site/CommitSummary.jsx
```diff
@@ -66,7 +66,7 @@ CommitSummary.propTypes = {
     branch: PropTypes.string.isRequired,
     username: PropTypes.string.isRequired,
     clonedCommitSha: PropTypes.string.isRequired,
-    createdAt: PropTypes.instanceOf(Date).isRequired,
+    createdAt: PropTypes.string.isRequired,
```
sknep, line 69: This was creating an error in the console. I know it "should" be a date but it's not by the time it gets here.

frontend/components/site/siteBuildTasks.jsx
```diff
@@ -108,15 +111,12 @@ const SiteBuildTasks = () => {
               </thead>
               <tbody>
                 {buildTasks.map((task) => {
```
sknep, line 113: If this gets any more complicated, I'm promoting it to its own component

frontend/components/site/siteBuilds.jsx
```diff
@@ -89,41 +92,56 @@ function branchLink(build) {
 
 function summarizeTaskResults(build) {
```
sknep, line 93: This could be its own child component too.

frontend/sass/_tables.scss
```diff
@@ -27,34 +27,60 @@
...
-        margin: 0;
-        white-space: initial;
-        line-height: 1.4em;
-      }
-      h3 {
-        margin: 0;
-        font-size: 1.4em;
-      }
+  }
+  .build-info-prefix {
```
sknep, line 41: someday i'm gonna BEM the heck out of our DOM

frontend/sass/_tables.scss
```diff
@@ -66,16 +92,19 @@
...
+  }
   .branch-info {
-    // white-space: nowrap;
-    // overflow: hidden;
-    // text-overflow: ellipsis;
-    // max-width: 30ch;
-    margin: 0.25em 0;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
sknep, line 101: do you remember why we disabled branch name truncation at the last minute?

test/frontend/util/datetime.test.js
```diff
@@ -11,11 +11,11 @@ describe('datetime', () => {
 
   describe('duration()', () => {
     it('shows a human readable duration when provided two dates', () => {
-      expect(duration(startTime, endTime)).to.equal('4 minutes');
+      expect(duration(startTime, endTime)).to.equal('3 minutes');
     });
 
     it('works if endTime isn\'t defined', () => {
-      expect(duration(+new Date())).to.equal('less than a minute');
+      expect(duration(+new Date())).to.equal('0 seconds');
```
sknep, line 18: The "strict" formatter is a little shorter, but is more precise. date-fns doesn't offer the floor rounding method on the regular formatter, and before, the two duration formatter time utils would give different results for the same duration.

frontend/components/site/siteBuilds.jsx
```diff
@@ -206,29 +229,64 @@ function SiteBuilds() {
...
+                                  { duration(build.startedAt, build.completedAt) }
+                                </>
+                              )}
+                              {(!done) && (
+                                <>
+                                  {message}
+                                  {' '}
+                                  for
+                                  {' '}
+                                  { duration(build.createdAt, build.completedAt) }
```
drewbo, line 284: I'm confused as to why this works (completedAt when not done?) but it does...

frontend/components/site/CommitSummary.jsx
```diff
@@ -49,9 +49,9 @@ function CommitSummary({ buildDetails }) {
         &nbsp;
         <span
           className="commit-time"
-          title={dateAndTime(buildDetails?.createdAt)}
+          title={dateAndTime(new Date(buildDetails?.createdAt))}
         >
-          {timeFrom(buildDetails?.createdAt)}
+          {timeFrom(new Date(buildDetails?.createdAt))}
```
drewbo, line 54: This time doesn't represent the commit time (we're duplicating the build time)

drewbo (changes requested):
Looks great, excited to beta test. A few minor issues noted

frontend/components/site/siteBuildTasks.jsx
```diff
@@ -130,28 +130,35 @@ const SiteBuildTasks = () => {
...
-                                Download scan results
-                              </Link>
-                              <span className="artifact-filesize">
-                                &nbsp;({ prettyBytes(task.artifact.size) })
-                              </span>
+                              {summary}
                             </li>
-                          )}
-                        </ul>
+                            {task.status === 'success' && task.artifact && task.artifact.url && (
```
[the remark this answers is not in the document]
sknep: I'm too used to the struggle 😅 will update

frontend/components/site/siteBuilds.jsx
```diff
@@ -89,41 +97,56 @@ function branchLink(build) {
...
+      {' '}
+      {build.BuildTasks.length}
+      {' '}
+      build scan(s)
+      {' '}
+      {summaryVerb}
+      {tasksWithResults.length > 0 && (totalResults(tasksWithResults))}
+      {' '}
+      (
+      {scansDocsLink('/link/url/here')}
```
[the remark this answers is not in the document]
sknep: it's already live! wowza

frontend/components/site/siteBuilds.jsx
```diff
@@ -89,41 +97,56 @@ function branchLink(build) {
...
+        <IconExperiment />
+        Public beta:
+      </b>
+      {' '}
+      {build.BuildTasks.length}
+      {' '}
+      build scan(s)
+      {' '}
+      {summaryVerb}
+      {tasksWithResults.length > 0 && (totalResults(tasksWithResults))}
```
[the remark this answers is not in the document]
sknep: Yeah, I wondered if you'd catch that. ;)  Would we want to say that the build scans are still in progress if _any_ are in progress, instead of complete if _any_ are complete?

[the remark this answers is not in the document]
drewbo: A few ideas:
- If we have the horizontal screen real estate, ideal is `1 scan processing, 1 complete with X issues`
- Maybe next best is `1 complete with X issues` which doesn't tell the user that another is still processing but they'll have the link which will show them
- Any other option feels like a lie but I could be swayed

frontend/components/site/siteBuilds.jsx
```diff
@@ -280,19 +320,30 @@ function SiteBuilds() {
...
                       </td>
+                      { process.env.FEATURE_BUILD_TASKS === 'active' && (
+                        hasBuildTasks(build)
+                          ? (
+                            <td data-title="Scans" className="scan-results-pseudo-row">
+                              { summarizeTaskResults(build) }
+                            </td>
+                          )
+                          : (
+                            <td className="no-scan-results usa-sr-only">No scans for this build</td>
```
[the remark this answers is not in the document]
drewbo: An assortment of thoughts:
- Screenreader and visual understanding should match as much as we can
- Totally independent of whether this feature is active, users without any build scans shouldn't have them displayed/mentioned (although maybe this is only true during the beta)
- One potential solution is not treating the scan portion as a `<td>` (goes back to our card vs table row discussion) which avoids the need to set a column header
- A quick proxy for whether the site has any tasks (across all builds) is whether `Site.SiteBuildTasks` has any results but we don't currently return that from the API
- Also it's theoretically very quick to check if the displayed builds have tasks (something like `builds.some(build => build.BuildTasks.length)` and it will correctly be true per page)

frontend/components/site/siteBuilds.jsx
sknep: the second param is not required; if not provided, it uses now.

frontend/components/site/CommitSummary.jsx
sknep: Not quite following this comment, maybe you could elaborate? is `new Date()` wrong?

frontend/components/site/siteBuilds.jsx
sknep: does thumbs up here mean do it now?

drewbo: Just general agreement, no need to add now

frontend/components/site/CommitSummary.jsx
sknep: I declare fixing this is out of scope <3

drewbo (dismissed):

apburnes (approved):
:accessibility:  🔐

sknep:
> @[apburnes] [force-pushed] the feat/update-build-scan-rollups-4446 branch
ty

PR #4486, opened 2024-04-17
chore: add build task cancelled state

drewbo:
- Add build task cancelled state
- Only show link to scans from logs page if there are scans for that build
- Close https://github.com/cloud-gov/pages-core/issues/4485

## security considerations
None

### DIFF
--- a/api/models/build.js
+++ b/api/models/build.js
@@ -168,15 +168,21 @@ const afterCreate = async (build) => {
 const afterUpdate = async (build) => {
   // we start certain build tasks on build completion
   const { BuildTask, BuildTaskType } = build.sequelize.models;
-  if (build.state === States.Success) {
-    try {
+  try {
+    if (build.state === States.Success) {
       const buildTasks = await BuildTask
         .byStartsWhen(BuildTaskType.StartsWhens.Complete)
         .findAll({ where: { buildId: build.id, status: BuildTask.Statuses.Created } });
       await Promise.all(buildTasks.map(async task => task.enqueue()));
-    } catch (err) {
-      console.error(err);
+    } else if (build.state === States.Error) {
+      const buildTasks = await BuildTask
+        .findAll({ where: { buildId: build.id } });
+      await Promise.all(buildTasks.map(async task => (
+        task.update({ status: BuildTask.Statuses.Cancelled })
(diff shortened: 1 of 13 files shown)


frontend/components/site/siteBuildLogs.jsx
```diff
@@ -38,10 +38,15 @@ function getSiteBuildLogTable(buildDetails, logs, state) {
   );
 }
 
+function hasBuildTasks(buildTasks) {
+  return process.env.FEATURE_BUILD_TASKS === 'active' && buildTasks?.length > 0;
```
apburnes, line 42: Does this account for a site branch not having any build tasks associated to it?

api/models/build.js
```diff
@@ -154,29 +154,27 @@ const afterCreate = async (build) => {
...
+    .byStartsWhen(BuildTaskType.StartsWhens.Build)
+    .findAll({ where: { buildId: build.id } });
+  await Promise.all(buildTasks.map(async task => task.enqueue()));
 };
 
 const afterUpdate = async (build) => {
   // we start certain build tasks on build completion
   const { BuildTask, BuildTaskType } = build.sequelize.models;
   if (build.state === States.Success) {
-    try {
```
apburnes, line 172: Where do the errors get handled on a model hook?

frontend/components/site/siteBuildLogs.jsx
```diff
@@ -38,10 +38,15 @@ function getSiteBuildLogTable(buildDetails, logs, state) {
   );
 }
 
+function hasBuildTasks(buildTasks) {
```
apburnes, line 41: Can we move this into the `useBuildTasks` hook to keep all the build tasks state there?

api/models/build.js
drewbo: I had added these a long time ago just as a safety measure during testing; I think it would probably crash the application since the model hooks occur outside the web request function call stack. Want me to restore and use something other than `console.error`?

apburnes: Can we pull in our event model to capture the error and log it with our logger?

apburnes (approved):

apburnes (approved):

PR #4521, opened 2024-05-29
Add build metrics query examples

### LINKED ISSUE #4500: Demonstrate the queryability of build metrics
Now that we have build metrics in the database (#4495), we should document how to use them. 

## Acceptance Criteria

- [ ] Document a few sample queries, including
  - [ ] How to find outliers in a build performance metric
  - [ ] How to identify all Org Managers using a site on node v18

svenaas:
Fixes #4500

- Updates OPERATIONS.md with sample build metrics queries

## security considerations
None. This documents possible queries against an open source schema.

### DIFF
--- a/docs/OPERATIONS.md
+++ b/docs/OPERATIONS.md
@@ -63,3 +63,59 @@ This is a CI task that runs nightly in order to rotate a subset of site bucket k
 Task name: `nightly-site-bucket-key-rotator`
 Task script: [`rotate-bucket-keys.js`](./ci/tasks/rotate-bucket-keys.js)
 Task partial: [`rotate-bucket-keys.yml`](./ci/partials/rotate-bucket-keys.yml)
+
+## Querying
+
+### Build metrics
+
+It may be useful to query the database directly to identify sites that are outliers in build resource usage, or those which rely on engines close to the end of their lives. Here are some examples:
+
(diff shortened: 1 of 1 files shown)


drewbo (changes requested):
@svenaas this looks good, I'd recommend keeping only three for brevity:
- either the disk or memory query (these are the most likely to be interesting, I'd set the memory limit at 2GB if you choose that since it's the limit for small containers)
- all sites using Node/Ruby
- all sites using X with site manager email

svenaas:
@drewbo 
> * either the disk or memory query (these are the most likely to be interesting, I'd set the memory limit at 2GB if you choose that since it's the limit for small containers)

I think memory is the more interesting but I think I misinterpreted the units as GB. It [looks like](https://github.com/cloud-gov/pages-build-container/blob/206740e6fd1b8bff30b79f36c12dc7a0bbed959b/src/log_utils/monitoring.py) what's logged is actually a _percentage_ of [virtual memory](https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory), which will vary by container) — am I reading that correctly? If so, maybe we should set the query threshold at 75% or something like that.

drewbo:
Between the two, memory is probably slightly less interesting since people run into it less frequently. Disk space is a common issue for build failures

PR #4523, opened 2024-05-29
send 401 response for bad auth callback

### LINKED ISSUE #4512: Wrong callback code should throw 4xx error rather than 500
These endpoints should throw a non-500 error
- https://pages.cloud.gov/admin/auth/uaa/callback?code=nonsensecode
- https://pages.cloud.gov/auth/uaa/callback?code=nonsensecode

## Acceptance Criteria

- [ ] Wrong callback code throws 4xx error rather than 500

drewbo:
- send 401 response for bad auth callback
- close #4512

## security considerations
Change to our OAuth callback flow. In theory it implements the same pattern as before but with additional error handling

### DIFF
--- a/api/routers/auth.js
+++ b/api/routers/auth.js
@@ -3,14 +3,25 @@ const { Event, User } = require('../models');
 const passport = require('../services/passport');
 const EventCreator = require('../services/EventCreator');
 
-const opts = {
-  failureRedirect: '/',
-  failureFlash: true,
-};
-
 function redirectIfAuthenticated(req, res, next) {
   // eslint-disable-next-line no-unused-expressions
   req.session.authenticated ? res.redirect('/') : next();
+}
+
+function uaaCallback(req, res, next) {
+  // provide this interior callback in place of the default passport.authenticate behavior:
+  // https://github.com/jaredhanson/passport/blob/master/lib/middleware/authenticate.js#L34
+  passport.authenticate('uaa', (err, user, info) => {
+    if (err) {
+      return res.unauthorized();
+    }
+    if (user) {
+      return req.logIn(user, next);
+    }
+    // no user, no error
+    req.session.flash = { error: [info.message] };
+    return res.redirect('/');
+  })(req, res, next);
 }
 
 function onSuccess(req, res) {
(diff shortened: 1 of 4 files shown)


svenaas (approved):
Looks good, glad for more tests, `/auth/uaa/callback?code=nope` returns a 403, approved.

drewbo:
One other thing I'd like to confirm on dev before merging, does anyone have a UAA login not connected to pages? I'd like to confirm the "you aren't authorized" path also

api/routers/auth.js
```diff
@@ -13,18 +13,38 @@ function redirectIfAuthenticated(req, res, next) {
...
+  passport.authenticate('uaa', (err, user, info) => {
+    if (err) {
+      return res.forbidden();
+    }
+    if (user) {
+      req.session.passport = { user: user.id };
+      req.session.authenticated = true;
+      req.session.authenticatedAt = new Date();
+    } else {
+      req.session.flash = { error: [info.message] };
```
apburnes, line 36: Was passport-oauth2's built-in callback doing anything to explicitly clear a session on an undefined `user`?

drewbo: Short answer: no; also there shouldn't be any relevant session information if the user comes back undefined. Longer answer:

The default behavior of `passport.authenticate` is detailed [here](https://github.com/jaredhanson/passport/blob/master/lib/middleware/authenticate.js). Providing a callback will cause an early return for three strategy methods: `allFailed` (all provided strategies failed to authenticate), `success` and `error`.
- Default error behavior is `next(error)` , we override with `res.forbidden` (though we should probably change to 401)
- Default failure behavior is to (based on options) set flash messages, session failure messages, and redirect. We show the flash messages and redirect.
- Default success behavior is mostly based on options and then calls [`req.logIn`](https://github.com/jaredhanson/passport/blob/0575de90dc0e76c1b8ca9cc676af89bd301aec60/lib/http/request.js#L24-L46) with the user. I'm going to rewrite our callback to call this rather than interacting with the session directly

api/routers/auth.js
```diff
@@ -13,18 +13,38 @@ function redirectIfAuthenticated(req, res, next) {
   req.session.authenticated ? res.redirect('/') : next();
 }
 
-function onSuccess(req, res) {
-  req.session.authenticated = true;
-  req.session.authenticatedAt = new Date();
-  req.session.save(() => {
-    res.redirect(req.session.authRedirectPath || '/');
-  });
+// function onSuccess(req, res, next) {
```
apburnes, line 16: Do you want to remove this commented out code?

api/routers/auth.js
apburnes: Thanks for the detail with the deep dive.

api/routers/auth.js
drewbo: yeah, I'm going to push another change to the callback based on your other questions/feedback

drewbo:
Confirmed the non-authorized path on dev

PR #4531, opened 2024-06-06
Feat build scan config

### LINKED ISSUE #4261: Implement User-side Config for Build Tasks (ZAP scans, et al)
Decide how configuration of a build task should be handled in the UI + task itself. 
## Acceptance Criteria

- [ ] Determine next steps and acceptance criteria

drewbo:
- Close #4261 

## security considerations
Users are able to configure the rules used in security and accessibility scans. Information about how this is used will be published in the reports, our site, and in this application prior to going to production


api/models/build-task.js
```diff
@@ -33,6 +34,9 @@ const associate = ({
     foreignKey: 'buildTaskTypeId',
     allowNull: false,
   });
+  BuildTask.belongsTo(SiteBuildTask, {
+    foreignKey: 'siteBuildTaskId',
```
apburnes, line 38: Since we don't explicitly set `allowNull: false`, does that mean we allow null on build task's site build task id?

api/utils/cfApiClient.js
```diff
@@ -178,10 +178,26 @@ class CloudFoundryAPIClient {
...
-    // TODO: link to template documentation
+    // https://github.com/cloud-gov/pages-cf-build-tasks/?tab=readme-ov-file#docker-command
     const template = parse(task.BuildTaskType.metadata.template);
 
-    const encryptedTask = Encryptor.encryptObjectValues(task, encryption.key);
+    // prior to adding to the template, SiteBuildTask metadata needs to be combined
+    // with the global rules and stringified. we add this to a separate object to
+    // avoid parameter assignment on the task (in case of retries)
+    const taskToEncrypt = JSON.parse(JSON.stringify(task));
+    const rulesForType = DEFAULT_SCAN_RULES[appMatch(taskToEncrypt.BuildTaskType)];
```
apburnes, line 188: Do we need to stringify these as well?

frontend/components/site/ScanConfig/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+import { useDefaultScanRules } from '../../../hooks/useDefaultScanRules';
+import { useSiteBuildTasks } from '../../../hooks/useSiteBuildTasks';
+
+function appMatch(type) {
+  // TODO: somehow remove this function or unduplicate from api utils
+  // get a readable string from a BuildTaskType appName to match the scan rule types
+  return type.metadata.appName.replace(/pages-(.*)-task-.*/, '$1');
+}
+
+function configToRuleArray(config) {
```
apburnes, line 15: Should this be in the useDefaultScansHook so we get the data structure we expect via the hook?

frontend/components/site/ScanConfig/index.jsx
```diff
@@ -0,0 +1,140 @@
+import React, { useState, useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+
+import api from '../../../util/federalistApi';
+import { BUILD_SCAN_RULES } from '../../../util/scanConfig';
+import { useDefaultScanRules } from '../../../hooks/useDefaultScanRules';
+import { useSiteBuildTasks } from '../../../hooks/useSiteBuildTasks';
+
+function appMatch(type) {
```
apburnes, line 9: Can we match the app on the API response?

apburnes (commented):
It looks pretty good and just did an initial pass with a few questions. Let me know if you want to do a demo and walk through your thoughts.

api/utils/cfApiClient.js
drewbo: This first stringify is really just for doing deep clone, the stringify below actually runs over the combined rulesets

api/models/build-task.js
drewbo: Yeah I was allowing it for now unless we want to go back and clear out old build tasks (assigning them retroactively is possible but tough)

apburnes: That's what I thought but just wanted to verify. I think it's fine to have old build tasks with no site id in the record.

PR #4533, opened 2024-06-13
fix: Only encrypt site build param values based on defined keys

apburnes:

- Switches to just encrypt values for predefined keys sent to the CF task as a parameter
- This work is related to the `4096` character cap on the command passed to a CF task

## security considerations
None

### DIFF
--- a/api/services/Encryptor.js
+++ b/api/services/Encryptor.js
@@ -3,9 +3,9 @@ const _ = require('underscore');
 
 const ALGORITHM = 'aes-256-gcm';
 
-function encrypt(value, key, { hintSize = 4 } = {}) {
+function encrypt(value, encryptionKey, { hintSize = 4 } = {}) {
   // Create a 32 byte hash from the secret key
-  const hashedKey = Crypto.createHash('sha256').update(key).digest();
+  const hashedKey = Crypto.createHash('sha256').update(encryptionKey).digest();
 
   // Generate a random 16 byte initialization vector
   const iv = Crypto.randomBytes(16);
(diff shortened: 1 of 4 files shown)


api/utils/cfApiClient.js
```diff
@@ -247,11 +247,28 @@ class CloudFoundryAPIClient {
...
+          'AWS_ACCESS_KEY_ID',
+          'AWS_SECRET_ACCESS_KEY',
+          'BUCKET',
+        ],
+      }
     );
-    const command = `cd app && python main.py -p '${encryptedParams.ciphertext}'`;
+    const commandParamsString = JSON.stringify(encryptedParams);
+
+    if (commandParamsString.length >= 4097) {
```
drewbo, line 265: this check should probably be against 4069 because it also counts the ~28 characters from the command template string below against the limit

api/services/Encryptor.js
```diff
@@ -33,30 +33,34 @@ function encrypt(value, key, { hintSize = 4 } = {}) {
...
 
-function encryptObjectValues(obj, key, { hintSize = 4 } = {}) {
-  return _.mapObject(obj, (value) => {
+function encryptObjectValues(
+  obj,
+  encryptionKey,
+  { hintSize = 4, onlyEncryptKeys = [] } = {}
+) {
+  return _.mapObject(obj, (value, key) => {
     if (_.isArray(value) || _.isFunction(value)) {
```
drewbo, line 42: I'm not sure that this is part of the issue but at the same time we're here, we should change `_isArray(value)` to return `_.map(v => encryptObjectValues(v, ...)`

drewbo (changes requested):

api/services/Encryptor.js
apburnes: It's not recursively mapping arrays with an item that is an array. Let me know what you think.

drewbo (approved):

PR #4546, opened 2024-06-22
Feature: style build scans ignore config

sknep:
- Moved scan config from it's own page to the Site settings page
- Split the stuff to ignore into separate rulesets by task id
- styled a table that would work for this

### Minor UX edits outside of just this work
- swapped the order of the env variables save buttons because they didn't match the rest of the settings page button order, made the labels match this ("save" vs "add")
- made the HTML  `input` border width less thick. They've always seemed rather heavy and particularly within a table inside an accordion, it was a little much. 
- the makefile being out of date was bugging me

## security considerations
N/A 

## documentation considerations
Some screenshots might be a little out of date (moved buttons, thinner borders)... I don't think it would affect comprehension though.

Screenshot 2024-06-21 at 7 25 46 PM
Screenshot 2024-06-21 at 7 25 38 PM


Makefile
```diff
@@ -19,44 +19,50 @@ db: ## Connect to the database - it must already be running ie with `make start`
...
 test-all: ## Run all tests
-	docker-compose run --rm app yarn test
\ No newline at end of file
+	docker compose run --rm app yarn test
+
+everything: # When you switch to a new branch and need to rebuild everything
+	make install
+	make migrate
+	make seed
+	make rebuild
```
drewbo, line 67: rebuild should happen first, then everything else looks right

api/utils/index.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable object-curly-newline */
+/* eslint-disable object-curly-newline */
```
drewbo, line 2: Not sure why this is doubled but it could be removed

drewbo (dismissed):
Looks great, thanks @sknep 

**issue (non-blocking)** It's my fault for introducing this but is there a way to better handle new elements prior to the user selecting a rule? Right now it creates an info link with no href but it will gladly save this "empty" rule.

I'm open to merging as is and fixing this at a later date

sknep:
i commit to fixing the empty info link

frontend/components/site/SiteSettings/EnvironmentVariables.jsx
```diff
@@ -81,7 +82,7 @@ class EnvironmentVariables extends Component {
     return (
       <>
         <h3>Environment Variables</h3>
```
apburnes, line 84: This is off topic for this PR, but with the addition of the trash icon for deleting scan rules, should we update the environment variable table to use the trash icon button to delete an environment variable?

frontend/components/site/SiteSettings/ScanConfigs.jsx
```diff
@@ -0,0 +1,271 @@
...
+
+  function handleUpdate(siteBuildTask) {
+    const typeRules = customerRules
+      .filter(rule => rule.type === siteBuildTask.id)
+      .filter(rule => rule.id.slice(0, 4) !== 'temp'); // don't save temp rules
+
+    return api.updateSiteBuildTask(id, siteBuildTask.sbtId, { rules: typeRules })
+      .then(() => notificationActions.success('Successfully saved scan configuration.'))
+      .catch(() => notificationActions.success('Error saving scan configuration.'))
+      .finally(() => setIsLoading(false));
```
apburnes, line 44: The `Save all rules` button is always enabled no matter if there are no changes or the config is being updated. Should we expand the use of useState to disable the button to add additional context to the user?

apburnes (commented):
I had just two thoughts but it looks great and functions well.

apburnes (approved):

PR #4569, opened 2024-08-07
chore: add zap scan

drewbo:
- Adds ZAP scanning for core/admin/queues
- Uploads the results to S3

## security considerations
This adds an admin test user to the database to scan sites, it is removed after scanning.


scripts/create-test-users.js
```diff
@@ -4,7 +4,7 @@ const factory = require('../test/api/support/factory');
 const { authenticatedSession } = require('../e2e/auth-session');
 
 async function createUsers() {
-  const user = await factory.user({ username: 'test-e2e-user' });
+  const user = await factory.user({ username: process.env.PAGES_TEST_USER || 'generic-test-user' });
```
apburnes, line 7: Do we want to update this user to belong to an org? Manage an org?

scripts/count-yaml.sh
apburnes, line 1: What is the purpose of the count-yaml.sh script?

ci/pipeline.yml
```diff
@@ -516,6 +592,8 @@ resources:
   - name: redis
   - name: postgres
   - name: node
+  - name: zap
+  - name: dind
```
apburnes, line 596: Removing the version will make updates across all of our pipelines much easier.

apburnes (commented):
Just a few questions but this is a great addition to our process. Many hours per month will be saved for better use. 🏆

scripts/count-yaml.sh
drewbo: a little fun on the side, will remove

scripts/create-test-users.js
drewbo: we don't really have standardized fixture data for them to test against but having them in an org could be okay

scripts/count-yaml.sh
apburnes: Oh. Hahah. Keep it.

drewbo:
Issues running this on admin/queues SPAs, closing for now

PR #4565, opened 2024-07-31
Reports API + Rendering

drewbo:
## Changes proposed in this pull request:
-
-
-

## security considerations
[Note the any security considerations here, or make note of why there are none]

PR #4577, opened 2024-08-21
feat: Add in app report rendering

apburnes:
Closes https://github.com/cloud-gov/pages-core/issues/4573
Supersedes #4565

- Adds new reports endpoints to serve updated site build task report api and site build report view
- Site build report views create a new react render and routing for reports
- Renders Zap and A11y reports

## security considerations
None


sknep (approved):
Looks like we're on our way

admin-client/src/components/TaskTable.svelte
```diff
@@ -1,5 +1,4 @@
 <script>
-    import prettyBytes from 'pretty-bytes';
```
sknep, line 2: If we aren't using prettyBytes anywhere else, we could remove it from the package.json file

apburnes: Done 🪄

sknep (approved):

PR #4587, opened 2024-08-30
feat: Automatically update build status on site build history page

apburnes:
Closes https://github.com/cloud-gov/pages-core/issues/4529
Closes https://github.com/cloud-gov/private/issues/1418

- Adds a new feature to automatically update a sites build task status
- Improves developer experience by providing mock report json data for site build tasks.

## security considerations
None


drewbo (dismissed):
Looks pretty good. A few notes but nothing that I think should block merge unless you want to take another look:
- The refresh triggers twice so the times will show up like: 10 seconds, 11, 25, 26, 40, 41; not sure what's causing the second render
- The refresh doesn't stop after the build completes. I'm not positive what the intent should be here but I think ideally it would stop refreshing after completion
- We could filter the localhost logic out of the controller for non-local builds by using a feature flag. I'd call that extremely low priority though

apburnes:
@drewbo I was thinking about switching to a feature flag for the dev reports. 

- For the build auto updates, I want the page to continue to update so new builds created by others would populate the page. Maybe we put logic to only refetch builds if the page is in focus. For the odd jump in build time since, it appears to be a subcomponent state re-rendering. I'll take a look.

apburnes:
@drewbo 

- Switched to a feature flag for dev reports
- Added a `document.hasFocus` method to only fetch while the user is on the page
- Created `refetchBuilds` action to avoid the multi-state update on subsequent build fetches.

drewbo (approved):

scripts/create-dev-data.js
```diff
@@ -640,6 +648,10 @@ async function createData() {
     message: 'Scan successfully completed. See artifact for details.',
     count: 3,
   });
+  localSiteBuildTasks.push({ id: btA11y1.id, type: 'a11y' });
+
+  // write localSiteBuildTasks.json file for viewing out reports
+  await writeFile(localSiteBuildTasksFile, JSON.stringify(localSiteBuildTasks), 'utf-8');
```
sknep, line 654: Is this file present but empty until then? or should it have something in it?

apburnes: You need to make sure your local environment is up to date by running the `yarn create-dev-data` script or the `make everything` script.

PR #4594, opened 2024-09-09
fix: Site builds polling to use doc visibilityState #4591

### LINKED ISSUE #4591: Fix auto-refresh pausing when browser window loses focus
see changes here:
https://github.com/cloud-gov/pages-core/pull/4587/files#diff-548742ab1172c818289c1d77b6cc3c0beef1be20417e9c3e5defd6fa66a10fdeR51-R62

Customer reported autorefresh not working after button was removed. Able to reproduce the build status failing to update if the status change occurs while tabbed away (browser window lacks focus). Status does not update when returning to the build history page, so autorefresh seems broken, and without the button, the customer must actually refresh the page to see new status. 

Alt-tabbing away or opening a new window while a build finishes is a normal/predictable behavior, but the refreshing pausing silently is not (at least to customers).

apburnes:
Closes #4591 

- Fixes site builds data auto update to use the document `visibilityState` to not poll the API when a user navigates to another browser tab or window.

## security considerations
None

### DIFF
--- a/frontend/components/site/siteBuilds.jsx
+++ b/frontend/components/site/siteBuilds.jsx
@@ -51,7 +51,7 @@ function SiteBuilds() {
     buildActions.fetchBuilds({ id });
 
     const intervalHandle = setInterval(() => {
-      if (document.hasFocus()) {
+      if (document.visibilityState === 'visible') {
         buildActions.refetchBuilds({ id });
       }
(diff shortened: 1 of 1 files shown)


drewbo (approved):
Confirmed on dev

sknep (approved):

PR #4599, opened 2024-09-17
Chore asr punchlist

apburnes:
Closes https://github.com/cloud-gov/pages-core/issues/4598

**General UX/design:**
- [x] Rework the info introducing the types of scans and their run days on the /sites/id/reports page, making sure it doesn't hide when a build filter is applied to the page
- [x] Wrap or truncate long paths on the a11y index page in the dropdowns
- [x] Make the page URL in the a11y single page header a link
- [x] Try to get the specific WCAG success criteria ("4.1.2") in the accessibility findings somewhere
- [x] Fix steps are buried under all of the evidence; bring it up further and style the additional info better

**Rule Suppression:**
- [x] Indicate whether a finding is suppressed more obviously; the italics aren't cutting it
- [x] Add a "what's this" or some kind of link to the suppression criteria section from the suppressed list so people can find out what that means
- [x] See about making the disabled trash can icon buttons less clicky looking / confusing
- [x] Alert the user they must save their changes to ruleset suppressions, especially if deleting a rule... or make that send to the backend right away?

## security considerations
None

### DIFF
--- a/frontend/components/site/SiteSettings/ReportConfigs.jsx
+++ b/frontend/components/site/SiteSettings/ReportConfigs.jsx
@@ -110,7 +110,7 @@ function ReportConfigs({ siteId: id }) {
     const isPagesRule = rule?.source === 'Pages';
     const customPlaceholder = '/assets class="ignore"';
     function pagesPlaceholder() {
-      if (rule.match) {
+      if (rule.match && Array.isArray(rule.match)) {
         return rule.match.map(match => (
           <code className="pages-suppressed-string" key="match">
             {match}
--- a/frontend/components/site/SiteSettings/ReportConfigs.jsx
+++ b/frontend/components/site/SiteSettings/ReportConfigs.jsx
@@ -108,7 +108,17 @@ function ReportConfigs({ siteId: id }) {
 
   function ruleRender(rule) {
     const isPagesRule = rule?.source === 'Pages';
-    const placeholder = (rule.match?.join(', ') || '') + (isPagesRule ? ' (not editable - suppressed by Pages)' : '/assets class="ignore"');
+    const customPlaceholder = '/assets class="ignore"';
+    function pagesPlaceholder() {
+      if (rule.match) {
+        return rule.match.map(match => (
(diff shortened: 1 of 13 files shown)


frontend/util/reports.js
```diff
@@ -91,3 +91,39 @@ export function plural(count, name) {
 export function relPath(url, baseurl) {
   return url.split(baseurl)[1];
 }
+
+export function getSuccessCriteria({ tags, id }) {
```
apburnes, line 95: set a default empty array `tags = []` incase the tags key is not an array passed to the function.

frontend/components/site/SiteSettings/ReportConfigs.jsx
```diff
@@ -108,7 +108,17 @@ function ReportConfigs({ siteId: id }) {
 
   function ruleRender(rule) {
     const isPagesRule = rule?.source === 'Pages';
-    const placeholder = (rule.match?.join(', ') || '') + (isPagesRule ? ' (not editable - suppressed by Pages)' : '/assets class="ignore"');
+    const customPlaceholder = '/assets class="ignore"';
+    function pagesPlaceholder() {
+      if (rule.match) {
+        return rule.match.map(match => (
```
apburnes, line 114: Will `rule.match` always exist as an array? Should we add a check in the `if`?

apburnes (commented):
This looks really good. Cleans up the UI and the copy. Just had a couple questions about adding some checks/defaults for arrays.

frontend/components/site/SiteSettings/ReportConfigs.jsx
sknep: @drewbo I'm not actually sure about this one

drewbo: yes, it's a little fragile but it's always set to an array via https://github.com/cloud-gov/pages-core/blob/main/frontend/components/site/SiteSettings/ReportConfigs.jsx#L64

drewbo (approved):
I went through all the checkboxes and confirmed on dev. LGTM

PR #4607, opened 2024-09-20
chore: UX improvements for findings with multiple match criteria

sknep:
- Allows for multiple match criteria in report
- Tells user to use commas when inputting more than one criteria in a customer-defined field
## security considerations
none - front end only


Ephraim-G (dismissed):

frontend/components/Reports/ScanFinding.jsx
```diff
@@ -25,15 +25,15 @@ const ScanFinding = ({
...
+    anchor = `finding-${finding.alertRef}${ignoreSource ? '-suppressed' : ''}`;
     count = parseInt(finding.count, 10);
     locations = finding.instances || [];
     references = finding.referenceURLs || [];
     hasMoreInfo = finding.otherinfo;
   }
   if (scanType === 'a11y') {
     title = `${finding.help}.`;
-    anchor = `finding-${finding.id}`;
+    anchor = `finding-${finding.id}${ignoreSource ? '-suppressed' : ''}`;
```
drewbo, line 36: I'd condition this on `ignore` rather than the source (since it's false for non-ignored items and custom ignored items)

drewbo (changes requested):

drewbo (changes requested):
Two other things I noticed:
- The summary table displays unsuppressed findings first but the list shows the suppressed findings first
- Both instances of the finding say "...finding was identified in X places." where X is the total between unsuppressed and suppressed; can we indicate how many instances are suppressed or not here also?

drewbo (approved):

Issue #4014, opened 2022-12-30
Refactor Test Suite


drewbo:
(mostly a note-taking exercise, need to update next steps)

As part of a broader [React and related dependency upgrade](https://github.com/cloud-gov/pages-core/issues/3372), we likely need to significantly change our current test suite:
- the primary test renderer `enzyme` does not have a supported adapter for React 18
- the current [React recommendation for test frameworks](https://reactjs.org/docs/testing.html#tools) is to use `jest` and `react-testing-library`; these seemingly have better integration with newer React features than `enzyme` or `mocha`

## Notes
- [React Testing Overview](https://reactjs.org/docs/testing.html)
- [React Testing Environments](https://reactjs.org/docs/testing-environments.html)
- [Jest](https://jestjs.io/)
- [React Testing Library](https://testing-library.com/docs/react-testing-library/intro)
- [RTL: migrate from enzyme](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme)

## Acceptance Criteria

- [ ] Determine next steps and acceptance criteria

drewbo:
Started initial work on this in https://github.com/cloud-gov/pages-core/tree/refactor-test-suite and updated the acceptance criteria

drewbo:
Update: I rebased the aforementioned branch and made a few updates to get linting + coverage working with latest main. The coverage stats show that this covers roughly 6% of the components and 4% of the whole frontend folder. An assortment of thoughts with no particular end goal:
- Switching from enzyme to RTL (without making changes to the underlying test intent) is a bit of a chore but also maybe could be partially automated with some regex?
- The non-component tests only need to be switched from chai to jest, which is mostly a matter of different assertion names. This is also a bit of a chore and maybe could be partially automated.
- The test logic itself probably needs a once over. That seems like a big undertaking to do all at once, maybe it could happen in conjunction with component updates on some sort of ongoing basis

PR #4613, opened 2024-10-04
Organize components

drewbo:
## Notes
I have some new thoughts on how we should proceed with #4014: basically I agree with @apburnes that our current tests aren't providing a lot of value and rather than porting them over, we should rewrite new RTL tests in an incremental fashion. Prior to doing that, I thought it would make some sense to organize our pages/components/containers in a way that better matches our [stated UI architecture design](https://github.com/cloud-gov/pages-core/blob/main/docs/DEVELOPMENT.md#architecture), this is a start on that work

- Move all files which render routes to `frontend/pages/`, don't use props for these (with one exception) and remove the older redux state connection pattern (replacing `connect` with `useSelector` will make it easier for us to update state management)

## security considerations
None


drewbo:
Noting for this thread that I'm not ignoring @apburnes we talked separately and agreed to fix up the aliases later

sknep (approved):
It's so beautiful

frontend/pages/organizations/List.jsx
```diff
@@ -1,9 +1,9 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
-import federalistApi from '../../util/federalistApi';
-import LoadingIndicator from '../LoadingIndicator';
-import { sandboxMsg } from '../../util';
+import federalistApi from '@util/federalistApi';
```
sknep, line 4: omg. yes

ci/tasks/test-api.sh
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 
 cd app
 
-./scripts/wait-for-it.sh db:5432 -- yarn test:cover; status=$?
-
-exit $status
+# ./scripts/wait-for-it.sh db:5432 -- yarn test:cover; status=$?
+echo "we're rewriting our tests!"
+exit 0
```
apburnes, line 8: Let's update this so we still run the server tests.

```suggestion
./scripts/wait-for-it.sh db:5432 -- yarn test:prepare && yarn test:server:cover
exit $status
```

frontend/routes.js
```diff
@@ -1,23 +1,13 @@
...
-import SitePublishedFilesTable from './components/site/SitePublishedFilesTable';
-import SiteSettings from './components/site/SiteSettings';
-import CustomDomains from './components/site/CustomDomains';
-import NewCustomDomain from './components/site/CustomDomains/New';
-import EditCustomDomain from './components/site/CustomDomains/Edit';
-import SiteReports from './components/site/SiteReports';
-import AddSite from './components/AddSite';
-import NotFound from './components/NotFound';
-import Error from './components/Error';
+import App from './pages/app';
```
apburnes, line 4: Is there a reason we can't alias `@pages`?

frontend/reportRoutes.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import { Route } from 'react-router-dom';
 
-import Report from './pages/Reports';
-import ShowReport from './pages/Reports/ShowReport';
-import ShowReportSubpage from './pages/Reports/ShowReportSubPage';
-import NotFound from './components/NotFound';
-import RouterError from './components/RouterError';
+import Report from './pages/reports';
```
apburnes, line 4: Is there a reason we can't alias @pages?

apburnes (commented):
The change of the folder structure is already much easier to understand and the aliasing is very nice. This has been sorely needed and it will improve our productivity immediately.

I have a question about pages aliases and running our api tests. Also, can you update the [`docs/DEVELOPMENT`](https://github.com/cloud-gov/pages-core/blob/main/docs/DEVELOPMENT.md) with the new frontend structure and aliases?

apburnes (approved):
💯

PR #4623, opened 2024-10-15
Feat: allow suppression of rules from the reports page

sknep:
Adds report suppression information to each finding
Allows the user to add or delete rules using the given rule id for each finding.

## security considerations
none


sknep:
Do you think text-based is enough? i'm worried about an actual "toggle" button being distractingly visible at the top of each finding.
Screenshot 2024-10-15 at 5 12 26 PM
Screenshot 2024-10-15 at 5 12 21 PM
Screenshot 2024-10-15 at 5 12 17 PM
Screenshot 2024-10-15 at 5 12 03 PM

apburnes:
@sknep will you run `make lint` and fix the linting errors before we review?

apburnes:
What's your thoughts on adding another toggle location at the top of the page in the violated rules list? Like 👇 
Screenshot 2024-10-16 at 12 27 04 PM

sknep:
hmm... I don't really want folks to be able to suppress without fully understanding the finding. Also that spot is already kind of busy/full, and keeping these reports scannable is important to me.

api/controllers/build-task.js
```diff
@@ -137,20 +119,32 @@ module.exports = wrapHandlers({
...
+      const reportReponse = await getObject(task.Build.Site, key);
+      const reportString = await reportReponse.Body.transformToString();
+      report = JSON.parse(reportString);
+    }
 
     const type = appMatch(task.BuildTaskType);
 
     const fullJSON = {
-      ...taskJSON, report, type, siteId: task.Build.Site.id,
+      ...taskJSON, report, type, siteId: task.Build.Site.id, siteBuildTaskId: task.SiteBuildTask.id,
```
drewbo, line 147: `siteBuildTaskId` doesn't need to be added here now that it's been added to the serializer/json/include

frontend/components/Reports/FindingSuppression.jsx
```diff
@@ -0,0 +1,136 @@
...
+  sbtId,
+  sbtType,
+  sbtCustomRules = [],
+}) {
+  const [customerRules, setCustomerRules] = useState(sbtCustomRules);
+  const willBeSuppressed = customerRules.find(r => r.id === ruleId);
+
+  useEffect(() => {
+    setCustomerRules(sbtCustomRules);
+  }, [sbtCustomRules]);
```
drewbo, line 21: I'm caveating this comment by saying we should probably investigate swapping in a consistent querying library like [`react-query`](https://tanstack.com/query/latest/docs/framework/react/overview) because I've also lost the plot a little on hook re-rending logic. But I think this might be the block that is causing you issues: I don't know that you need a `useEffect` call because `sbtCustomRules` are used to initialize the local state already. And they shouldn't change since we aren't refetching the report while on this page.

drewbo (changes requested):
I like the idea of having this alongside the rule (not in the top summary) and we can test out whether folks notice/use it in the future. I think maybe it could be more noticable but 🤷 . It looks like the suppression toggle logic has 2 issues:
- I think it's reversed because of some react hook "stuff" going on 
- If a user removes a suppression that had a `match` property, that would be removed when trying to re-enable it (maybe that's okay, we should just note)

frontend/components/Reports/FindingSuppression.jsx
sknep: So i deleted the useEffect, but I am not sure that the set state for customer rules is working... it seems to issue at the same time as the update function, so I just passed in the new array as a param instead. Is that bad? I feel like it needs a promise somewhere, but I haven't seen a pattern like that in how we manage updates

drewbo (approved):

apburnes (approved):

PR #4628, opened 2024-10-21
chore: improve component separation of concerns, hooks, purity

drewbo:
- Updates to each `page` component with the following heuristics in-mind:
  - Move things to custom hooks where possible (separation of concerns and will make it easier for us to standardize fetching logic, potentially with [react-query](https://tanstack.com/query/latest))
  - Move larger chunks of rendering logic into separate components (better modularity, testing, pure rendering)
  - Try [component composition](https://tkdodo.eu/blog/component-composition-is-great-btw) where applicable (increase readability, avoiding giant immediately invoked functions)

## security considerations
None


drewbo:
@apburnes I updated this to re-nest all the page templates to follow a file-based routing system (even though we're still using component based). I also removed four unused components, and moved things out of `shared` if they are only used by a single page (with the one exception of `FilterIndicator` which @sknep did such a nice job of writing in an abstract way that I feel like it warranted staying in `shared` and hoping for a future use)

apburnes (commented):
This is a great rework and a solid convention to set us up to improve our UI composition and testing in the future. With all theses updates, we may need to tweak the linting and bundling in CI to get them to pass.

frontend/pages/sites/$siteId/custom-domains/DomainLink.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function DomainLink({ domain }) {
```
apburnes, line 4: Add `export default`

frontend/pages/sites/$siteId/published/$name/FilesTable.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import globals from '@globals';
+
+import BranchFileRow from './BranchFileRow';
+
+function FilesTable({ files, name, children }) {
```
apburnes, line 8: Add `export default`

apburnes (approved):
This looks good to me. Making a note that we need to fix the [cron config](https://github.com/cloud-gov/pages-core/blob/main/api/workers/index.js#L44) on the recurring jobs before are next release.

PR #4639, opened 2024-10-28
chore: Update linting and formatting

apburnes:
This PR adds prettier formatting and updated eslint config. To minimize noise, future PR's will reformat the codebase using prettier and additional rules will be added and enforced.

## Notes
- Disabled sonarcube
- Removed airbnb defaults and switched eslint recommended

- Updates to eslint flat file config
- Updates app and admin client eslint
- Add prettier formatting and linting rules

## security considerations
Updates linting and formatting


api/controllers/site-branch-config.js
```diff
@@ -145,6 +145,7 @@ module.exports = wrapHandlers({
     const { id, site_id: siteId } = params;
 
     const site = await Site.forUser(user)
+      // eslint-disable-next-line testing-library/await-async-queries
```
drewbo, line 148: **non-blocking** I think this rule only applies to test files anyway so we should remove it from the code lint globally (rather than line by line). Also I'm not sure why it's only showing up in a few places since `find*` is a common pattern

ref: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-queries.md#rule-details

drewbo (dismissed):

api/controllers/site-branch-config.js
apburnes: I scoped it down to just `test/`, `*.spec.js|x`, `*.test.js|x` and removed the disable

drewbo (approved):

PR #4646, opened 2024-10-29
chore: Run prettier formatting on codebase

apburnes:

- Runs prettier formatting on lint
- Updates CI lint step to also check the formatting
- Adds format commands to docs and makefile

## security considerations
None


drewbo (changes requested):
Looks good. I ran `yarn format` and it made no changes so at least it's idempotent. The test failures are happening because they are expecting a message string with a newline which has been removed by the formatting. Otherwise good to go

drewbo (approved):

Issue #4611, opened 2024-09-27
Upgrade to Express 5


drewbo:
Express 5 is the first major version upgrade in a decade. It has two nice improvements for us: improved promise rejection handling (which has caused errors or annoying code her in the past) and reduction of (older) dependencies

## Notes

- https://www.trevorlasn.com/blog/whats-new-in-express-5
- https://expressjs.com/en/guide/migrating-5.html

## Acceptance Criteria

- [x] Application uses express 5
- [x] Confirm no use of removed methods and properties
  - [x] app.del()
  - [x] app.param(fn)
  - [x] Pluralized method names
  - [x] Leading colon (:) in the name for app.param(name, fn)
  - [x] req.param(name)
  - [x] res.json(obj, status)
  - [x] res.jsonp(obj, status)
  - [x] res.redirect(url, status)
  - [x] res.redirect('back') and res.location('back')
  - [x] res.send(body, status)
  - [x] res.send(status)
  - [x] res.sendfile()
- [x] Check new path route matching syntax
- [x] Accommodate new rejected promises handled from middleware and handlers
- [x] Review other changes related to these methods
  - [x] express.urlencoded
  - [x] app.router
  - [x] req.body
  - [x] req.host
  - [x] req.query
  - [x] res.clearCookie
  - [x] res.status
  - [x] res.vary
  - [x] res.render

drewbo:
Closed via #4664

PR #4664, opened 2024-11-15
Upgrade to express 5

drewbo:
- Upgrade to express 5
- Switches the local port for UAA to 9001

## security considerations
Upgrading express should provide security upgrades by avoiding relying on outdated dependencies


drewbo:
I used the checklist in #4611 (generated from the migration guide) to ensure we hit all the necessary items. A summary of the changes:
- Updated optional path parameter syntax
- Updated wildcard path syntax (requires naming)
- Removed `extended: false` from `urlencoded` since it's now the default

The new promise rejections should only be helpful, we might have a few unnecessary `try/catch` blocks out there but I didn't want to investigate each use.

cloud-gov-pages-operations:
  Total coverage (lines): **5.48%**

api/routers/report.js
```diff
@@ -4,6 +4,6 @@ const MainController = require('../controllers/main');
 const { csrfProtection } = require('../middlewares');
 
 // add csrf middleware to app route so that we can use request.csrfToken()
-router.get('/report(/*)?', csrfProtection, MainController.report);
+router.get('/report{*splat}', csrfProtection, MainController.report);
```
apburnes, line 7: Should this be `{/*splat}` with the forward slash like the `routes/main.js` changes?

PR #4683, opened 2024-11-25
Remove SiteUsers

### LINKED ISSUE #4339: Remove SiteUsers concept
Prior to the introduction of organizations, users were given access to a site via the `SiteUser` model. We need to remove this concept (all models, tables, endpoints, and related code) from the database, API, and application

## Acceptance Criteria

- [x] SiteUser model, Site Users column is removed in production database via migration
- [x] Authorizer and model scope code for builds (`.forSiteUser`) and for site (`.forUser`) are updated to only reference organization membership as the site access condition
- [x] SiteUser audit event is removed
- [x] SiteUser Redux actions, reducers, and associated front end code is removed
- [x] Rewrite `loadBuildUserAccessToken` to check organization users rather than `site.Users` for a valid access token
- [x] Remove or rewrite all associated tests

drewbo:
- closes #4339 
- Rewrites `Build.forSiteUser` and `Site.forUser` to only reference organization membership as a condition for viewing Builds/Sites.
- Removes the ability to add/remove users from a Site throughout the application (api/controller/actions/reducers)
- Adds a new convenience instance method to the `Organization` model, `addRoleUser` for more easily adding users to an organization
- Adds a new convenience instance method `get(Site)OrgUsers` to `Build` and `Site` for getting users which match the site's organization.
- Adds a new convenience test support function `createSiteUserOrg` which creates, as needed:
  - A `user`
  - Who is a member of an `org`
  - To which a `site` belongs
- A significant test suite rewrite to update our now singular access pattern (i.e. users must have an organization membership matching the site to access site resources instead of using `Site.users`)
  - Also removes "Basic Auth" tests and routes
  - Removes tests which reference organization membership as an alternative access pattern
  - Rewrites many tests (though not all, I got tired) to prefer `async` over `Promise`-based testing.

## Todo:
- [x] Both `loadBuildUserAccessToken` and the `createBuildForWebhookRequest` need a significant rewrite to query organization users rather than site users.
- [x] Analyze the potential impact of the above change. While we haven't added any new users this way for ~18 months, I'm still concerned that a number of sites build using much older user tokens
- [x] Test the new model methods
- [x] Remove all associated SiteUser concepts (table, `Site.users` column) from the database(s)

## security considerations
This is likely a significant security improvement (removes one access method) but it's also a major change that we should review thoroughly


drewbo:
I could be missing something but it seems like this change (if implemented correctly) won't impact any builds. This returned zero results:

```js
notoken = sites.filter(site => site.Organization.OrganizationRoles.every(role => !role.User.githubAccessToken))
```

drewbo:
I should note that I didn't verify the permissions of each token (by calling `GitHub.checkPermissions`) so it's possible some of the tokens aren't valid for the given repo and we could still run into some, albeit more minor, issues.

cloud-gov-pages-operations:
  Total coverage (lines): **14.51%**
  Coverage diff: **-0.32%** :chart_with_downwards_trend:

sknep:
We should also updated related documentation to make sure people understand that they can no longer add or remove people from individual sites.

drewbo:
@sknep do you know if our docs say that anywhere? We should update the bottom of https://cloud.gov/pages/documentation/access-permissions/ for inaccuracies but the top describes (correctly) organization access instead of the old method

sknep:
afaik, just that section. @Ephraim-G are you any more familiar with our docs? Or will you need to update any ZD macros now that everyone must be added to that site's assigned org to get access?

apburnes (approved):
This well tested and is a much needed cleanup of our user, org, and site model. We've worked through this update a lot and now it looks good to go.

Issue #4668, opened 2024-11-20
Add RTL tests to `@shared/UsaIcon.jsx`


apburnes:
Add a test to `@shared/UsaIcon.jsx` using the RTL framework. See our [docs](https://github.com/cloud-gov/pages-core/blob/main/docs/DEVELOPMENT.md#front-end) for writing frontend tests.

## Acceptance Criteria

- [x] Colocate test file to `@shared/UsaIcon.test.jsx`
- [x] Test the icon is showing as expected
- [x] Test when the icon is focusable
- [x] Test the size property
- [x] Test the role property
- [x] Remove component `defaultProps` and add the to the component args

PR #4698, opened 2024-12-20
chore: refactor @shared/UsaIcon.jsx and test  #4668

sknep:
- removed `aria-hidden` because node visibility ought to be determined by the `role` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden
- Removed unnecessary `aria-hidden` and `focusable` props, using `role` to determine visibility and focusability
- `focusable` is now `false` by default, and only set to `true` for `role="button"` and `role="link"`, per a11y guidelines
- replaced deprecated `xlinkHref` with`href` https://github.com/uswds/uswds-site/pull/2926 and https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href
- `role` must now be one of the following, suitable for icons: `img, presentation, button, link, status, none` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles
- `size` must be one of the USWDS icon size options, 3-9 https://designsystem.digital.gov/components/icon/#component-variants-icon
- add tests per #4668

## security considerations
- None, UI changes only

### DIFF
--- a/frontend/shared/UsaIcon.jsx
+++ b/frontend/shared/UsaIcon.jsx
@@ -1,30 +1,22 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-function UsaIcon({ focusable = false, role = 'img', name, size = null }) {
+function UsaIcon({ role = 'img', name, size = null }) {
   return (
     <svg
       className={size ? `usa-icon usa-icon--size-${size}` : 'usa-icon'}
-      aria-hidden="true"
-      focusable={focusable}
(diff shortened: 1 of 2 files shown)


cloud-gov-pages-operations:
  Total coverage (lines): **18.06%**
  Coverage diff: **0.12%** :chart_with_upwards_trend:

apburnes (approved):

PR #4711, opened 2025-01-29
update role authorization checks, auth logging, and code updates

### LINKED ISSUE #4656: Bug: Broken login method on chrome
Currently when trying to login to [https://admin.pages.cloud.gov](https://admin.pages.cloud.gov/login) from a Google chrome browser after using GSA credentials to authenticate the login page hangs. Initially the `callback` shows a `200` but when the page hangs and is refreshed a `500` error is then thrown.  

## Notes

Screenshot 2024-11-07 at 10 38 38 AM

* This does not occur in Firefox or Safari
* This error does not occur in an Google chrome Incognito window 

## Acceptance Criteria

- [ ] Login from a Google chrome browser is successful and does not return a `500`

drewbo:
- Does better assignment of `role` to avoid confusing admin error logging
- The above change prompted me to look at the removal more frequent checks against the user's UAA profile (because we can are storing role information in the session). This is a change to our authorization logic, more information below
- ~Reuses the same auth strategy on queues/bull-board for consistency~. Unfortunately, the bull-board app can't easily reuse the main `UAAClient`: The client has to read from the config files, which means that any time it's included, we need to full populate all the configuration variables. This would require that the bull-board deployment be bound to the same services as the primary application (database, credentials, etc) which seems unnecessary.

## security considerations
My understanding is that the authorization is checked in at least four places in the current application:
- API level: passport authentication only assigns sessions to pages users (for the main API) and pages support or admin staff (for the admin API). This provides a first intrinsic level of authorization.
- Route level: for the admin API, we check the `role` in the user session to authorize certain routes.
- Controller level: logic across various controllers checks whether users can access resources based on their organization.
- Service level: some services perform checks similar to the controller level. There were two `OrganizationService` methods which checked the users associated UAA user for groups to determine whether the user was an admin before performing certain actions. 
 
The final service level check has been moved to controller level logic. This introduces a possible future security weakness in that `createOrganization` or `inviteUserToOrganization` methods could be called with insufficient access. Currently the former is only called from the admin API and the latter is being supplied with a new parameter (default false) to determine whether it's being used in the admin API or not.


cloud-gov-pages-operations:
  Total coverage (lines): **18.28%**

test/api/unit/services/uaaStrategy.test.js
```diff
@@ -99,11 +102,14 @@ describe('verifyUAAUser', () => {
 
     cfUAANock.mockVerifyUserGroup(uaaId, uaaUserResponse);
```
apburnes, line 103: Do we still need to mock this since we took it out of the `verifyUAAUser`?

apburnes (commented):
This simplifies the logic and gives us better standard data values to expect. For the test's I noted there may be some mocks that are no longer necessary with the updated uaa client queries.

test/api/unit/services/uaaStrategy.test.js
drewbo: Good call. We actually aren't even using the `verifyUserGroup` method of the `UAAClient` (outside of one function noted below) so I'm inclined to remove the entire method (and all associated mocks).

The one remaining use is in the `Organization` service method `isUAAAdmin`. This in turn is used in only two other functions: invites to an org and creating orgs. Instead of checking the `UAAIdentity`, I'm going to see if there is an easy way to pass on the user role information that we set via passport

apburnes (approved):
This helps simplify the logic in the service and does a better job of simplifying the authorization logic in the controller. Updates and expands the test cases gives me more confidence in the refactor. Looks good to me.

api/services/organization/Organization.js
```diff
@@ -59,17 +59,6 @@ module.exports = {
     return uaaClient.inviteUserToUserGroup(targetUserEmail, accessToken);
   },
 
-  /**
-   * @param {UAAIdentityType} currentUserUAAIdentity
-   */
-  async isUAAAdmin(currentUserUAAIdentity) {
```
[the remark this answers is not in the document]
drewbo: I covered this in a more abstract way above but basically it was used in two spots:
- it was used in `createOrganization`, a method that was only called within the organization `create` controller, which is only called from the admin API. Here, I just removed it as duplicative
- it was used in `inviteUserToOrganization` which is called from both APIs. I added a new parameter to track where it's being called from to act accordingly

PR #4715, opened 2025-02-06
Feat add api endoints for file storage 2082

apburnes:
- Closes https://github.com/cloud-gov/private/issues/2082

## security considerations
Adds api endpoints for managing file storage service


cloud-gov-pages-operations:
  Total coverage (lines): **18.23%**
  Coverage diff: **-0.07%** :chart_with_downwards_trend:

api/services/file-storage/index.js
```diff
@@ -0,0 +1,297 @@
...
+      fileStorageServiceId: this.id,
+      fileStorageFileId: id,
+      method: FileStorageUserAction.METHODS.DELETE,
+      description: FileStorageUserAction.ACTION_TYPES.DELETE_FILE,
+    });
+
+    return result;
+  }
+
+  async deleteDirectory() {}
```
drewbo, line 187: Can be removed?

api/utils/index.js
```diff
@@ -205,18 +205,23 @@ function omitBy(fn, obj) {
...
+) {
   const limit = toInt(params.limit) || 25;
   const page = toInt(params.page) || 1;
   const offset = limit * (page - 1);
 
   const pQuery = {
     limit,
     offset,
-    order: [['createdAt', 'DESC']],
+    order: [order],
```
drewbo, line 222: This might have already been incorrect but `api/admin/controllers/domain.js` uses the paginate function twice with `order` supplied in the `query`; I'd check that this doesn't create two separate sorts

api/utils/index.js
```diff
@@ -303,6 +308,60 @@ function appMatch(type) {
...
+
+  if (!['string', 'number'].includes(argType)) {
+    throw new Error('Text must be a string or number.');
+  }
+
+  const str = text.toString();
+
+  const [base, extension] = splitFileExt(str);
+
+  if (str.length > 200) {
```
drewbo, line 332:
```suggestion
  if (str.length > len) {
```

drewbo (changes requested):
Great, well thought-out, feature with tests to boot. I added a few notes on some small changes that I think could clean up the code a little bit but overall looks great. I didn't focus too much on models since I had already seen most of that.

Haven't checked on dev yet but I'll see if I can mess with the functionality today

api/admin/controllers/file-storage.js
```diff
@@ -0,0 +1,32 @@
...
+        Event.labels.ADMIN,
+        user,
+        'Site File Storage Service Created',
+        fss,
+      );
+
+      const data = serializeFileStorageService(fss);
+      return res.send(data);
+    } catch (error) {
+      return res.status(error.status).send(error);
```
drewbo, line 29: One issue with this pattern is that unhandled errors won't be logged anywhere, it will first show that it doesn't have an associated status:
`TypeError: Invalid status code: undefined. Status code must be an intege`

api/authorizers/file-storage.js
```diff
@@ -0,0 +1,82 @@
+const siteErrors = require('../responses/siteErrors');
+const { FileStorageService, Site } = require('../models');
+const { isSiteOrgManager, isOrgManager, isOrgUser } = require('./utils');
+
+const canCreateSiteStorage = async ({ id: userId }, { id: siteId }) => {
```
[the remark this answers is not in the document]
apburnes: We had this `fn({ id: userId }, { id: siteId })` convention with other authorizers so I kept it but it did give me a bit of a bad taste. I like the `fn(userId, siteId)` so I will update them to that new signature.

api/middlewares/mulipart-form.js
```diff
@@ -0,0 +1,6 @@
+const multer = require('multer');
```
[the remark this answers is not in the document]
apburnes: I went with multer because it was specifically called out in [the middleware docs](https://expressjs.com/en/resources/middleware/multer.html) but we can test it against [formidable](https://www.npmjs.com/package/formidable) since that package has a more recent update cadence.

api/utils/index.js
apburnes: I'll take a look

apburnes: It is just overwriting the preset `order` if the `order` is added in the query params. I'm going to move the order into the `query` params.

api/admin/controllers/file-storage.js
apburnes: Updated

api/controllers/file-storage-service.js
```diff
@@ -0,0 +1,218 @@
...
+const {
+  serializeFileStorageService,
+  serializeFileStorageFile,
+} = require('../serializers/file-storage');
+const { wrapHandlers } = require('../utils');
+const { Event } = require('../models');
+const { SiteFileStorageSerivce } = require('../services/file-storage');
+const badRequest = require('../responses/badRequest');
+
+module.exports = wrapHandlers({
```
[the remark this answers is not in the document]
apburnes: For DRY, this seems like a bit premature for how often we use the class but there could be two options to optimize. 

Create a private method that checks if the credentials have been fetched for all methods and then either fetches or uses the instance credentials when a method is called. 

```js
const client = new SiteFileStorageSerivce(fileStorageService, user.id);
const results = await client.<some method>
```

Or we create a functions that instantiate the class, create the client and then pass its args to the method.

```js
async function createFile(fileStorageService, userId, key, name, blob) {
  const siteStorageService = new SiteFileStorageSerivce(fileStorageService, userId);
  const client = await siteStorageService.createClient();

  return client.createFile(key, name, blob)
}
```

Our S3 services rotate credentials so we would have to build some type of credentials session to share between instances and identify the refresh schedule. That adds additional work beyond any possible gains and UX improvements at this point in development. Tracking this going forward would be worth while to track and make that decision at a later date.

drewbo (dismissed):

drewbo (approved):

PR #4722, opened 2025-02-13
fix: admin auth still requires certain roles

drewbo:
- fixes an access probably introduced in #4711
- fixes prior to production release

## security considerations
Stops non admin/support users from accessing the admin panel


apburnes (approved):
🎉

Issue #4690, opened 2024-12-16
Test pages/organizations/OrganizationCard.jsx 


sknep:
Add a test to pages/organizations/OrganizationCard.jsx using the RTL framework. See our [docs](https://github.com/cloud-gov/pages-core/blob/main/docs/DEVELOPMENT.md#front-end) for writing frontend tests.

### Acceptance Criteria

- [x]  Colocate test file to pages/organizations/OrganizationCard.test.jsx
- [x]  Test the component renders with the proper organization and role
- [x]  Test that the sandbox notification shows up for orgs that are in a sandbox, using daysuntil
- [x]  Link should show up if the org is active AND `role.name === 'manager'; test that the link uses the correct org id
- [x]  Remove component defaultProps (if any) and add the to the component args

PR #4817, opened 2025-09-09
chore: test for OrganizationCard.jsx (#4690)

afishman-2025:

-Test the component renders with the proper organization and role
-Test that the sandbox notification shows up for orgs that are in a sandbox, using daysuntil
-Link should show up if the org is active AND `role.name === 'manager'; test that the link uses the correct org id

https://github.com/cloud-gov/pages-core/issues/4690

## security considerations
There are no anticipated security issues.
Does not leverage any secret keys or expose data.

### DIFF
--- a/frontend/pages/organizations/OrganizationCard.test.jsx
+++ b/frontend/pages/organizations/OrganizationCard.test.jsx
@@ -1,0 +1,184 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { createFixtureOrg } from '../../../test/frontend/support/data/organizations';
+import { randomUUID } from 'node:crypto';
+import OrganizationCard from './OrganizationCard';
+
+// Mocking react-router-dom's <Link> component
+// to avoid rendering actual routing elements in tests.
+// This prevents warnings  when <Link> is used outside a Router
+jest.mock('react-router-dom', () => {
+  const original = jest.requireActual('react-router-dom');
+  return {
+    ...original,
+    Link: ({ to, children }) => {
+      return `Link[to="${to}"] ${children}`;
+    },
+  };
+});
+
+describe('OrganizationCard', () => {
+  function testOrganizationCard(organizationName) {
+    expect(screen.getByText(organizationName)).toBeInTheDocument();
+    expect(screen.getByRole('listitem')).toHaveClass('usa-card');
+  }
+
+  function testMessage(isSandbox, days) {
+    const m = `All data for this sandbox organization will be removed in ${days} days.`;
+    const emElement = screen.queryByText(m, { selector: 'em' });
+    if (isSandbox) {
+      expect(emElement).toBeInTheDocument();
+      expect(emElement.tagName).toBe('EM');
+    } else {
+      expect(emElement).not.toBeInTheDocument();
+    }
+  }
+
+  function testOrganizationActiveStatus(isActive) {
+    const footer = screen.queryByText('Inactive');
+    if (!isActive) {
+      expect(footer).toHaveClass('usa-card__footer');
+    } else {
+      expect(footer).not.toBeInTheDocument();
+    }
+  }
+
+  function testOrganizationLink(isLink, organizationId) {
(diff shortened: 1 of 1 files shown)


frontend/pages/organizations/OrganizationCard.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+      role: {
+        id: randomUUID(),
+        name: roleName,
+      },
+    };
+    render(<OrganizationCard {...testProps} />);
+    return testProps;
+  }
+
+  it('renders correctly active production organization for a non-manager role', () => {
```
sknep, line 73: `correctly` or `currently`? I don't quite follow this phrasing.

We have two user types right now, `user` and `manager`. Does it make more sense to say that the test is for the user role, or is it better to write it this way in case we ever have 3 or more roles? (genuine question)

frontend/pages/organizations/OrganizationCard.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+  it('renders correctly active production organization for a manager role', () => {
+    const testData = {
+      isActive: true,
+      isSandbox: false,
+      roleName: 'manager',
+    };
+    const testProps = initTest(testData);
+
+    testOrganizationCard(testProps.organization.name);
+    testMessage(testData.isSandbox, testProps.organization.daysUntilSandboxCleaningdays);
```
sknep, line 96: if we're checking the contents of the sandbox message, then is this more than just checking that we're showing the right org? is it about showing more than that?

or is this test just being super thorough?

frontend/pages/organizations/OrganizationCard.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+    return testProps;
+  }
+
+  it('renders correctly active production organization for a non-manager role', () => {
+    const testData = {
+      isActive: true,
+      isSandbox: false,
+      roleName: '!manager',
+    };
+    const testProps = initTest(testData);
```
sknep, line 79: The test label says its checking that such-and-such `renders correctly`... but all I'm seeing the test do is check that the props the component has match the props that were passed in. Could we actually check the rendered DOM instead? Or, if you think that's not needed, what would be a better test verb than "renders" since it doesn't seem to match what you're testing?

frontend/pages/organizations/OrganizationCard.test.jsx
afishman-2025: testMessage() checks if message is shown or not shown.

frontend/pages/organizations/OrganizationCard.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { createFixtureOrg } from '../../../test/frontend/support/data/organizations';
+import { randomUUID } from 'node:crypto';
+import OrganizationCard from './OrganizationCard';
+
+// Mocking react-router-dom's <Link> component
+// to avoid rendering actual routing elements in tests.
+// This prevents warnings  when <Link> is used outside a Router
+jest.mock('react-router-dom', () => {
```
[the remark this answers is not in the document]
afishman-2025: I have to revert commit which applied Link mock to all tests. There are multiple tests which fail with Link mock because they expect a link as html element and it's specific format.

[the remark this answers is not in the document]
sknep: fine to revert while you're working, but I think Andrew is still enforcing conventional commits, which means a rebase to clean the history and squash all the commits into one. I may be wrong though!

frontend/pages/organizations/OrganizationCard.test.jsx
sknep: OH, I see. so abstract, so clean!

frontend/pages/organizations/OrganizationCard.test.jsx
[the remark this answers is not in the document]
afishman-2025: done

sknep (approved):
Thanks! This is spotless ✨

PR #4857, opened 2026-03-03
Feat create workshop oauth flow 4855

afishman-2025:
-OAuth flow for Workshop integration

## security considerations
GitLab access and refresh tokens stored in the database are encoded using the same pattern, service, and encryption key as other parts of the application.


api/services/GitLab.js
```diff
@@ -0,0 +1,54 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${user.gitlabToken}`,
+    },
+    body: JSON.stringify({
+      client_id: gitlabConfig.clientID,
+      client_secret: gitlabConfig.clientSecret,
+    }),
+  })
+    .then((response) => {
```
apburnes: Since this is an `async` and `await` pattern function, we don't need to use the Promise `then` pattern.  We cannot guarantee the `refreshToken` will finish by the time the function call completes.

Let's do

```js

try {
  const revokeResponse = await fetch(`${gitlabConfig.baseURL}/oauth/revoke`, ...)

  if (!revokeResponse.ok && revokeResponse.status !== 404) {...}

  const refreshResponse = await refreshToken(user)

if (refreshResponse.status !== 401) { ... }

} catch (error) {
  logger.error('Error revoking GitLab token', error);
}


```

api/services/user.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  async updateGitLabTokens(user, { accessToken, refreshToken, expiresIn, createdAt }) {
+    await user.update({
```
apburnes: If we ever need access to the user object at some point we can return the Promise and it will complete.

```suggestion
    return user.update({
```

api/services/user.js
```diff
@@ -0,0 +1,16 @@
...
+      gitlabToken: accessToken,
+      gitlabRefreshToken: refreshToken,
+      gitlabExpiresAt: new Date((createdAt + expiresIn) * 1000),
+    });
+  },
+  async resetGitLabTokens(user) {
+    await user.update({
```
apburnes: If we ever need access to the user object at some point we can return the Promise and it will complete.

```suggestion
    return user.update({
```

PR #4891, opened 2026-04-14
fix: Public File details modal buttons

apburnes:
- Updated the dialog modal buttons to fix secondary and dismiss buttons
- Updated the file list row delete button hover
- Updated the public file details delete button style to match the file list row delete

File row delete button hover:

**Before**
Screenshot 2026-04-14 at 8 59 08 AM

**After**
Screenshot 2026-04-14 at 8 59 28 AM

File Details actions:

**Before**
Screenshot 2026-04-14 at 9 01 08 AM

**After**
Screenshot 2026-04-14 at 9 01 01 AM

### DIFF
--- a/frontend/pages/sites/$siteId/storage/FileDetails.jsx
+++ b/frontend/pages/sites/$siteId/storage/FileDetails.jsx
@@ -79,13 +79,12 @@ const FileDetails = ({
           </tr>
           <tr>
             <th scope="row">Actions</th>
-            <td style={{ display: 'flex', gap: '12px', flexWrap: 'wrap' }}>
+            <td className="file-actions-cell">
               <a
                 href={fullPath}
                 target="_blank"
                 rel="noopener noreferrer"
-                className="usa-button"
-                style={{ margin: '0' }}
+                className="usa-button margin-0"
               >
                 View File
               </a>
--- a/frontend/pages/sites/$siteId/storage/FileDetails.jsx
+++ b/frontend/pages/sites/$siteId/storage/FileDetails.jsx
@@ -79,12 +79,13 @@ const FileDetails = ({
           </tr>
           <tr>
             <th scope="row">Actions</th>
-            <td>
+            <td style={{ display: 'flex', gap: '12px', flexWrap: 'wrap' }}>
               <a
                 href={fullPath}
                 target="_blank"
                 rel="noopener noreferrer"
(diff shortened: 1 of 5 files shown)


KKAtila (dismissed):

frontend/pages/sites/$siteId/storage/FileDetails.jsx
```diff
@@ -79,12 +79,13 @@ const FileDetails = ({
           </tr>
           <tr>
             <th scope="row">Actions</th>
-            <td>
+            <td style={{ display: 'flex', gap: '12px', flexWrap: 'wrap' }}>
```
sknep: you sure you don't want to do this in the css file?

frontend/sass/_tables.scss
```diff
@@ -725,20 +725,39 @@ $highlightedAndSelectedCellBackground: #dde7ef;
...
-.file-details {}
+.file-details {
+  .delete-button {
+    padding: 0.4em 0.6em;
+    font-size: 1.1em;
+    border-radius: 4px;
+    white-space: nowrap;
+    margin-right: 0;
+    // mimics error colors from USWDS -- no outline button for this
+    box-shadow: inset 0 0 0 2px #d83933;
```
sknep, line 749: could we use vars instead? i can jump in if you wnt

afishman-2025 (dismissed):

frontend/pages/sites/$siteId/storage/FileDetails.jsx
apburnes: Fixed

frontend/sass/_tables.scss
apburnes: Do you want to put together the right vars?

sknep: we could spec them here: https://github.com/cloud-gov/pages-core/blob/1759530f79a05f98a4ee53ad40d63f74a2ba6fe4/frontend/sass/_variables.scss

sknep: there are a lot of other places we use unnamed colors though, so I wouldn't throw a fit without it. just thought I'd ask in case it was already in use -- I don't see a place where these are used elsewhere at the moment. what do you think?

sknep (approved):
thanks for addressing the inline styles

PR #4902, opened 2026-04-23
fix: unable to find valid access token to report build status 4898

afishman-2025:
- GitLab token deleted only if user starts "Reset your GitLab Access Token." in UI.
- Renamed isWorkshop() to isWorkshopPlatform() and use isWorkshopUrl() or isWorkshopPlatform() based on the argument type.
- Extra check that GitLab token exists before refresh.
- In SourceCodePlatformHelper.getAccessTokenWithCertainPermissions() fixed getToken(), which returned GitHub token instead GitLab token.
- Fixed stack overflow in SourceCodePlatformHelper.getAccessTokenWithCertainPermissions(), which happened if is error thrown.
- Redesigned flows where GitLab token is required so token refresh happens at the beginning of the flow and only if it is valid for less than an hour.
- For gitlab build, user with a valid and refreshed token is added to the build upon creation.
- Error from commit state post to GitLab is ignored if it occurs because the same state was posted more than once.
- When looking for a user with GitLab permissions for the build, added a check for the user from the latest successful site build before looping through all organization users.

## security considerations
None

### DIFF
--- a/test/api/unit/services/Webhooks.test.js
+++ b/test/api/unit/services/Webhooks.test.js
@@ -13,6 +13,10 @@ const githubAPINocks = require('../../support/githubAPINocks');
 const { createSiteUserOrg } = require('../../support/site-user');
 
 const Webhooks = require('../../../../api/services/Webhooks');
+const config = require('../../../../config');
+
+const { authorizationOptions: gitlabConfig } = config.passport.gitlab;
+gitlabConfig.baseURL = 'https://workshop.cloud.gov/';
 
 describe('Webhooks Service', () => {
   const buildWebhookPayload = (user, site, pushedAt = new Date().getTime() / 1000) => ({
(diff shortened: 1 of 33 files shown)


cloud-gov-pages-operations:
  Total coverage (lines): **38.78%**

apburnes (commented):
Getting the `Failed to send commit status ` error in the `sendCommitStatus` function in the `GitLabHelper` file. The Github sourcecode provider works as expected

afishman-2025:
> Getting the `Failed to send commit status ` error in the `sendCommitStatus` function in the `GitLabHelper` file. The Github sourcecode provider works as expected

Fixed.

user.json
apburnes, line 1: Can we delete this?

test/api/unit/services/Webhooks.test.js
```diff
@@ -698,4 +702,93 @@ describe('Webhooks Service', () => {
...
+        ),
+      ).to.deep.equal({
+        owner: 'anna.fishman',
+        repository: 'site-4',
+      });
+    });
+  });
 });
+
+const gitlabWebhookPayload = {
```
apburnes: Let's use fake data

apburnes (commented):
Great addition. I added a couple comments about clean up.

user.json
afishman-2025: Done.

test/api/unit/services/Webhooks.test.js
afishman-2025: Fixed.

apburnes (approved):
This looks good to go.

